PR #2, opened 2023-05-20
Added Catppuccin Color Scheme

Stetsed:
Added a catppuccin color scheme


Stetsed:
https://pocketbase.selfhostable.net/api/files/images/i7p29hh351u1k07/2023_05_20_t09_01_50_utc_XHf6howYyf.png

Screenshot of how it looks

Stetsed:
Also this isn't in this commit but don't you want the bar to be exclusive so it doesn't overlap stuff? Right now you have exclusive set to false.

Stetsed:
Please do not merge this yet I just realised I need to add the hypr color section

Stetsed:
Fixed

end-4:
I have the bar not exclusive because i set monitor `addreserved` manually (i don't want windows to move if I have to restart the bar or something)

And did you forget to actually modify the colors so it's actually catppuccin?

Stetsed:
Oh I see I did something wrong and it's weird, the calendar is catppuccin the bar isn't.

end-4:
As I use a mix of Material and Pywal colors, modifying only one file won't work

Stetsed:
@end-4 yep I saw it, see that I need to modify those 3 files which I am doing now.

Stetsed:
@end-4 Okay now I have added the catppuccin color scheme, mostly did it by eye by what looked good.

https://pocketbase.selfhostable.net/api/files/images/8cesx0hc6pl69nw/2023_05_20_t09_30_54_utc_4KyOfDP5mY.png

end-4:
- if you're going to name it catppuccin at least copy the color pallete? (btw which variant? would be nice if you could specify it) 
- else you could just use another name
- also why are you changing the y offset of the `osettings` window?

Stetsed:
Ah sorry I was on the wrong branch, I'll remove that commit as the osettings thing is for dynamic sizing. Also I did copy the right color pallete I used catppuccin mocha(I'll rename it to that)

end-4:
It looks a bit dark and has no purple... but the colors seem right. I'll just find a way to improve the color selection later.

PR #14, opened 2023-06-10
Update allappnames, prevent failure for invalid locations 

realDragonium:
Prevent allappnames from failing if directory its looking for doesn't exists


end-4:
thanks

PR #15, opened 2023-06-11
Extend pip install dependency README.md

realDragonium:
Extend pip install command with dependency necessary for `.config/eww/scripts/get-icon.py`

Question about the icons from that script, they aren't included in this repository right?

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@
  - Python
 ```
 [ Command ]
-pip install pywal desktop_entry_lib poetry build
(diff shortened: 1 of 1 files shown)


end-4:
oh
that script is no longer used
now [geticons](https://aur.archlinux.org/packages/geticons) is used
i'll remove that script now

end-4:
icons are bundled in this repo
you can find them in .local/share/icons/

PR #16, opened 2023-06-11
Extend package list

realDragonium:
extend readme packages which need to be installed with `cliphist`, required for: `SUPER, V`

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ pip install pywal desktop_entry_lib poetry build
  - Other Dependencies (install it with your distro's package manager)
 ```
 [ Possible package names ]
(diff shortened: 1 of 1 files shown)


end-4:
alright, really appreciate how you're checking all the dependencies!

realDragonium:
No problem! To be honest, im only updating the things relevant for me (sorry). 

For example these are also missing from the readme (if you compare it with keybinds.conf): yuzu, wezterm, foot, microsoft-edge-stable

end-4:
hmm i should add foot.
i have wezterm only for messing around, i dont actually use it
i just think adding those packages would be too much
who tf recommends an emulator and browser in a dotfiles repo lol
they have nothing to deal with the hyprland experience

realDragonium:
True, but on the other hand, its provided in keybinds.conf in this repo. So I wouldn't think its too weird if it was. Im seeing more as: "Hey, here is a list of packages you need if you want to run setup as is."  

Not that either of them are wrong, its just you prefer

PR #17, opened 2023-06-11
Update eww/scripts/memory to work without swap

realDragonium:
Update script to check or system has swap, if it doesn't it will skip over the calculation to prevent divide by 0 error.

Something else i found, this script only works with locales which uses `.` as decimal separator. Other languages which use a `,` will not work. We could fix this for them by adding `LC_NUMERICAL=en_US.UTF-8` to the script. But i will leave it up to you to decide.

### DIFF
--- a/.config/eww/scripts/memory
+++ b/.config/eww/scripts/memory
@@ -22,7 +22,7 @@ while true; do
   perc=$(printf '%.1f' "$(free -m | rg Mem | awk '{print ($3/$2)*100}')")
 
   swapfree=$(printf '%.1fG' "$(bc -l <<< "($swapt - $swapu) / 1000")")
-  swapperc=$(printf '%.1f' "$(free -m | rg Swap | awk '{print  ($2!=0) ? ($3/$2)*100 : 0)")
+  swapperc=$(printf '%.1f' "$(free -m | rg Swap | awk '{print  ($2!=0) ? ($3/$2)*100 : 0}')")
 
   echo '{ "total": "'"$total"'", "used": "'"$used"'", "free": "'"$free"'", "swaptotal": "'"$swaptotal"'", "swapused": "'"$swapused"'", "swappercentage": '"$swapperc"', "swapfree": "'"$swapfree"'", "percentage": '"$perc"' }'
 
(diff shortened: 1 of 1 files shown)


.config/eww/scripts/memory
```diff
@@ -22,7 +22,7 @@ while true; do
   perc=$(printf '%.1f' "$(free -m | rg Mem | awk '{print ($3/$2)*100}')")
 
   swapfree=$(printf '%.1fG' "$(bc -l <<< "($swapt - $swapu) / 1000")")
-  swapperc=$(printf '%.1f' "$(free -m | rg Swap | awk '{print ($3/$2)*100}')")
+  swapperc=$(printf '%.1f' "$(free -m | rg Swap | awk '{print  ($2!=0) ? ($3/$2)*100 : 0)")
```
end-4: you forgot to close the curly bracket and apostrophe `}'`
should be
```bash
swapperc=$(printf '%.1f' "$(free -m | rg Swap | awk '{print ($2!=0) ? ($3/$2)*100 : 0}')")
```

end-4:
a
@realDragonium are you alive

realDragonium:
I think so, are you? @end-4

end-4:
just fix resolve that review lol

realDragonium:
I don't see a review?

realDragonium:
image
Im not blind, right?

end-4:
bruh
i didnt press comment review

.config/eww/scripts/memory
realDragonium: Woops, good catch my bad. Looks like forgot to copy that part. 
```suggestion
  swapperc=$(printf '%.1f' "$(free -m | rg Swap | awk '{print  ($2!=0) ? ($3/$2)*100 : 0}')")
```

.config/eww/scripts/memory
```diff
@@ -22,7 +22,7 @@ while true; do
   perc=$(printf '%.1f' "$(free -m | rg Mem | awk '{print ($3/$2)*100}')")
 
   swapfree=$(printf '%.1fG' "$(bc -l <<< "($swapt - $swapu) / 1000")")
-  swapperc=$(printf '%.1f' "$(free -m | rg Swap | awk '{print ($3/$2)*100}')")
+  swapperc=$(printf '%.1f' "$(free -m | rg Swap | awk '{print  ($2!=0) ? ($3/$2)*100 : 0}')")
```
realDragonium:
```suggestion
  swapperc=$(printf '%.1f' "$(free -m | rg Swap | awk '{print ($2!=0) ? ($3/$2)*100 : 0}')")
```
Lets also remove the extra space

end-4:
thanks

PR #19, opened 2023-06-15
Replaced /home/end to $HOME

D3vil0p3r:
Replaced /home/end to $HOME in order to work with all users for main branch.


end-4:
Ahh stupid me
That line doesn't do anything, you should delete it lol

PR #58, opened 2023-08-21
Move blur and it's relevant config paths to subcategory

SpaceNerden:
Required by hyprland `v0.28.0` as it moved the blur settings from `decorations` to it's own subcategory, `decorations:blur`

Pardon the stray .gitignore though. I don't know how to only make a PR without it.

### DIFF
--- a/.config/eww/scripts/hyprsettings
+++ b/.config/eww/scripts/hyprsettings
@@ -8,10 +8,10 @@ getoptions(){
   input_sensitivity=$(hyprctl getoption input:sensitivity -j | gojq -r -c '.float')
   touchpad_disable_while_typing=$(hyprctl getoption input:touchpad:disable_while_typing -j | gojq -r -c '.int')
   touchpad_clickfinger_behavior=$(hyprctl getoption input:touchpad:clickfinger_behavior -j | gojq -r -c '.int')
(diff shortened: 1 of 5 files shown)


SpaceNerden:
By the way, thanks so much for this cool rice! I hope you don't mind me using it.

I'm new to this part of linux so I don't really know much about eww and hyprland but would it be possible to have 2 rices and switch between them? 
One idea I though of right now was to link the .config/{hypr, eww} in the git repo and other folders to .config and just switch branches? Not sure if that'll work.

end-4 (approved):
alrighty 🍚

end-4:
there's actually an undocumented system for switching between configs. it's very poorly set up, that's why. anyway, here's how
- create folder `~/.config/__enderice`
- copy eww and hypr folders to that folder and rename them to `hypr_RICENAME` and `eww_RICENAME`. RICENAME can be anything, just make sure you follow that syntax.
- open the launcher and type `>swr SAVING_RICE_NAME LOADING_RICE_NAME`
so your current setup will be saved as SAVING_RICE_NAME and the LOADING_RICE_NAME set will be loaded

SpaceNerden:
Thanks! I'll try it when I have the time.

PR #92, opened 2023-10-14
better bottomsheet

OnlyCS:
Looks like this:
example


end-4:
i don't see how it's better
you removed the background cover

OnlyCS:
the glass looks better in my opinion. especially because it now syncs with the material colors (didn't before)

end-4:
of course, i appreciate the effort
i just don't like how the cover pic is removed
but on second thought, that background cover does break sometimes and creates performance issues, removing it isn't a bad idea

PR #130, opened 2023-12-03
Added kb layout icons for the Hyprland kb_layout option

SuperSuslik312:


.config/ags/scripts/layout.sh
end-4, line 1: I'd rather have this in `statusicons.js`. just run hyprctl devices with an execAsync then do the json grabbing in js.

end-4 (approved):
Looks good
Thanks!

PR #132, opened 2023-12-04
remove linux 'date' command from js

SuperSuslik312:
why use 'date' if there is a 'GLib.DateTime'?


end-4:
right.
i had these since my hello world days... the `date` method was from the example config

Issue #126, opened 2023-11-27
Dynamically parse icons for "overview" widget


Gufderald:
- **I have read the instructions** and I have a problem with branch: [illogical-impulse](https://github.com/end-4/dots-hyprland/tree/illogical-impulse)
---
## The issue
- Hi again! This issue follows from https://github.com/end-4/dots-hyprland/issues/118, it also should be  considered a feature request and low priority.
 
So, it would be great to parse substitutions for icon names dynamically; `applications` folders (`/usr/share/applications`, `~/.local/share/applications`) can be used for that, as `.desktop` files store correct icon names. Also this probably may be achieved via [Applications ags  service](https://github.com/Aylur/ags/wiki/Applications).

gori-yar4e:
Hi! It seems to me that in most cases icon name can really be covered by lowercasing class name.  Prepared an PR, seems to work good for now.

Gufderald:
Hi guys! Just actualized my configs from branch - it's better now, but still missing some icons (smth around 50% of my apps).  I'll also try solution from @gori-yar4e in the nearest future

Gufderald:
So, seems that @gori-yar4e's solution works better.

```js
const iconNameMap = new Map()
Applications.list.forEach((app) => {
    iconNameMap.set(app.desktop.split('.desktop')[0].toLowerCase(), app['icon-name'])
})

function substitute(str) {
    const subs = [
        { from: 'Gimp-2.10', to: 'gimp' },
    ];

    for (const { from, to } of subs) {
        if (from === str)
            return to;
    }
    return iconNameMap.get(str.toLowerCase()) || 'image-missing';
}

```
produces 
![image](https://github.com/end-4/dots-hyprland/assets/75615922/fd365848-909a-494f-8ed6-2571003bff04)


and source config produses 
![image](https://github.com/end-4/dots-hyprland/assets/75615922/1d898717-99b0-4220-b804-4ff86aefe9d5)
(same apps)

Gufderald:
There is some another trouble with apps installed by snap - only telegram does not require handmade substitution. Going to research it in a near future.

clsty:
Any progress on this issue? I see Tilix missing its icon on the overview widget currently. The desktop file of Tilix has a line `Icon=com.gexperts.Tilix` and this icon appears normal when running Tilix on xfce4.

yrnmsk:
I also want to bump on this issue.

---

[...]

PR #135, opened 2023-12-11
feat: auto detect app icon into dock

gori-yar4e:
Here's improved method of icons substitution, based on ags Apllications service usage and idea that in most cases window class name is same to it's `.desktop` file. Related to issue https://github.com/end-4/dots-hyprland/issues/126


end-4:
- Please
  - Don't remove my substitutions
  - Keep the replace function, since window classes can have spaces.
- Can you tell me what apps benefit from this? I'll be happy to merge with the above changes applied, but it'd be nice if I could actually test it

end-4:
oops
i think my repo cleanup (removes history) overwrote this

PR #136, opened 2023-12-12
Update install.sh

redglobuli:
added starship and wl-copy, starship is configured but not installed, wl-copy is needed for screensnip, also not installed


end-4:
realized wl-copy is the command and wl-clipboard is the real package name lol
but i fixed that

redglobuli:
oh sorry, i confused that too.

PR #147, opened 2023-12-18
Illogical impulse

samvictordr:
Added really basic error handling to the shell script, and edited it to clear the terminal it's running on after every stage of the install.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -19,14 +19,12 @@ echo "sudo usermod -aG video $user"
 sudo usermod -aG video "$user" || echo "failed to add user to video group"
 echo "sudo usermod -aG input $user"
 sudo usermod -aG input "$user" || echo "failed to add user to input group"
-clear
 echo "Step 1 Complete."
 #####################################################################################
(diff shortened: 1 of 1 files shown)


end-4:
Let's not `clear` after every step. 
In case the user changes their mind and wants to go back, they can scroll up and see what changes were made so they can revert them easily

samvictordr:
In that case, I could add a switch to the script that allows reverting (removing all dependencies of hypr along with hyprland) so the user doesn't have to go through the process of finding and removing all the dependencies themselves

end-4 (approved):
alright

end-4:
> In that case, I could add a switch to the script that allows reverting (removing all dependencies of hypr along with hyprland) so the user doesn't have to go through the process of finding and removing all the dependencies themselves

i guess an `uninstall.sh` would be nice

sanhariharan:
HYPRLAND IS AWESOME

end-4:
it's not awesome it's hyprland

PR #148, opened 2023-12-19
Uninstall Script

samvictordr:
Added uninstall script to undo all changes done to the system after running install.sh, removes all installed yay packages and their dependencies, and ags.

yes, it removes the user from video/input groups as well.

### DIFF
--- a/uninstall.sh
+++ b/uninstall.sh
@@ -19,6 +19,7 @@ echo 'User removed from video and input groups.'
 
 # Removing installed yay packages and dependencies
 
-yay -Rns brightnessctl foot fuzzel gjs grim gtk-layer-shell libdbusmenu-gtk3 plasma-browser-integration playerctl python-pywal ripgrep sassc slurp starship swayidle swaylock typescript webp-pixbuf-loader wireplumber wl-clipboard tesseract yad ydotool adw-gtk3-git cava gojq gradience-git gtklock gtklock-playerctl-module gtklock-powerbar-module gtklock-userinfo-module hyprland-git lexend-fonts-git python-material-color-utilities python-poetry python-build swww ttf-material-symbols-variable-git ttf-space-mono-nerd ttf-jetbrains-mono-nerd wlogout
+yay -Rns adw-gtk3-git brightnessctl cava foot fuzzel gjs gojq gradience-git grim gtk-layer-shell gtklock gtklock-playerctl-module gtklock-powerbar-module gtklock-userinfo-module hyprland-git lexend-fonts-git libdbusmenu-gtk3 plasma-browser-integration playerctl python-build python-material-color-utilities python-poetry python-pywal ripgrep sassc slurp starship swayidle swaylock swww tesseract ttf-jetbrains-mono-nerd ttf-material-symbols-variable-git ttf-space-mono-nerd typescript webp-pixbuf-loader wl-clipboard wlogout yad ydotool
+
 
 echo 'Uninstall Complete.'
--- a/uninstall.sh
+++ b/uninstall.sh
@@ -15,7 +15,7 @@ echo 'Removing user from video and input groups and removing packages...'
 user=$(whoami)
 sudo deluser "$user" video
 sudo deluser "$user" input
-echo 'User rmoved from video and input groups.'
+echo 'User removed from video and input groups.'
 
 # Removing installed yay packages and dependencies
 
--- a/uninstall.sh
+++ b/uninstall.sh
@@ -5,10 +5,10 @@ echo 'Removing copied config and local folders...'
 rm -rf "$HOME/.config/ags fish frontconfig foot fuzzel gtklock hypr mpv swaylock wlogout starship.toml"
 rm -rf "$HOME/.local/bin"
 
-# Undo Step 2: Uninstall AGS
-echo 'Uninstalling AGS...'
-sudo meson uninstall -C ~/ags/build
-rm -rf ~/ags
+# Undo Step 2: Uninstall AGS - Disabled for now
+# echo 'Uninstalling AGS...'
+# sudo meson uninstall -C ~/ags/build
+# rm -rf ~/ags
 
 # Undo Step 1: Remove added user from video and input groups and remove yay packages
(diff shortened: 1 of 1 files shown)


uninstall.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+# Undo Step 3: Removing copied config and local folders
+echo 'Removing copied config and local folders...'
+rm -rf "$HOME/.config"
```
end-4: no no no no :skull:

uninstall.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+# Undo Step 3: Removing copied config and local folders
+echo 'Removing copied config and local folders...'
+rm -rf "$HOME/.config"
+rm -rf "$HOME/.local"
+
+# Undo Step 2: Uninstall AGS
+echo 'Uninstalling AGS...'
+sudo meson uninstall -C ~/ags/build
```
end-4: ags isn't here though? 
1. the install script is meant to be run from the repo folder, not home
2. ags is actually something in /usr/bin/ags... but that's just the launcher script. i'm not sure how we can actually get rid of the other libraries

uninstall.sh
```diff
@@ -0,0 +1,24 @@
...
+# Undo Step 1: Remove added user from video and input groups and remove yay packages
+echo 'Removing user from video and input groups and removing packages...'
+user=$(whoami)
+sudo deluser "$user" video
+sudo deluser "$user" input
+echo 'User rmoved from video and input groups.'
+
+# Removing installed yay packages and dependencies
+
+yay -Rns brightnessctl coreutils curl fish foot fuzzel gjs gnome-bluetooth-3.0 gnome-control-center gnome-keyring gobject-introspection grim gtk3 gtk-layer-shell libdbusmenu-gtk3 meson networkmanager nlohmann-json npm plasma-browser-integration playerctl polkit-gnome python-pywal ripgrep sassc slurp starship swayidle swaylock typescript upower xorg-xrandr webp-pixbuf-loader wget wireplumber wl-clipboard tesseract yad ydotool adw-gtk3-git cava gojq gradience-git gtklock gtklock-playerctl-module gtklock-powerbar-module gtklock-userinfo-module hyprland-git lexend-fonts-git python-material-color-utilities python-pywal python-poetry python-build python-pillow swww ttf-material-symbols-variable-git ttf-space-mono-nerd ttf-jetbrains-mono-nerd wlogout
```
end-4: i'm having doubts... this is kinda destructive
i don't think it's a good idea to remove `coreutils upower ripgrep wget`
also remove nlohmann json please, this is no longer needed but i forgot to remove

end-4 (changes requested):

uninstall.sh
samvictordr: Makes sense, some of the packages are included with a gnome install after all

I'll get back to this after testing it on a system with hyprland installed

uninstall.sh
samvictordr: I really didn't think that through lmao

samvictordr: Fixed it

uninstall.sh
samvictordr: I've removed arch-crucial packages and nlohmann-json, after comparing the vanilla arch packages installed with the packages to the removed above (I've removed wireplumber from the list to be uninstalled as well as it's usually installed with pipewire)

uninstall.sh
samvictordr: Ah well, I've not had much experience with ags
We could just comment that out for now, and enable it after we get a viable solution to remove ags in it's entirety. Until then, I'll create a issue for this.

samvictordr (commented):
Done with two requested fixes, and disabled the ags step until a good solution is found to fully uninstall ags

end-4:
- Don't delete the whole `~/.local/bin`, for the same reason as `~/.config`
- And I almost forgot, the script should print the commands that it runs, like `install.sh`. Yknow, in case that huge uninstall command fails (since it might break their other things), the user knows what packages there are to uninstall manually.

samvictordr:
Updated the script, and included a few touches from the original install script as well

uninstall.sh
```diff
@@ -1,24 +1,44 @@
...
+echo '            3. will show all commands that it runs.'
+echo '            4. should be run from its folder.'
+echo 'Ctrl+C to exit. Enter to continue.'
+read
+
+##############################################################################################################################
+
 # Undo Step 3: Removing copied config and local folders
 echo 'Removing copied config and local folders...'
+echo 'rm -rf "$HOME/.config/ags fish frontconfig foot fuzzel gtklock hypr mpv swaylock wlogout starship.toml" '
```
end-4: this syntax won't work
remove each folder in one command

uninstall.sh
```diff
@@ -1,24 +1,44 @@
...
 rm -rf "$HOME/.config/ags fish frontconfig foot fuzzel gtklock hypr mpv swaylock wlogout starship.toml"
-rm -rf "$HOME/.local/bin"
+echo 'rm -rf "$HOME/.local/bin/fuzzel-emoji rubyshot" '
+rm -rf "$HOME/.local/bin/fuzzel-emoji rubyshot"
```
end-4: same here

end-4 (changes requested):

uninstall.sh
end-4: actually, no need to remove each in a command
just have multiple targets in one line but not this way
something like `rm -rf "$HOME/.config/ags" "$HOME/.config/fish"`

uninstall.sh
samvictordr: Fixed with [91dc06d](https://github.com/end-4/dots-hyprland/pull/148/commits/91dc06dd1b7272b0516c86b881ca649a777557d5) as well

uninstall.sh
samvictordr: Fixed with commit [91dc06d](https://github.com/end-4/dots-hyprland/pull/148/commits/91dc06dd1b7272b0516c86b881ca649a777557d5)

Issue #53, opened 2023-08-16
As the framerate size of `scripts/custom_configs/cava` increases, the bottomsheet will crash more frequently and at a faster rate.


TheSonder:
I am unable to find certain useful messages in the eww logs.
When I use `./script/cavajson` in the terminal, everything works fine and it prints numbers correctly without any application crashes.

TheSonder:
No matter what cava configuration file I set, it will cause eww to crash completely.

TheSonder:
I got this 
`Caused by:
    channel closed
 2023-08-16T18:02:23.954Z ERROR eww::error_handling_ctx > sending response from main thread

Caused by:
    0: Failed to send success response from application thread
    1: channel closed
`

end-4:
i think it's just becausse of my eww abuse
too many animated boxes

TheSonder:
How can I solve this issue? Having an i7 12900H processor on my computer is advantageous.

end-4:
im clueless. maybe reduce the number of columns?
edit line 6 of `~/.config/eww/scripts/custom_configs/cava`  (which is currently `bars = 70`)

end-4:
I'm closing this, since I no longer support EWW-powered setups

PR #153, opened 2023-12-28
Add config for swaylock

navidmafi:
Tried to follow colors and general themes used in illogical-impulse
image
image

### DIFF
--- a/.config/swaylock/config
+++ b/.config/swaylock/config
@@ -1,22 +1,22 @@
 ignore-empty-password
-font=monospace
-font-size=80
+font=sans-serif
+font-size=50
 
 clock
 timestr=%R
 #datestr=%a, %e of %B
 
-screenshots
+#screenshots
+color=00000099
 
-fade-in=0.1
-
-effect-blur=20x2
-# effect-greyscale
-effect-scale=0.3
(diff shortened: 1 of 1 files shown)


end-4:
- why is it so huge :skull:
- i'll merge if you 
  - decrease the font-size, indicator-radius and indicator-thickness
  - change the font to sans-serif (monospace is meant to be used in code and terminal and not to be abused everywhere)
  - have the background opaque and not use blur, since it's slow

- i'm still kinda unsatisfied since even swaylock-effects still looks very barebones, but oh well
- thanks for contributing

navidmafi:
thanks for the feedback

> i'm still kinda unsatisfied since even swaylock-effects still looks very barebones

Yes, same here (for anyone reading this in the future: gtklock won't work utill some solution is found for #53 and swaylock is the most "correct" implementation of wayland session locking for now.)

navidmafi:
I agree on blur being slow 
- making it fully opaque would compromise the looks 
- making it partially opaque would compromise the privacy 

I suggest that we should have blur enabled, but you decide.

end-4:
well
i'm going to change that myself

PR #154, opened 2023-12-28
reactor: extract window and layer rules to seperate file

navidmafi:
shortens hyprland.conf


end-4:
okay, but the dynamic colors should be sourced at the bottom so that it overrides defaults

PR #155, opened 2023-12-28
feat: add configurable tray icon size and gap 

navidmafi:


end-4:
no point in adding such a minor configuration. this belongs to the stylesheet. 
the icon should fit in the space and should scale if you change the font size. i'm closing, but i guess i'll make the icons a little larger so that they're more readable

PR #159, opened 2023-12-30
Heavy fix on (un)installation scripts

clsty:
Fixes:
- cd into dir ags before building.
- add ags dir to gitignore.
- other minor fixes.

Features:
- More options during installation.
- Now the script becomes idempotent.


clsty:
NOTE: I've tested the installation script on my own PC, but not the uninstallation script.

end-4:
thanks, this is a lot better.

PR #160, opened 2023-12-30
Fixes on installation script.

clsty:
There's a bug in my previous PR, leading to exiting when ags is not installed.
I'm sorry that I didn't tested out previously.

This PR fixes it.

Also some minor fixes.


end-4:
It looks good to me

clsty:
I'm so sorry that I found more problems during testing.
I'll make a third PR when it's really done.

PR #162, opened 2024-01-01
Supporting multiple monitors

EpixMan:
First this is my first JS contribution so I will definitely appreciate any help

This is simple tinkering that should not affect single monitor users, What makes me a little troubled is my unfamiliarity with JS and there are some random errors are thrown (although even  #Aylur/dotfiles show them so I supposed that it is normal to see those, either due to errors in my device or that they do not cause any side effects such as ram drainage or freezing)

The real problem I am facing is that indicator's widget is not working, tried to make it work on multiple monitors  but without avail. I appreciate any help with that, even so, the problem will only be that the widget will appear in monitor 0 but can be controlled from other monitors


end-4:
- the code looks fine to me on first look, but i'd rather find out if we can do anything about the errors first
> although even #Aylur/dotfiles show them so I supposed that it is normal to see those
- it doesn't for me though? his config throws different errors
- left: aylur's; right: mine with your fixes
image

EpixMan:
> * the code looks fine to me on first look, but i'd rather find out if we can do anything about the errors first
> 
> 
> > although even #Aylur/dotfiles show them so I supposed that it is normal to see those
> 
>     * it doesn't for me though? his config throws different errors
> 
>     * left: aylur's; right: mine with your fixes
>       image

Hey I have updated my system and magically all the bugs disappeared, I am really confused on why and how they did, but it seems to be some backend bugs or so ( I uninstalled swaylock, maybe they are related?)

Anyways, now I am bug free! yay
image

quadbyte:
As a side note I noticed that there are 2 calls to "swww query" which on multi-monitors setup will return one line per monitor.   In the end toggling between light and dark will fail.

I patched those 2 calls (in generate_colors_material.py and switchwall.sh)  by adding "| tail -n 1" at the end  to fix it.  It's fine with me because I use the same wallpaper on both monitors.

end-4:
## About this PR
- My apologies that I actually forgot about this.
- However, I did try to implement multi-monitor support myself and somehow it just worked without returning any errors
  - Here: https://github.com/end-4/dots-hyprland/commit/3f0e064009e0b6943c3effb57ea58b235504389a (i made so many changes until i was comfortable to push so it ended up being a big bad syncing push)

## Wallpaper
- I'm currently trying to implement my own service that handles wallpapers (`swww` is cool, but i wanna add moving animations on workspace change). So I'll add multi-monitor support there.

end-4:
By "worked" I mean it runs normally on my single-monitor system
I think I'll still have to turn instances into callbacks for this to work properly

EpixMan:
> I think I'll still have to turn instances into callbacks for this to work properly

Thank you for helping multi monitor systems users, and sorry for not responding sooner, for some reason, latest branch works flawlessly in single monitor, but crashes within a millisecond after starting and gives this error:
```
 /var/home/first/git/dots-hyprland/.config/ags: > ags -c ./config.js

(com.github.Aylur.ags:80772): Gjs-WARNING **: 00:12:42.717: Unhandled promise rejection. To suppress this warning, add an error handler to your promise chain with .catch() or a try-catch block around your await expression. Stack trace of the failed promise:
  _init/GLib.MainLoop.prototype.runAsync/</<@resource:///org/gnome/gjs/modules/core/overrides/GLib.js:266:34


Gjs-Console-Message: 00:12:42.782: [LOG] Styles loaded

(com.github.Aylur.ags:80772): Gtk-WARNING **: 00:12:42.839: Attempting to add a widget with type Ags_AgsBox to a container of type Ags_AgsBox, but the widget is already inside a container of type Ags_AgsBox, please remove the widget from its existing container first.

(com.github.Aylur.ags:80772): Gtk-WARNING **: 00:12:42.839: Attempting to add a widget with type Ags_AgsBox to a container of type Ags_AgsBox, but the widget is already inside a container of type Ags_AgsBox, please remove the widget from its existing container first.

(com.github.Aylur.ags:80772): Gtk-WARNING **: 00:12:42.885: Attempting to add a widget with type Ags_AgsBox to a container of type Ags_AgsRevealer, but the widget is already inside a container of type Ags_AgsRevealer, please remove the widget from its existing container first.

(com.github.Aylur.ags:80772): Gtk-WARNING **: 00:12:42.886: Attempting to add a widget with type Ags_AgsBox to a container of type Ags_AgsBox, but the widget is already inside a container of type Ags_AgsBox, please remove the widget from its existing container first.

(com.github.Aylur.ags:80772): Gjs-Console-CRITICAL **: 00:12:43.173: Error: There is already a window nameddesktopbackground
addWindow@resource:///com/github/Aylur/ags/app.js:123:27
_load@resource:///com/github/Aylur/ags/app.js:156:36
async*vfunc_activate@resource:///com/github/Aylur/ags/app.js:70:14
_init/GLib.MainLoop.prototype.runAsync/</<@resource:///org/gnome/gjs/modules/core/overrides/GLib.js:266:34

```

I will update what might be the cause but I might need sometime to test every change to debug what is the cause

end-4:
oooh right forgot the naming of that one

EpixMan:
Hey sorry I might have made a mistake and closed it..

**anyways here are my points**
```
forMonitors(Indicator), //this is for sorry useless, not working as intended, but I think it is ignorable
```
```
forMonitors(Bar), this needs to be edited to make the bar launch on every screen
```

Forgive my unfamiliarity, but I really do not think this is necessary at all:
```
const OptionalWorkspaces = async () => {
    try {
        return (await import('./workspaces_hyprland.js')).default();
    } catch {
        // return (await import('./workspaces_sway.js')).default();
        return null;
    }
};
const optionalWorkspacesInstance = await OptionalWorkspaces();
```

as it was actually breaking the hyprland workspace widget on the second monitor, is it required to run ags on sway or something like that?

end-4:
> this is for sorry useless, not working as intended

yea like i said i have to change stuff from instances to callbacks which i haven't

> I really do not think this is necessary at all:

i tried to add sway compatibility. with this kind of try-catch, the setup shouldn't crash if ran on sway

EpixMan:
Ok I finally made a normal PR, can you check it please?
I tried to be creative with the importing stuff but it always fails. I think setting which WM to be set up in install.sh might be a somewhat more logical for me

clsty:
> I tried to be creative with the importing stuff but it always fails. I think setting which WM to be set up in install.sh might be a somewhat more logical for me

I do understand your point, however I also think that 2 WMs together could be a good way to switch between "performance mode" (sway) and "nice-GUI mode" (hyprland).

> The problem is that it is breaking the hyprland widget in the second monitor, I tried to solve it but couldn't make anything work other than importing at the top

Maybe we could use another minimal and lightweight bar for sway, and at the same time we still use ags for hyprland, but no need to consider its capability for sway.

clsty:
> * I don't wanna use another config, since I want my experience to be consistent.

Just in case, have you been using sway? If not, then let me friendly notice you that sway actually [does not support animation](https://github.com/swaywm/sway/issues/3380#issuecomment-451760626).
Of course, it's up to you whether animation matters.

clsty:
Thank you for submitting the PR.

The `illogical-impulse` branch is merged to `main` branch, and will be deleted soon, see <https://github.com/end-4/dots-hyprland/pull/262> for more.

Please target this PR to the `main` branch instead.

Refer to <https://stackoverflow.com/questions/24159036/how-to-modify-a-pull-request-on-github-to-change-target-branch-to-merge-into> for instruction.

PR #165, opened 2024-01-02
Remove execute permission for non-executables.

clsty:
`chmod -x <file>` for all files that definitely not need execute permissions.


end-4:
:skull:

PR #170, opened 2024-01-03
Improvement on scripts

clsty:
See the titles of commits.


clsty:
Notice that now `./install.sh -f` do things automatically without confirm.
> That's pretty handy for me when I'm debugging. i.e. After I edited the repo, I use the script with `-f` and it finish in seconds.

clsty:
About "Read dependencies from a file":
Now each line will be asked during installation.

Packages in `dependencies.txt` can be splited by spaces or returns.
But comments in this file is currently not supported.

clsty:
@end-4 Any comment on this PR is welcomed. 
Please let me know if there's anything not good enough on this PR.

install.sh
```diff
@@ -4,23 +4,23 @@ export base="$(pwd)"
...
   execute=true
   if $ask;then
     while true;do
-      echo -e "\e[34mDo you want to execute this command? \e[0m"
+      echo -e "\e[34mDo you want to execute the command shown above? \e[0m"
       echo "  y = Yes"
-      echo "  a = Abort this script"
-      echo "  s = Skip this command; NOT recommended unless you really sure"
+      echo "  e = Exit now"
+      echo "  s = Skip this command; NOT recommended (may break functions needed by the dot files!)"
```
end-4: "dotfiles" not "dot files"

install.sh
```diff
@@ -29,26 +29,58 @@ function v() {
...
-printf "\e[36m== PLEASE BACKUP \"$HOME/.config\" AND \"$HOME/.local\" BY YOURSELF IF NEEDED! ==\n\e[97m"
+printf '\n'
+printf "\e[31m"
+printf "=============\n"
+printf "PLEASE CONFIRM\n"
+printf "YOU HAVE\n"
+printf "  \"$HOME/.config\"\n"
+printf "  AND\n"
+printf "  \"$HOME/.local\"\n"
+printf "ALREADY BACKUPED\n"
```
end-4: - "backed up" not "backuped"
- Also I don't think it's necessary to have this on so many lines. A non-white color that stands out should be enough.
- So I'd rather have something like this:
```
Please CONFIRM that you HAVE ALREADY BACKED UP $HOME/.config and $HOME/.local folders
Enter "YES" (in capitals, without quotes) to continue: 
```

end-4 (changes requested):

end-4:
finally have some time to review this, sorry for the slow response

clsty:
Thanks for reviewing. Is that better now?

end-4 (approved):
alright

PR #176, opened 2024-01-06
Fix and improvement on installation script.

clsty:
1. Remove execute permission for non-executable

> I noticed them because I'm using vifm. Every file with execute permission will be highlighted with green color (when using dracula colorscheme).

2. Fix hyprland reporting error because `colors.conf` does not exists.

3. Better error handling. Not abruptly `exit 1` as before, user will now be informed of the failed command, and be provided with an option to repeat this command (default), ignore this failure or exit. Also in the end print info about the status of command (finished, failed, ignored or skipped).

4. Better ags installation process. Now when `$ask == true` internally, user can choose to update to the latest version of ags even when the command `ags` already exists.

5. Add `install-rubik` (similar to `install-ags`) for the commit 06a995427f3b626fd41d8e96a7abc68d4d4dfba4 .

> I've tried AUR package `ttf-rubik`, but it downloads font files from google, which domain is restricted in some regions. For those who already have Rubik installed, the script will detect the font and then prompt option for user when `$ask == true`, or skip to next when `$ask == false`.


clsty:
That's another several hours spent by me today... 
Finally, I can't see more things to do on this PR now.

Please review this PR when you have time and feel like it, thanks! @end-4

install.sh
```diff
@@ -77,27 +112,49 @@ v sudo usermod -aG video "$(whoami)"
 v sudo usermod -aG input "$(whoami)"
 
 #####################################################################################
-printf '\e[36m2. Installing AGS from git repo\e[97m\n'
+printf '\e[36m2. Installing AGS and rubik from git repo\e[97m\n'
```
end-4: capitalize R

install.sh
```diff
@@ -77,27 +112,49 @@ v sudo usermod -aG video "$(whoami)"
...
+  cd $base/ags
+  try git init -b main
+  try git remote add origin https://github.com/Aylur/ags.git
+  git pull origin main && git submodule update --init --recursive
+  npm install
+  meson setup build 
+  meson install -C build
+  cd $base
+}
+install-rubik (){
```
end-4: - If you're going to handle font installation, it'd be nice if you could make it install the font "Gabarito" too
  - It's used in title texts
  - I think it's here https://github.com/naipefoundry/gabarito
  - But it's up to you. I'll let you decide whether or not to add this, no pressure

install.sh
```diff
@@ -7,27 +7,62 @@ function v() {
...
   if $execute;then
-  "$@"
+    "$@" || cmdstatus=1
   fi
+  while [ $cmdstatus == 1 ] ;do
+    echo -e "\e[31m[$0]: Command \"\e[32m$@\e[31m\" has failed."
+    echo -e "You may need to resolve the problem manually BEFORE repeating this command.\e[0m"
+    echo "  r = Repeat this command (DEFAULT)"
+    echo "  e = Exit now"
+    echo "  i = Ignore the error and continue this script anyway (may break functions needed by the dotfiles!)"
```
end-4: I think this would sound more natural for option `i`:
```
Ignore this error and continue (your setup might not work correctly)
```

end-4 (commented):
- I must thank you for your time and effort put into this, and I'm sorry to hear about those restrictions.
- I might be a little nitpicky but since you're really paying attention to detail I figured I'd point out those little wording choices :p
- I can't say for sure since my bash scripting skills are far from this level, but at least I don't find anything seriously wrong. Have you tested this? I don't have a 2nd machine to test, and Hyprland just refuses to work in a VM...

install.sh
clsty: Done.

clsty:
> I might be a little nitpicky but since you're really paying attention to detail I figured I'd point out those little wording choices :p

It's totally OK, English is not my native language, I'm grateful about you read them so carefully and help fix the problems I didn't find out!

> I can't say for sure since my bash scripting skills are far from this level, but at least I don't find anything seriously wrong. Have you tested this? I don't have a 2nd machine to test, and Hyprland just refuses to work in a VM...

As I said before, the script is **idempotent**, which basically means you can run it as many times as you want. So yes, I have tested the script many times.

clsty:
Oh, and one more thing:

the `install-*` (except for yay) here depend on the folder structure of the git repos.

If the git repo upstream changes, the build process will possibly fail.

To prevent this thing from happen, you may fork the repos as backup, and optionally let the install script to use your forked repo as source.

clsty:
I guess that's it for now. Please have a look at the new changes when you feel like it!
@end-4

end-4:
awesome

PR #178, opened 2024-01-07
fix permissions.

clsty:
It's not a serious problem, but non-executables just should not have execute permission...

> I'm wondering that why you frequently make non-executables with execute permission.
> Maybe check your editor (or IDE)?


end-4:
idk lol

end-4:
could be because the way i create a new file is copying another file to the place i want then modify it
which often happens to be a shell script since they're everywhere :skull:

PR #187, opened 2024-01-15
dependencies.txt

Aman49362:


end-4:
eww no

end-4:
closing a pr/issue with only 2 words doesn't sound that cool, so i'm going to provide a short explanation:
- thee main reason: those packages have nothing to do with my dotfiles. there's no keybind to launch them or anything.

- more personally: i don't want them, because:
  - flatpak? i'd rather not have duplicate drivers which will result in many extra gigabytes taken up
  - discord? i'd rather not have an electron app (200mb at least (data included)) just to do what my browser already does

PR #199, opened 2024-01-20
Fix CJK copied by Ctrl+C became \uxxxx

clsty:
When content copied by Ctrl+C , CJK character becomes `\uxxxx` in `cliphist list`, which reason is actually inside the wl-paste.

You can test this out by adding following to `keybinds.conf` in `~/.config/hypr`:
```conf
bind=,F11,exec, pkill wl-paste && wl-paste --type text --watch cliphist store & wl-paste --type image --watch cliphist store & wl-paste -p --watch wl-copy -p ""
bind=,F12,exec, pkill wl-paste && wl-paste --watch cliphist store & wl-paste -p --watch wl-copy -p ""
```
When you use `F12`, and `Ctrl`+`C` copy this text: `test 测试`, then `cliphist list` in terminal, it becomes `test \u6d4b\u8bd5`.

When you use `F11` (corresponding to the fix included in this PR), and `Ctrl`+`C` copy this text: `test 测试`, then `cliphist list` in terminal, it becomes `test 测试`.

### DIFF
--- a/.config/hypr/execs.conf
+++ b/.config/hypr/execs.conf
@@ -15,7 +15,8 @@ exec-once = swayidle -w timeout 450 'pidof java || systemctl suspend' &
 exec-once = sleep 1 && dbus-update-activation-environment --systemd WAYLAND_DISPLAY XDG_CURRENT_DESKTOP # Some fix idk
 
 # Clipboard: history
-exec-once = wl-paste --watch cliphist store &
+exec-once = wl-paste --type text --watch cliphist store
+exec-once = wl-paste --type image --watch cliphist store
 # Clipboard: disable middle-click paste
 exec-once = wl-paste -p --watch wl-copy -p ""
 


clsty:
In case your system can't display CJK character, you can install a font eg:
```bash
sudo pacman -S wqy-microhei
```
And uninstall the font after testing:
```bash
sudo pacman -R wqy-microhei
```

clsty:
And, (sorry it's out of topic), I still don't see my ID listed in `Contributor` on the sidebar of this project.
Maybe you could include me there to give me some credits? Thanks!

end-4:
## CJK problem
- Currently those characters load just fine for me though? Even I'm not in East Asia? Maybe it's some EndeavourOS magic that I never had to take into account.
image

> I still don't see my ID listed in Contributor on the sidebar of this project.
- Unfortunately GitHub doesn't count anything not committed to the main branch. Idk... should I make illogical-impulse the main one? It just feels wrong because the main readme is supposed to show other stuff too...
- Don't get me wrong - I highly appreciate your work. Yknow what? I'll just put you in the acknowledgements section of the readme, if you think that's good enough

clsty:
# CJK problem

Hmmm, it seems really weird... But I tested it again, and this time `F11` works normal as before, while `F12` gives an even weirder result: `test $(A2bJT` (I paste it here using that cliphist widget).

`cliphist list` display it as `test 2bJT`:

clsty:
I can't tell where the problem locates. But if this PR does not break clipboard history functionality on your system (EndeavourOS), and does fix for the problem appeared on my system (Arch Linux), then I think it's ok to be merged, right?

end-4:
sure i guess
extra 131.1kb ram usage is no big deal

PR #209, opened 2024-01-27
New workflow: Supporting bar and overview for Workspaces > 10

midn8hustlr:
First attempt at JS ... wish me luck :(

This feature will be useful for:
- Users like me who shamelessly opens dozen of workspaces and wants a non-cluttered way of navigating through them. 
- Users having multiple monitors and want to have awesome-wm like experience, where each monitors will have its own set of 10 workspaces.
- Multitaskers who want to context-switch to a different task altogether (like office or personal) and don't want to clutter apps or rearrange workspaces.

This will not affect any functionality for users having workspace <= 10

Description:
- The workspaces will be grouped in a set of 10. If we have 30 workspaces then they will be divided into 3 groups (1-10, 11-20 and 21-30). The current focused group will depend upon the active workspace. If active ws is in between 1-10, group 1 (ws 1-10) will be focused, if active ws is in between 11-20, then group 2 will be on focus, and so on.
- The workspace panel in the bar (showing occupied or active status) will change based upon the focused group/ active workspace. Eg. if focused group is 1, it will show the status for workspaces 1-10. Similarly, for focused group being 2, the workspace panel will show the status for 11-20. For identifying the focused workspace group, the ' • ' decoration is replaced with the current workspace ID. I didn't like the workspace panel to be cluttered with 2-digit numbers. 
- The overview panel is also affected in the same way.
- Added support for navigating and moving workspaces in hypr keybinds. Eg. pressing Super+3 will move you to workspace-3 if current workspace is between 1-10, or workspace-13 if current workspace is in between 11-20.
- As of now it supports till 30 workspaces (3 groups of 10 each) but this can be increased by modifying NUM_WS_GROUPS in workspaces_hyprland.js and overview_hyprland.js 

Video:
https://github.com/end-4/dots-hyprland/assets/47367711/a0fc9590-a0bb-409f-ade2-52aad1b601a1

Thanks

### DIFF
--- a/.config/hypr/workspace_action.sh
+++ b/.config/hypr/workspace_action.sh
@@ -1,0 +1,2 @@
+#!/bin/bash
+hyprctl dispatch $1 $((((`hyprctl activeworkspace -j | jq -r .id` - 1)  / 10) * 10 + $2))
--- a/.config/ags/widgets/bar/workspaces_hyprland.js
+++ b/.config/ags/widgets/bar/workspaces_hyprland.js
@@ -8,13 +8,14 @@ import Widget from 'resource:///com/github/Aylur/ags/widget.js';
 const { Box, DrawingArea, EventBox } = Widget;
 import Hyprland from 'resource:///com/github/Aylur/ags/service/hyprland.js';
 
-const NUM_OF_WORKSPACES = 10;
+const NUM_WS_GROUPS = 3;
+const NUM_OF_WORKSPACES_PER_GROUP = 10;
 const dummyWs = Box({ className: 'bar-ws' }); // Not shown. Only for getting size props
 const dummyActiveWs = Box({ className: 'bar-ws bar-ws-active' }); // Not shown. Only for getting size props
 const dummyOccupiedWs = Box({ className: 'bar-ws bar-ws-occupied' }); // Not shown. Only for getting size props
 
 // Font size = workspace id
-const WorkspaceContents = (count = 10) => {
+const WorkspaceContents = (offset = 0, count = 10) => {
     return DrawingArea({
         css: `transition: 90ms cubic-bezier(0.1, 1, 0, 1);`,
         attribute: {
@@ -26,8 +27,8 @@ const WorkspaceContents = (count = 10) => {
                 let workspaceMask = 0;
                 for (let i = 0; i < workspaces.length; i++) {
                     const ws = workspaces[i];
-                    if (ws.id <= 0) continue; // Ignore scratchpads
-                    if (ws.id > count) return; // Not rendered
+                    if (ws.id <= offset) continue; // Ignore scratchpads
+                    if (ws.id > offset + count) continue; // Not rendered
                     if (workspaces[i].windows > 0) {
                         workspaceMask |= (1 << ws.id);
                     }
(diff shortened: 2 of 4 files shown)


.config/hypr/workspace_action.sh
clsty, line 1: Since you're using `jq` here, you'd better add `jq` to `scriptdata/dependencies.txt`.

clsty (commented):
Personally I like this new workflow, especially about multitasks support.

.config/hypr/workspace_action.sh
end-4: use `gojq`, since it's already in dependencies.txt

end-4:
This is the kind of innovation I always strive for. Very damn cool!

clsty:
I've tested this on my computer. But I still don't get it well, like, how do you go across different groups?

I see that in your video, you jumped from WS3 to WS16, but what key did you press to do that?

Or, if there are no windows on WS11 - WS20 (empty group), how will you quickly add a WS or a window into that group?

end-4:
Huh what happened
I used gh pr checkout and pushed
I thought that would edit this pr but it ended up merged
Anyway, thanks for the contribution!

.config/hypr/workspace_action.sh
clsty: > use `gojq`, since it's already in dependencies.txt

`gojq` seems not providing command `jq`. And I don't know if their usage are same. Maybe need some tweaks to migrate to `gojq`?

midn8hustlr:
> But I still don't get it well, like, how do you go across different groups?
>
This is mostly to not limit the workspaces to 10. If all of your workspaces are already occupied, and you are on workspace 10, you can just can move to the next workspace using hypr keybinds or scrolling the workspace bar. This will spawn workspace 11 and take you to the next group, while the previous group will be hidden.

However, my usecase is different as I have multiple monitors and I bind workspaces 11-20 to the secondary monitor using bind in hypr config (see [this](https://github.com/sopa0/hyprsome)). This will help me get the status of workspaces in the secondary monitor in the bar. Also I will have separate overview panels for all the monitors. 

However in multi-monitor setup, the top bars in both monitors will always get updated based upon the current workspace group (would be much intuitive if 1st bar shows ws 1-10 and 2nd bar shows ws 11-20). But I dont use the 2nd bar, I always disable the bar on my 2nd monitor, as most of it has the same info as primary bar. Also, implementing this functionality will also cause some weird behavior for single monitor users and limit them to 10 workspaces again. So, I wanted to keep it simple and suit most of the users.

If you are a multi-monitor user, on startup you can manually move the 2nd workspace (on monitors 2) to 11th workspace or use workspace bind to start workspace 11 on monitor 2.

For moving workspaces between monitors using overview, few improvements can be done in overview panel. Or, just focus on two workspaces in both monitors and move/swap windows using keybinds.

> I see that in your video, you jumped from WS3 to WS16, but what key did you press to do that?
> Or, if there are no windows on WS11 - WS20 (empty group), how will you quickly add a WS or a window into that group?
> 
Apart from whatever modified in PR.  I use the following hypr keybinds
```
$mainMod = SUPER
bind = $mainMod, mouse_down, workspace, m+1
bind = $mainMod, mouse_up, workspace, m-1
bind = $mainMod, Z, workspace, m-1
bind = $mainMod, X, workspace, m+1
bind = $mainMod CONTROL, Z, workspace, r-1
bind = $mainMod CONTROL, X, workspace, r+1
bind = $mainMod ALT, Z, workspace, r-10
bind = $mainMod ALT, X, workspace, r+10
```
To navigate from WS3 to WS16, i used SUPER + Z/X which will only move to occupied workspace on the current monitor.
To directly move to the next WS group, I use SUPER ALT + Z/X, which will jump by a factor of 10. I didn't put these in PR as it will mess up with end4's messed up windoes keybinds.

PR #210, opened 2024-01-28
Dynamically change all opened terminals colorscheme with wallpaper

midn8hustlr:
No need to relaunch opened terminals after applying colorscheme.

Video:

https://github.com/end-4/dots-hyprland/assets/47367711/3035ef13-fcd7-4f81-9310-e05f1470c36e


end-4:
cool

PR #214, opened 2024-01-29
Update issue template

clsty:
- Linux distro matters sometimes.
- The log may very long sometimes.


end-4:
i don't know if i can support other distros, but fine

Issue #225, opened 2024-02-01
[Issue] awk: cmd. line:1: (FILENAME=- FNR=3) fatal: division by zero attempted


loeken:
- **I have read the instructions**
---
## The issue
- Linux distro: manjaro
- in output of "journalctl -f" i can see a lot of those error messages
- I`m using gdm
<details>
<summary>Click to expand</summary>

```plain
Feb 01 13:37:04 framwork /usr/lib/gdm-wayland-session[1426]: awk: cmd. line:1: (FILENAME=- FNR=3) fatal: division by zero attempted
Feb 01 13:37:09 framwork /usr/lib/gdm-wayland-session[1426]: awk: cmd. line:1: (FILENAME=- FNR=3) fatal: division by zero attempted
Feb 01 13:37:14 framwork /usr/lib/gdm-wayland-session[1426]: awk: cmd. line:1: (FILENAME=- FNR=3) fatal: division by zero attempted
Feb 01 13:37:19 framwork /usr/lib/gdm-wayland-session[1426]: awk: cmd. line:1: (FILENAME=- FNR=3) fatal: division by zero attempted
Feb 01 13:37:24 framwork /usr/lib/gdm-wayland-session[1426]: awk: cmd. line:1: (FILENAME=- FNR=3) fatal: division by zero attempted
```

</details>

loeken:
correct me if im wrong but i believe this comes from some part of the bar ( when i comment out [the ](https://github.com/end-4/dots-hyprland/blob/2b241a16ddfc2a1cf914bc91e6d5f2b4dfff28d5/.config/ags/config.js#L68) ) the error messages disappear

end-4:
Do you have swap or zram? i recommend setting up the latter
My guess is that the swap memory indicator threw that when attempting to get the percentage by calculating `used/total*100`

loeken:
I have neither, ( also havent heard of zram before )  I would rather like my applications to crash then swap to disk and i have plenty of ram so i dont want to use compression in ram (if i understand zrams purposes correctly )

loeken:
could we add an option to check if total not null ? and can you point me to where this is calculated at?

loeken:
i raised a small pr to fix it in the 2 locations ive found this removes the logspam for me

loeken:
thanks

PR #227, opened 2024-02-01
use case no swap, no zram

loeken:
https://github.com/end-4/dots-hyprland/issues/225

this checks if swap is null and avoids dividing by zero


end-4:
yes

Issue #230, opened 2024-02-02
color picker not working properly


Iridule:
I have multiple permission denied trying to run 
` .config/ags/scripts/color_generation/switchcolor.sh`...choosing background works but theme will not change and color picker does not work ...
```shell
Permission denied
sed: couldn't open temporary file /home/iridule/.config/fuzzel/sed5iaJqU: Permission denied
sed: couldn't open temporary file scripts/templates/gradience/sedoe1p8G: Permission denied
sed: couldn't open temporary file /home/iridule/.config/fuzzel/sed2Mbc3U: Permission denied
sed: couldn't open temporary file scripts/templates/gradience/sedRLMUDp: Permission denied
sed: couldn't open temporary file scripts/templates/gradience/sed58SmE9: Permission denied
sed: couldn't open temporary file /home/iridule/.config/fuzzel/sedpCv
```

clsty:
I guess it's not a issue related to this repo...
Please refer to <https://wiki.archlinux.org/title/Users_and_groups#Permissions_and_ownership> to fix those permissions first.

Iridule:
@end-4 chown fixed it. Color picker works now. Thank you.

clsty:
Welp, since `chown` fixed the problem, it's very possible that you have messed up with permission and ownership somewhere... 

E.g. maybe you have misused `sudo` (which can be fixed by `chown`) or abuse `chmod 777` (which can be fixed somehow but is hard to entirely fix).

This can cause many **weird** problems in your system, not only for setup of Hyprland.

And, for color picker in AGS bar... Again, you need to provide more info about it (such as run `pkill ags;ags` and paste output if there're any errors), not just say "not working" because hence its reason could be anything and no one could figure out what's happening.

Iridule:
not sure how permissions would be messed with this is a new install of Arch I only have used it to test install Hyprland I only have used sudo a few times when installing some things with pacman to get the system usable. 
`pkill ags;ags` (seems to be no errors...

Gjs-Console-Message: 10:07:11.407: [LOG] Styles loaded

clsty:
> seems to be no errors

And what happened when you click color picker on AGS bar? such as new output

Iridule:
@clsty there was no output

clsty:
You may fix those permissions from paccheck, but it may not related to the color picker problem.

Iridule:
[...]

Issue #222, opened 2024-01-31
[Issue] GTK_IM_Module set?


Iridule:
I get a Wayland diagnostic notification saying "GTK_IM_MODULE being set and Wayland input method frontend is working. It is recommended to unset GTK_IM_MODULE."

I've tried adding the lines below to ~/.bash_profile but same notification appears after login. 
export GTK_IM_MODULE=fcitx
export QT_IM_MODULE=fcitx
export XMODIFIERS=@im=fcitx

clsty:
What's your output when run
```bash
echo $GTK_IM_MODULE
```
?

Also please check `~/.config/hypr/env.conf`, where environment variables including `GTK_IM_MODULE` are defined.

clsty:
I don't know but maybe you're using ibus and not fcitx5?

Iridule:
I'm at a loss on what to do or how it impacts my setup. I've noticed some times of an app like vlc is open the keyboard won't work and the right mouse button doesn't work...not sure if these things are related to the diagnostic message or not

clsty:
I guess you need to install the input method first, since they are not listed in `scriptdata/dependencies.txt`:
```bash
sudo pacman -S fcitx5{,-{qt,lua,rime,configtool,gtk}} ibus
```

Or, if you don't need input method, you can just remove the related lines in `~/.config/hypr/env.conf` and `~/.config/hypr/execs.conf`.

Iridule:
@clsty I installed those packages yet the issue persists....
![IMG_20240131_211013397~2](https://github.com/end-4/dots-hyprland/assets/135180769/d9dcf0de-e9ef-4d58-87ed-5abf1ab153e3)

clsty:
From <https://fcitx-im.org/wiki/Using_Fcitx_5_on_Wayland>:

> For GTK_IM_MODULE, as for now, the modern Gtk 3/4 application should be able to use text-input-v3 which is supported almost by every compositor, except weston. There are a few different options in terms of setting the value of GTK_IM_MODULE. When it is unset, the Gtk built-in Wayland im module will be used for Gtk3 and Gtk4. While you can also enforce it by doing GTK_IM_MODULE=wayland, remember it will be also picked up by Gtk2. Setting GTK_IM_MODULE=fcitx can still work, and it is necessary if your compositor does not support Wayland input method frontend.

clsty:
[...]

PR #233, opened 2024-02-03
Bash-style comments for dependencies list, other fixes

clsty:
- Comments for `scriptdata/dependencies.txt` is implemented in Bash-style, and `.txt` renamed to `.conf` for better highlighting.
Hopefully this will make that file more readable and maintainable.

- I've noticed that you have 3 duplicated lines `.config/ags/style.css` in `.gitignore`, so I think you actually want to use `git rm <file>` to let the ignore rules take effect. So, I did it for you.

- As for `colors.conf`, since you're not having `colors_default.conf` file anymore in this repo, I commented it out from `.gitignore` and also commented out the related lines in `install.sh`. If you're sure to add them to `.gitignore` (NOT recommended because this will make hyprland report error since there's no `colors.conf` present after running `install.sh`), you need `git rm <file>` again.
  - I'm commenting them out instead of delete them directly, because I'm not entirely sure and I'll let you decide eventually.

- In order to fix the problems like #222, some environment vars are added/modified.
  - Reference 1: <https://fcitx-im.org/wiki/Using_Fcitx_5_on_Wayland>
  - Reference 2: <https://wiki.archlinux.org/title/Fcitx5#Integration>
  - Reference 3: <https://github.com/end-4/dots-hyprland/issues/222#issuecomment-1920398154> (which includes the snippet to detect the problem inside fcitx5 repo).

- To prevent misuse of `sudo` like #230 , `install.sh` will detect it at first.


end-4:
tested, works

end-4:
i realized `env = GTK_IM_MODULE, wayland` would crash electron apps in xwayland

PR #234, opened 2024-02-03
On-Screen-Keyboard improvements

Motschen:
- Pressing shift will now cycle between shift, caps lock, and lowercase
- Visual representation of uppercase letters/symbols, as well as AltGr for German keyboards
  - You can also add the equivalent Alt keys for QWERTY keyboards, I just don't know what they'd be
- Added German QUERTZ virtual keyboard layout

### DIFF
--- a/.config/ags/data/keyboardlayouts.js
+++ b/.config/ags/data/keyboardlayouts.js
@@ -71,7 +71,9 @@ export const oskLayouts = {
                 { keytype: "normal", label: "\\", labelShift: "|", shape: "expand", keycode: 43 }
             ],
             [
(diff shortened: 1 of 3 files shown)


end-4:
cool

end-4:
any idea for that empty capslock? :)

Motschen:
> any idea for that empty capslock? :)

Maybe the left and right arrow keys?
That'd allow for easier navigation in texts.
Though idk if the position would be fitting.

kxrur:
Is this supposed to work for the non-qwerty characters such as `ü`? 
ydotool types `[` when I click that key.

I am trying to add a cyrillic layout but am having a hard time making ydotool work with a different layout.

PR #237, opened 2024-02-04
Add a sidebar assistant based on Ollama

Motschen:
- Run a multitude of AI models locally using Ollama
- Change model using /model <NAME>
- List models using /list
- Start/Stop ollama server using /start and /stop


end-4:
## Chat
My machine is too slow and the connection always times out before a response is given 💀
Maybe try to set the timeout to 1 minute or something? 😅 The Soup session init should look like this ([reference](https://gjs-docs.gnome.org/soup30~3.0/soup.session#property-timeout)):
```js
const session = new Soup.Session({
    timeout: 60, // Seconds
});
```
## User experience
- Currently, model selection is treated similar to how API keys in other models are treated. However, it isn't as clear since when there isn't yet a model, 
  - The message that tells you to setup ollama doesn't show
  - The prompt of the text entry doesn't tell you to choose a model (although the code does seem to do that)
  - *And when the user types in a message, the thing assumes that should be the model name and sets it*

Motschen:
Thank you very much for the feedback, I'll try to address soon.

clsty:
Thank you for submitting the PR.

The `illogical-impulse` branch is merged to `main` branch, and will be deleted soon, see <https://github.com/end-4/dots-hyprland/pull/262> for more.

Please target this PR to the `main` branch instead.

Refer to <https://stackoverflow.com/questions/24159036/how-to-modify-a-pull-request-on-github-to-change-target-branch-to-merge-into> for instruction.

end-4:
huh i changed branch and it closed automatically... but feel free to submit a new pr if you still wanna contribute

end-4:
a

PR #247, opened 2024-02-08
Fix workspace panel in bar not updating when directly moving to occupied workspace in the next group

midn8hustlr:
When I am using keybinds

bind = SUPER, Z, workspace, m-1
bind = SUPER, X, workspace, m+1

which switches between the active workspaces only. This does not trigger 'notify::workspaces' hook from ags/services/hyprland, since no new workspace is created or destroyed. So the workspace group is not updated in the bar and occupied masks are shown for previous workspace group.

Better to add a hook when the group is changed based upon the active workspace.


midn8hustlr:
Also this is a pain in case you have allotted different groups to different monitors and switching between monitors does not updates the workspace masks.

end-4:
lgtm
i gotta apply this to overview too hmm

PR #254, opened 2024-02-12
Update issue template, add link to wiki

clsty:
Add link to wiki and minor improvements.


clsty:
Oh, one more thing. Since we're adding multi-language for the wiki, there may be people submitting language other than English.
This is not expected behavior, right? Therefore, I suggest we add a comment to the template, such as:
```markdown
Please use English language for writing issue.
To get output in English, use `LANG=C ` before a command, e.g. `LANG=C date`.
```

end-4:
Yes you should add that comment

clsty:
> Yes you should add that comment

Sure, it's done.

PR #264, opened 2024-02-16
Add notification about using main branch.

clsty:
The illogical-impulse branch is migrating to main branch, and the old branch will certainly stop updating, but the existing users will know nothing about it, even if they run `git pull` as documented for updating.

This PR will make a notification for them, and also provide them commands to execute.


end-4:
hmm yes

PR #273, opened 2024-02-20
Fix: Change swaylock to hyprlock in session window

Ninjdai1:
Change swaylock to hyprlock in session window


clsty:
I don't see any reason to switch from swaylock.

hyprlock is not a built-in program of Hyprland, it's a new project and it even did not get into AUR until yesterday (or the day before?).

In short, I don't see any "fix" here, at least for now.

Ninjdai1:
Hi,
Swayidle/lock were replaced by hypr's in [7a4a61f](https://github.com/end-4/dots-hyprland/commit/7a4a61fb00fa9fc5be5dd99302b5e26620254451), so I thought it missing from the session window was a mistake, that's why I labeled it as a fix
No worries :p

end-4:
alright

PR #293, opened 2024-02-29
Global colorscheme for all terminals and transperancy effects

midn8hustlr:
This will not affect any existing functionality, but will enable users to use any terminal or shell.
Some unnecessary files are also removed.

Problem:
1. Some people might prefer to use alternative to foot terminal, like alacritty/wezterm.
2. Many of us might not prefer to use fish shell either and choose alternatives like bash, dash or zsh.
3. Configs like foot.ini can be customized by user, but everytime a new theme is applied it is overwritten with the template config.
4. For peoples using programs like tmux, the proper colorscheme might not be applied dynamically if colorscheme is pre-configured using a terminal config (ig some bug with foot terminal, for kitty it worked fine).
5. I miss terminal transparency+blur :(

Solution:
1. Do not use configuration template for any specific terminal to apply colorscheme, instead make it generic for all terminals by using OSC11 escape sequences via any shell like fish/zsh. This functionality was there already implemented in #210, but somehow apply_foot() in applycolor.sh reappeared in later commits. apply_foot() is unnecessary and causes problem 3 and 4. 
2. Added  transparent mode for any terminal by setting  `term_alpha`  less than 100 in  `.config/ags/scripts/color_generation/applycolor.sh`.


.config/fish/config.fish
midn8hustlr, line 1: We should not store sequences.txt in fish, as it is not specific to fish shell and can be used by any other shell.

end-4: that's fair

PR #308, opened 2024-03-07
Toggling refresh rate on the basis of battery status.

harshafaik:
A simple script to automatically switch your refresh rate on the basis of your battery status.  Install acpi before using the script. Change values according to your system.


end-4:
This script will only work for your device, not mine or others'. No, thanks

PR #311, opened 2024-03-07
Toggling refresh rate on the basis of battery status.

harshafaik:
Posted before but deleted as it was not generalized (sorry for that :pray:)

Configure according to your system, install acpi before executing the script.
Add it later as an exec-once instance in your execs.conf.


end-4:
I'm not interested and will not add this.

- I use a 1920×1080 screen at 60Hz
- This runs constantly without an interval
- "configure according to your system" *but will break on others* is not something I want.

If I were to make this,
- I'd make it auto-detect
- There's a more elegant solution with agsa

Thank you, but keep the script to yourself.

harshafaik:
Alright no worries.

PR #320, opened 2024-03-10
Update spaceleft.js

arcatva:
Bug:
The Hyprland.active.client.title can be too long to attach closely to the next widget (normally when a browser's tab has too long title).

I just change the substring the long title and let it end with "..."

Should be a more graceful way to handle this. But that's beyond my scope.


end-4:
actually f you have a touchpad you can do horizontal scroll to see the rest of the title... the same goes for the music widget. though surely it's weird.

and about truncation, this isn't the way. let me tell you about the `truncate` property:
image

end-4:
fixed so i'm closing

PR #326, opened 2024-03-12
hypr: update config files to comply with hyprlang

Cruleo:
Hyprlang is in use since commit https://github.com/hyprwm/Hyprland/commit/13f6f0b923ff3ec94a3bec886c28b90402ceef91 which changes how comments are parsed. Comments of the type `############# Tearing #############`  or `## Tearing` are no longer valid. This wasn't an observed issue until commit https://github.com/hyprwm/Hyprland/commit/66330281ff19e84c986c5ce639b670929e5dadd6 landed (sourced files weren't taken into consideration till then). 

I just removed any double `#` and added a space after the first `#` on any longer strings. An explanation to this can be found under https://github.com/hyprwm/Hyprland/commit/fae47ef462d1d7969afe001070839391ba6dbaa8. 

Any feedback appreciated.

### DIFF
--- a/.config/hypr/custom/env.conf
+++ b/.config/hypr/custom/env.conf
@@ -1,2 +1,2 @@
-## You can put extra environment variables here
-## https://wiki.hyprland.org/Configuring/Environment-variables/
+# You can put extra environment variables here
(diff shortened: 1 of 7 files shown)


clsty:
Thanks!

PR #335, opened 2024-03-15
Add config for brightness control

jsl9208:
Hi, thank you for your awesome project! I've encountered two issues while using it on my ASUS G14 laptop:

 1.  The default brightnessCtl command controls the wrong device, so I need to pass -d amdgpu_bl1 to make it work correctly.
 2.  The G14 uses an OLED display. I inadvertently set the brightness to 0 several times, rendering it totally unusable. I think it would be better to have a brightness configuration to prevent this situation from happening again.

I have added some configuration to solve these problems, brightness.device and brightness.minPercent.

### DIFF
--- a/.config/ags/modules/.configuration/user_options.js
+++ b/.config/ags/modules/.configuration/user_options.js
@@ -22,7 +22,7 @@ let configOptions = {
         'critical': 10,
     },
     'brightness': {
-      'device': '', // `-d` param passed to brightnessctl, leave empty to use brightnessctl default
+      'device': '', // `-d` param passes to brightnessctl, leave it empty to use brightnessctl default
       'minPercent': 0, // Lowest brightness (0-100)
     },
     'music': {
--- a/.config/ags/modules/.configuration/user_options.js
+++ b/.config/ags/modules/.configuration/user_options.js
@@ -21,6 +21,10 @@ let configOptions = {
         'low': 20,
         'critical': 10,
     },
+    'brightness': {
+      'device': '', // `-d` param passed to brightnessctl, leave empty to use brightnessctl default
+      'minPercent': 0, // Lowest brightness (0-100)
+    },
     'music': {
         'preferredPlayer': "plasma-browser-integration",
     },
--- a/.config/ags/services/brightness.js
+++ b/.config/ags/services/brightness.js
@@ -20,10 +20,10 @@ class BrightnessService extends Service {
 
     // the setter has to be in snake_case too
     set screen_value(percent) {
-        percent = clamp(percent, 0, 1);
+        percent = clamp(percent, this._minValue, 1);
         this._screenValue = percent;
 
-        Utils.execAsync(`brightnessctl s ${percent * 100}% -q`)
+        Utils.execAsync(`${this._brightnessctlCmd} s ${percent * 100}% -q`)
             .then(() => {
                 // signals has to be explicity emitted
                 this.emit('screen-changed', percent);
@@ -37,8 +37,12 @@ class BrightnessService extends Service {
 
     constructor() {
         super();
-        const current = Number(exec('brightnessctl g'));
-        const max = Number(exec('brightnessctl m'));
+        const device = userOptions.brightness.device;
+        this._brightnessctlCmd = `brightnessctl ${device ? `-d ${device}` : ''}`
+        this._minValue = clamp((userOptions.brightness.minPercent || 0)/100, 0, 1);
+
+        const current = Number(exec(`${this._brightnessctlCmd} g`));
+        const max = Number(exec(`${this._brightnessctlCmd} m`));
         this._screenValue = current / max;
     }
 


clsty:
While I do appreciate this work, I also personally suggest to consider about ddc control.

Modern monitors support DDC protocol, so when they are connected with computers via DP or HDMI, they can be controlled via DDC.
`brightnessctl` however does not support it currently.

Basically we can use ddcutil to do this, but it requires specifying the device and is not automatic.

A better solution will be `clightd`, which provides a dbus interface to control all backlights, including the normal ones (which `brightnessctl` does) and the DDC ones (normally the connected monitors via DP or HDMI cable).

A script to use `clightd` to control brightness: <https://github.com/clsty/dots-hyprarch/blob/main/.local/bin/clightctl>

Usage example in keybinds.conf:
```conf
bindle = , XF86MonBrightnessUp, exec, ~/.local/bin/clightctl backlight r
bindle = , XF86MonBrightnessDown, exec, ~/.local/bin/clightctl backlight l
```

jsl9208:
Thanks for the advice. I don't use a external monitor, just the internal one. 

I think the problem is, my laptop has two GPU device, integrated one and a Nvidia dGPU. Both of them can handle the display output by switching to different modes (Optimus, Integrated, etc. ). Those tools don't know which device is in-using.

In my case, the default one is nvidia for most of the tools. I have tested brightnessCtl, clightd above, brillo. Only brillo has the feature to detect the in-using device automatically, worked correctly. 

I think it's too heavy and complicated to use the vendor-wise tool to determine which mode the system is currently at, and then choose the correct device to control. A additional config is better for most of the cases. But maybe using brillo is the best approach?

Zeps2369:
> Thank you for the advice. I don't use an external monitor, just the internal one.
> 
> I believe the issue is that my laptop has two GPU devices, an integrated one and an Nvidia dGPU. Both of them can handle the display output by switching to different modes (Optimus, Integrated, etc.). These tools are unable to determine which device is in use.
> 
> In my case, the default device is Nvidia for most of the tools. I have tested brightnessCtl, clightd, and brillo. Only brillo has the feature to detect the in-use device automatically and it worked correctly.
> 
> I think it's too heavy and complicated to use the vendor-specific tool to determine which mode the system is currently in, and then choose the correct device to control. An additional config is better for most cases. But perhaps using brillo is the best approach?

Hi. Im having the same isues using the g14 2024 model. Can you explain how you fixed it using brillo in detail please :)

jsl9208:
> > Thank you for the advice. I don't use an external monitor, just the internal one.
> > I believe the issue is that my laptop has two GPU devices, an integrated one and an Nvidia dGPU. Both of them can handle the display output by switching to different modes (Optimus, Integrated, etc.). These tools are unable to determine which device is in use.
> > In my case, the default device is Nvidia for most of the tools. I have tested brightnessCtl, clightd, and brillo. Only brillo has the feature to detect the in-use device automatically and it worked correctly.
> > I think it's too heavy and complicated to use the vendor-specific tool to determine which mode the system is currently in, and then choose the correct device to control. An additional config is better for most cases. But perhaps using brillo is the best approach?
> 
> Hi. Im having the same isues using the g14 2024 model. Can you explain how you fixed it using brillo in detail please :)

just install brillo, then `brillo -S 20%`

Zeps2369:
> > > Thank you for the advice. I don't use an external monitor, just the internal one.
> > > I believe the issue is that my laptop has two GPU devices, an integrated one and an Nvidia dGPU. Both of them can handle the display output by switching to different modes (Optimus, Integrated, etc.). These tools are unable to determine which device is in use.
> > > In my case, the default device is Nvidia for most of the tools. I have tested brightnessCtl, clightd, and brillo. Only brillo has the feature to detect the in-use device automatically and it worked correctly.
> > > I think it's too heavy and complicated to use the vendor-specific tool to determine which mode the system is currently in, and then choose the correct device to control. An additional config is better for most cases. But perhaps using brillo is the best approach?
> > 
> > 
> > Hi. Im having the same isues using the g14 2024 model. Can you explain how you fixed it using brillo in detail please :)
> 
> just install brillo, then `brillo -S 20%`

what is your solution for keyboard backlight?

jsl9208:
> > > > Thank you for the advice. I don't use an external monitor, just the internal one.
> > > > I believe the issue is that my laptop has two GPU devices, an integrated one and an Nvidia dGPU. Both of them can handle the display output by switching to different modes (Optimus, Integrated, etc.). These tools are unable to determine which device is in use.
> > > > In my case, the default device is Nvidia for most of the tools. I have tested brightnessCtl, clightd, and brillo. Only brillo has the feature to detect the in-use device automatically and it worked correctly.
> > > > I think it's too heavy and complicated to use the vendor-specific tool to determine which mode the system is currently in, and then choose the correct device to control. An additional config is better for most cases. But perhaps using brillo is the best approach?
> > > 
> > > 
> > > Hi. Im having the same isues using the g14 2024 model. Can you explain how you fixed it using brillo in detail please :)
> > 
> > 
> > just install brillo, then `brillo -S 20%`
> 
> what is your solution for keyboard backlight?

As far I am only able to switch it on/off using [asusctl](https://asus-linux.org/manual/asusctl-manual/)
`asusctl led-pow-2 keyboard`

Issue #283, opened 2024-02-24
[Issue] Cheatsheet appears too large on small displays when using display scaling


b4shful:
- **I have read the [wiki](https://end-4.github.io/dots-hyprland-wiki/en/i-i/03troubleshooting)**: yes
- Linux distro: Arch

---
## The issue

When display scaling is set in hyprland.conf on a smaller display (for example the 13 inch display of my XPS 13 9360) the cheatsheet is too big and exceeds the boundaries of the display.

The hyprland config setting used was:
```
monitor=,preferred,auto,1.5
```

It is worth noting this bug also happens if you are not using scaling but instead using a larger font size. (Specified in `gsettings` as outlined in the Usage instructions "How to resize the UI / change font size?")

Apologies, couldn't take a screenshot so hopefully a photo suffices, here is the cheatsheet when triggered with `Super` + `/`:
<img src='https://github.com/end-4/dots-hyprland/assets/28313200/62c7bf57-c38e-4a88-94f1-52140c3598df' width=50%>

<details>
<summary>Click to expand</summary>

Monitor resolution: 1920x1080
Monitor size: 13.3 inches
</details>

end-4:
you'll just have to either reduce the monitor scacle or the font size

b4shful:
@end-4 the scaling is great everywhere else on the system, which is why I am reluctant to do either of those things. Is there a way I could somehow either reduce scale or font size for only the cheatsheet?

b4shful:
oh sorry, what I meant was that I am satisfied with the font size and scaling for all the other elements of the UI and so don't want to have to change it globally like that (it's a laptop so only one monitor)

daUnknownCoder:
> you'll just have to either reduce the monitor scale or the font size

i also have this problem:
![image](https://github.com/end-4/dots-hyprland/assets/84800625/5dab61cb-db44-4795-839c-3e3cca524de3)
tried adding `width` and `height` in cheatsheet's export window to find that they dont have such propertied fiddled with scss too which didnt yield any results

end-4:
> tried adding width and height in cheatsheet's export window to find that they dont have such propertied fiddled with scss too which didnt yield any results

no point in adjusting the size when the children content still take up that much space

b4shful:
Think this one got fixed a while back (perhaps by the new cheatsheet system?) - closing

PR #336, opened 2024-03-16
Add HIDPI variables and auto-scale feature for cheatsheet

jsl9208:
This PR adds some HIDPI variables and uses them to automatically scale down the cheatsheet, making it usable on scaled screens. Takes effect on Hyprland.

Solves https://github.com/end-4/dots-hyprland/issues/283. 

Tested 2880x1800, 2x scaled:
图片

1920x1080, 2x:
图片

### DIFF
--- a/.config/ags/modules/cheatsheet/main.js
+++ b/.config/ags/modules/cheatsheet/main.js
@@ -4,6 +4,7 @@ import Service from 'resource:///com/github/Aylur/ags/service.js';
 import { Keybinds } from "./keybinds.js";
 import { setupCursorHover } from "../.widgetutils/cursorhover.js";
 import PopupWindow from '../.widgethacks/popupwindow.js';
+import { SCREEN_REAL_WIDTH } from '../../variables.js';
 
 const cheatsheetHeader = () => Widget.CenterBox({
     vertical: false,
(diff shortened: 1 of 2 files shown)


end-4:
this is really interesting... but why not have this for all windows? :thinking:

jsl9208:
> this is really interesting... but why not have this for all windows? 🤔

Sorry I didn't understand the question. You mean other widges? Only the cheatsheet and overview window looks way much bigger after scaled IMO.

jsl9208:
Fixing the issue with the overview window is simple. All that needs to be done is to change all occurrences of the variables such as SCREEN_* in its code to SCREEN_REAL_*. I am able to work on it after this merged.

Mooling0602:
> Fixing the issue with the overview window is simple. All that needs to be done is to change all occurrences of the variables such as SCREEN_* in its code to SCREEN_REAL_*. I'd like to work on it later.

Hey bro, you may forgot to import { SCREEN_REAL_HEIGHT } in the `.config/ags/modules/cheatsheet/main.js`, so the change will be not complete. 
But I can't fix this, so I need your help to improve this.

Mooling0602:
> Fixing the issue with the overview window is simple. All that needs to be done is to change all occurrences of the variables such as SCREEN_* in its code to SCREEN_REAL_*. I'd like to work on it later.

Hey bro, you may forgot to import { SCREEN_REAL_HEIGHT } in the `.config/ags/modules/cheatsheet/main.js`, so the change will be not complete. 
But I can't fix this, so I need your help to improve this.

PR #337, opened 2024-03-17
Material You schemes support, transparency in AGS, colorscheme UI and more

midn8hustlr:
Major Changes:
- Added settings in colorscheme widget to switch light/dark mode, and transparency effects in AGS. I feel that light mode looks better in transparent mode and is actually usable now.
- Buttons to switch material schemes provided by Material-foundation (make MaterialYou colorscheme less or more boring). Setting a wallpaper and switching between few material schemes yields pretty satisfactory results. This eventually solves #281.
- Better color generation script with [materialyoucolor-python](https://github.com/T-Dynamos/materialyoucolor-python) which is more aligned with the official [material-color-utilities](https://github.com/material-foundation/material-color-utilities) library and generates more material colors.

https://github.com/end-4/dots-hyprland/assets/47367711/611d4fc6-d794-4828-9664-fb7a10e7128c

Note:
- I have ideas for few more features to implement in this, but rushed for this PR as I see @end-4 is working on revamping the scss files recently and this would create a lot of merge conflicts for me later-on. Plus with this PR I am doing few things differently than end-4 (mentioned below) and this PR would help to resolve any conflict of interests.
- I am not a fan of creating new material colors by mixing with other colors or changing the opacity, as this makes the layers or containers look similar and reduce the vibrancy making material very boring. For some cases like hover/focus colors it is fine though. So I have replaced most of the colors in SCSS files with the standard material colors provided by the Python library. This might not suit everyone, but its fine since we can just change the material scheme according to our taste. I feel instead of manually tinkering with colors in SCSS, this task can be offloaded to theme library itself.
- I have removed most of the transparentize() functions in _colors.scss as this was a pain for supporting the transparency mode in AGS. In transparent mode all the surface colors are transparentized by same amount. 
- I am still using the older [material-color-utilities-python](https://github.com/avanisubbiah/material-color-utilities-python) library to extract colors from image, as with the [materialyoucolor-python](https://github.com/T-Dynamos/materialyoucolor-python), i am not able to extract good colors from image with color scoring algorithm. I will try to raise an issue for this. Personally I am using the c++ library from [material-color-utilities](https://github.com/material-foundation/material-color-utilities) in my system which gives me better colors from wallpaper.
- Now the colorscheme generation  is handled by [materialyoucolor-python](https://github.com/T-Dynamos/materialyoucolor-python).
- materialyoucolor-python is not available in AUR yet, so I have modified the installation script to install it in a virtual environment in /opt/materialyoucolor. Please install it first before changing colorschemes.
- Things like pywal/badapple might break, I have not checked it properly. I will do a PR later if I find any issues.

Suggestions are welcome.


end-4:
Very ~~awesome~~hypr
> I am not a fan of creating new material colors by mixing with other colors or changing the opacity

Well, same. I admit that it sucks.

I very welcome this change and will try to review as soon as possible

end-4:
Works out of the box
So far my favourite is "Vibrant"

end-4:
@clsty what do you think about this way of installing the python library? I'm not a fan of having the shebang point to a specific location in `/opt`, but idk if there's a better way

clsty:
> @clsty what do you think about this way of installing the python library? I'm not a fan of having the shebang point to a specific location in `/opt`, but idk if there's a better way

I'm not very experienced with python lib installaion too, but `/opt` sure is a proper place to install programs without package manager, so I think it is OK.

midn8hustlr:
> If it is just about shebang, then instead of changing shebang, how about using `source /opt/materialyoucolor/bin/activate`?
I did try to do it in this way first, but after executing the python script you also have to run the 'deactivate' command to exit from virtual environment, otherwise any python script downstream using other python libs from system like gradience-cli will not work. You have add three lines of bash everytime using this script (in AGS modules too inside execAsync), this makes the script less portable imo.
Also instead of /opt, we can install the libs in home directory which will not need sudo permissions, but I wasn't able to find a way to point the shebang to ${HOME} or ~.
> 
> If you still want a better way, then submitting a new AUR package I guess, which I don't have any experience on for now. [Reference PKGBUILD of python-material-color-utilities](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=python-material-color-utilities)
Thanks for the reference, i can volunteer to package this.

midn8hustlr:
>If it is just about shebang, then instead of changing shebang, how about using `source /opt/materialyoucolor/bin/activate`?
>
I did try to do it in this way first, but after executing the python script you also have to run the 'deactivate' command to exit from virtual environment, otherwise any python script downstream using other python libs from system like gradience-cli will not work. You have add three lines of bash everytime using this script (in AGS modules too inside execAsync), this makes the script less portable imo.
Also instead of /opt, we can install the libs in home directory which will not need sudo permissions, but I wasn't able to find a way to point the shebang to ${HOME} or ~.
> 
> If you still want a better way, then submitting a new AUR package I guess, which I don't have any experience on for now. [Reference PKGBUILD of python-material-color-utilities](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=python-material-color-utilities)
>
Thanks for the reference, i can volunteer to package this.

end-4:
Alright. If you're doing it, let me know when you're done
btw I'm going to note that the transition of bluetooth device when opening sideright is a bit weird

https://github.com/end-4/dots-hyprland/assets/97237370/b6f46890-9e41-497e-ac4f-682d54f6d414

midn8hustlr:
I don't have any previous experience of packaging in AUR, i will try tonight, not sure if I will suceed.
I will let you know how it goes.

midn8hustlr:
> I'm going to note that the transition of bluetooth device when opening sideright is a bit weird
> 
Yes, I have encountered this too, idk how the bluetooth device widget is done in AGS, as I did the SCSS styling before the widget was introduced. Git somehow merged to make this work. Need to check where the delay is coming from.

end-4:
> idk how the bluetooth device widget is done in AGS

no worries, i'll handle that when we have this merged

midn8hustlr:
Here you go: python-materialyoucolor-git

end-4:
eyy that works

end-4:
There are many colors I wanna change, but I'll just merge this then figure it out later. Thanks!

PR #343, opened 2024-03-21
For gemini memory

antkss:
image


end-4:
- let's not remove init prompts
- let's store the history in `~/.cache/ags/user/ai/chats/gemini_history.txt`

antkss:
you can edit if you want, i just want it to be pure

PR #348, opened 2024-03-23
Fix full-screen recording for multimonitors

midn8hustlr:
If multiple monitors are present, the wf-recorder command will require to select the monitor on stdin. This overcomes that by specifying the monitor on '-o' or '--output' flag.


end-4:
thanks

PR #369, opened 2024-03-28
exported useFlag from statusicons to user_options

SuperSuslik312:


end-4:
hmm
actually i'd prefer a section called "appearance"

PR #392, opened 2024-04-06
Translation keybind using fuzzel

Nevrdid:
Alright, if some people may me interested by, here it is.
Script source : https://gist.github.com/Isti115/114b3a6ba953ec483a954499772ed5f8

### DIFF
--- a/.config/hypr/hyprland/keybinds.conf
+++ b/.config/hypr/hyprland/keybinds.conf
@@ -33,6 +33,7 @@ bind = Control+Shift, Escape, exec, gnome-system-monitor
 
 # Actions
 bind = Super, Period, exec, pkill fuzzel || ~/.local/bin/fuzzel-emoji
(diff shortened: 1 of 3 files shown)


clsty:
We already have a AI sidebar which can do a great translation job, so maybe this one is redundant.
Thank you anyway.

Maybe have this one in Import Manually folder? I'll let @end-4 decide.

end-4:
I don't like it, sorry
Having to type both languages and not being able to paste doesn't make it the best translator...

end-4:
> We already have an AI sidebar which can do a great translation job, so maybe this one is redundant.

but you have to ask a question and it'll probably overkill

I already have the idea of adding a translation api to the sidebar in mind, but haven't got to it yet

end-4:
> Maybe have this one in Import Manually folder? I'm not sure, and I'll let @end-4 decide

I don't like it, won't use this so I won't. Users can add it themselves

PR #396, opened 2024-04-09
Added proxy url setting in user options

SuperSuslik312:
I checked it out in Gemini, works fine.


end-4:
@dengxk1234 this is what you were looking for in a question in #372 right

PR #422, opened 2024-04-17
update dock

casglistro:
I'm used to dock so I updated the `dock` in the `modules`, I don't know why it's not being used.
图片
It generally continues the design from the dock, but it is always placed at the top level and only exists on one monitor at most. When focusing on another monitor, the dock will move to that monitor. It also hides itself when `onHoverLost` is triggered, and there is a button to hold the dock in place.
For the missing icons, I do a recursive search in the system's icon path, preprocessing the path of each image under the path and storing it in an array at ags startup. When an icon is needed use `Levenshtein Distance` to find the closest string according to the class of the window. In my tests it seems that a lot of all icons can be corresponded to. However, there are requirements for the search path: if there are too many images there is an expensive performance hit, if there are too few images there may be missing icons, so I chose `tela-icon-theme`.
If you have suggestions or complaints please let me know.

### DIFF
--- a/.config/ags/modules/.configuration/user_options.js
+++ b/.config/ags/modules/.configuration/user_options.js
@@ -82,14 +82,9 @@ let configOptions = {
         'pinnedApps': ['firefox', 'org.gnome.Nautilus'],
         'layer': 'top',
         'monitorExclusivity': true, // Dock will move to other monitor along with focus if enabled
-        // It's useful to keep the icons consistent, which is useful if you're OCD :)
-        'searchPinnedAppIcons': false,
-        // available: client_added, client_move, workspace_active, client_active
-        'trigger': ['client-added', 'client-removed'],
-        // Automatically hide dock after a period of time
-        // after a trigger has been triggered.
-        // Time in milliseconds. empty if always displays.
-        // { 'trigger': 'client-added', interval: 1000, }
+        'searchPinnedAppIcons': false, // Try to search for the correct icon if the app class isn't an icon name
+        'trigger': ['client-added', 'client-removed'], // client_added, client_move, workspace_active, client_active
+        // Automatically hide dock after `interval` ms since trigger
         'autoHide': [
             {
                 'trigger': 'client-added',
@@ -104,10 +99,9 @@ let configOptions = {
     // Longer stuff
     'icons': {
         // Find the window's icon by its class with levenshteinDistance
-        // All file paths are preprocessed and stored at ags startup, so if there
-        // are so many files under the path it will affect performance
-        // Maybe you need a comprehensive icon theme
-        // Example: ['/usr/share/icons/Tela-nord/scalable/apps', 'others...']
+        // The file names are processed at startup, so if there
+        // are too many files in the search path it'll affect performance
+        // Example: ['/usr/share/icons/Tela-nord/scalable/apps']
         'searchPaths': [''],
 
         substitutions: {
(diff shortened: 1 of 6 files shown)


clsty:
> I'm used to dock so I updated the dock in the modules, I don't know why it's not being used.

I don't know either, but there are some people who don't like dock, such as me.
I suggest you to make an option to enabled/disabled the dock in `user_options.js`.

end-4:
I disabled dock since I haven't managed to get it work correctly and don't really care (I mean, we use a *tiling* window manager). But having it optional is the way now that there's a config file.

.config/ags/modules/dock/dock.js
```diff
@@ -2,19 +2,26 @@ const { Gtk } = imports.gi;
...
 
 import Hyprland from 'resource:///com/github/Aylur/ags/service/hyprland.js';
 import Applications from 'resource:///com/github/Aylur/ags/service/applications.js';
 const { execAsync, exec } = Utils;
 const { Box, Revealer } = Widget;
 import { setupCursorHover } from '../.widgetutils/cursorhover.js';
+import { getAllFiles, searchIcons } from './icons.js'
+import { MaterialIcon } from '../.commonwidgets/materialicon.js';
+
+const icon_files = getAllFiles("/usr/share/icons/Tela-nord-dark/scalable/apps")
```
end-4: don't hard code it

.config/ags/modules/dock/dock.js
```diff
@@ -177,82 +236,87 @@ const PinnedApps = () => Widget.Box({
...
-            dockRevealer.revealChild = false;
-        },
+        // onHoverLost: () => {
+        //     if (Hyprland.active.client.attribute.class.length === 0) { return }
+        //     dockRevealer.revealChild = false;
+        // },
         child: Box({
             homogeneous: true,
-            css: 'min-height: 2px;',
+            css: 'min-height: 20px;',
```
end-4: 20px would be too much and might block apps. can you add a config option and have it default to 5? (since that's the hyprland outer gap value)

end-4 (changes requested):

.config/ags/modules/.configuration/user_options.js
```diff
@@ -71,11 +71,46 @@ let configOptions = {
         'dateFormat': "%d/%m", // On notif time
     },
     'weather': {
-        'city': "",
+        'city': "Chengdu",
```
end-4: ughh don't

.config/ags/modules/.configuration/user_options.js
```diff
@@ -71,11 +71,46 @@ let configOptions = {
...
+        'hoverMinHeight': 5,
+        'pinnedApps': ['firefox', 'org.gnome.Nautilus'],
+        // top or bottom
+        'layer': 'top',
+        // Find the window's icon by its class with levenshteinDistance
+        // All file paths are preprocessed and stored at ags startup, so if there
+        // are so many files under the path it will affect performance
+        // Maybe you need a comprehensive icon theme
+        // Example: ['/usr/share/icons/Tela-nord-dark/scalable/apps', 'others...']
+        'iconSearchPaths': [''],
```
end-4: let's move this to an 'icon' category since i'll have this icon guessing for overview too after this

.config/ags/modules/.configuration/user_options.js
mon3stera: ohh i forget it :(

end-4:
i removed the active workspace trigger since the dock showing up every time switching workspace is kinda weird

end-4:
alrighty, thanks for your work!

Issue #424, opened 2024-04-18
[Issue] Overview workspace view doesn't scale properly with set DPI and force_zero_scaling


tcharchar:
- **(For usage) I have read the [wiki](https://end-4.github.io/dots-hyprland-wiki/en/i-i/02usage)**
- **(For troubleshooting) I have read the [wiki](https://end-4.github.io/dots-hyprland-wiki/en/i-i/03troubleshooting)**
- Linux distro: Arch Linux x86_64, Kernel 6.8.6-zen1-1-zen

---
## The issue

EDIT: Forgot to mention, but I'm using `wlr-randr --scale 1.25`
As title mentions, for some reason having `xwayland{ force_zero_scaling = true }` messes up with the overview widget, specifically the workspace preview. See image below:
![image](https://github.com/end-4/dots-hyprland/assets/110164189/dcb87669-17a1-49fc-a7f3-8b1b300937f0)

end-4:
maybe adjust overview scale in user config?

tcharchar:
> maybe adjust overview scale in user config?
Looks better, but the issue of apps not properly filling up the preview is still present.

![image](https://github.com/end-4/dots-hyprland/assets/110164189/165ae55a-982a-4947-81a5-1022b234956b)

alok-debnath:
You can reduce the number of workspaces shown from 10 to 8, that's how I fixed mine. I current don't have access to my laptop, will update this reply shortly.

tcharchar:
Changing the number of columms from 10 to 8 makes it now fit without changing the scale from its base 0.18 value, but it still has the issue of not properly filling up a workspace preview.
![image](https://github.com/end-4/dots-hyprland/assets/110164189/fadf8850-656c-4681-a6e3-7502d6d3763a)

DiXN:
What I have done in the past, is to divide the width / height by the current scale in `variables.js` :

```
export const SCREEN_WIDTH = Number(exec(`bash -c "hyprctl monitors -j | gojq '.[].width / .[].scale' | awk '{print int($1+0.5)}'"`));
export const SCREEN_HEIGHT = Number(exec(`bash -c "hyprctl monitors -j | gojq '.[].height / .[].scale' | awk '{print int($1+0.5)}'"`));
```
However, this approach has some flaws. This approach only works with Hyprland and does not work with multiple monitors. In order to solve this problem, I think there has to be a screen width and a screen height per monitor and the actual screen size after scaling has to be determined.

[...]

Issue #417, opened 2024-04-16
[Issue] Sessionscreen interface display is misaligned when enable HiDPI


Mooling0602:
- **(For usage) I have read the [wiki](https://end-4.github.io/dots-hyprland-wiki/en/i-i/02usage)**
- **(For troubleshooting) I have read the [wiki](https://end-4.github.io/dots-hyprland-wiki/en/i-i/03troubleshooting)**
- Linux distro: Arch Linux

---
## Detail
When enable HiDPI, in config ~/.config/hypr/custom/env.conf
```
# You can put extra environment variables here
# https://wiki.hyprland.org/Configuring/Environment-variables/
monitor = eDP-1,highres,auto,1.6
...
```
The sessionscreen will be displayed like this: 
![image](https://github.com/end-4/dots-hyprland/assets/147964268/b63277b4-8ff7-4387-9539-57828945bd6f)
And if nothing changed, the sessionscreen will be normal: 
![image](https://github.com/end-4/dots-hyprland/assets/147964268/079a19a4-89f1-4a99-a4d6-42e41c66346e)
I don't know how to fix it, and I need help so I issue here.

alok-debnath:
yes that happens with me too when I use higher scaling
![image](https://github.com/end-4/dots-hyprland/assets/67475527/d2e65f6a-2d45-4c84-87a6-21238e8e4cae)

alok-debnath:
Btw, this can be fixed if we use this pull request #336 (mainly this file `.config/ags/variables.js`).
We just need to replace
```js
min-width: ${SCREEN_WIDTH}px;
min-height: ${SCREEN_HEIGHT}px;
```
with
```js
min-width: ${SCREEN_REAL_WIDTH}px;
min-height: ${SCREEN_REAL_HEIGHT}px;
```
after importing those 2 variables from the mentioned file.
@end-4 Is there any plans for merging that PR, not sure how it will behave on a multi monitor setup but it is working fine on my laptop

Mooling0602:
Sorry I forgot I should import this PR first.

Ate329:
Substitute SCREEN_WIDTH and SCREEN_HEIGHT in /.config/ags/variables.js with these two lines can solve the issue and that's how my problem was solved.
```
export const SCREEN_WIDTH = Number(exec(`bash -c "hyprctl monitors -j | gojq '.[].width / .[].scale' | awk '{print int($1+0.5)}'"`));
[...]

PR #436, opened 2024-04-24
Use `GDK` to determine the monitor geometry.

DiXN:
With GDK the screen height and width can be determined. This has the added benefit that if scaling is applied, it reports the final screen width and height after scaling. This solves various scaling issues, as reported in https://github.com/end-4/dots-hyprland/issues/424 and https://github.com/end-4/dots-hyprland/issues/417. I also had some ideas on how to consider different screen sizes for multiple monitors in https://github.com/end-4/dots-hyprland/issues/424#issuecomment-2067395463.

### DIFF
--- a/.config/ags/variables.js
+++ b/.config/ags/variables.js
@@ -16,8 +16,21 @@ globalThis['openColorScheme'] = showColorScheme;
 globalThis['mpris'] = Mpris;
 
 // Screen size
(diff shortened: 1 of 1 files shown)


end-4:
I've made ags get monitor size from hyprctl 18de9e2fea19f919dce04c1a6dafa7b49903716a, which is better for #424, I'll explain why in that thread
Thanks for your PR anyway!

PR #455, opened 2024-05-01
fix multi-monitor widgets

MoetaYuko:
these widgets were declared for each monitor but never triggered due to:
```
╰─ xrandr --listmonitors
Monitors: 2
 0: +DP-1 3840/600x2160/340+0+0  DP-1
 1: +DP-2 3840/530x2160/300+7680+0  DP-2

╰─ xrandr --listmonitors | grep -c 'Monitor'
1
```

this pr fixes keybinds and widget implementations for the second monitor

### DIFF
--- a/.config/ags/modules/cheatsheet/main.js
+++ b/.config/ags/modules/cheatsheet/main.js
@@ -70,7 +70,7 @@ const CheatsheetHeader = () => Widget.CenterBox({
     }),
 });
 
-const sheetContents = {};
+const sheetContents = [];
 const SheetContent = (id) => {
     sheetContents[id] = ExpandingIconTabContainer({
         tabsHpack: 'center',
--- a/.config/ags/modules/cheatsheet/main.js
+++ b/.config/ags/modules/cheatsheet/main.js
@@ -10,12 +10,12 @@ const cheatsheets = [
     {
         name: 'Keybinds',
         materialIcon: 'keyboard',
-        contentWidget: Keybinds(),
+        contentWidget: Keybinds,
     },
     {
         name: 'Periodic table',
         materialIcon: 'experiment',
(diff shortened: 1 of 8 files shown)


.config/ags/scripts/color_generation/switchwall.sh
```diff
@@ -5,11 +5,13 @@ if [ "$1" == "--noswitch" ]; then
     # imgpath=$(ags run-js 'wallpaper.get(0)')
 else
     # Select and set image (hyprland)
-    cd "$HOME/Pictures"
+    cd "$(xdg-user-dir PICTURES)"
```
clsty, line 8: Need extra dep here. Added.

.config/ags/modules/cheatsheet/main.js
```diff
@@ -70,20 +70,24 @@ const CheatsheetHeader = () => Widget.CenterBox({
...
-    transitionDuration: userOptions.animations.durationLarge * 1.4,
-    icons: cheatsheets.map((api) => api.materialIcon),
-    names: cheatsheets.map((api) => api.name),
-    children: cheatsheets.map((api) => api.contentWidget),
-    onChange: (self, id) => {
-        self.shown = cheatsheets[id].name;
-        if (cheatsheets[id].onFocus) cheatsheets[id].onFocus();
-    }
-});
+const sheetContents = {};
```
end-4: why use an object here :thinking:

end-4:
i totally forget about stuff that aren't visible on the initial view of the notifs page...

.config/ags/modules/cheatsheet/main.js
moetayuko: yup array is better here

PR #458, opened 2024-05-01
ags: fix systray icon warning

moetayuko:
take code snippet from ags wiki to fix:
```
(com.github.Aylur.ags:748022): Gjs-Console-WARNING **: 21:42:04.733: Error: can't assign "[object instance wrapper GIName:GdkPixbuf.Pixbuf jsobj@0x37f2745abd08 native@0x62e804712490]" as icon, it is not a file nor a named icon
```


end-4:
i don't see why, but it makes the code shorter and doesn't break, so alright

MoetaYuko:
I don't see why, either. The warning just magically disappeared... also no functional difference was observed

PR #459, opened 2024-05-02
replace wlsunset with gammastep

MoetaYuko:
wlsunset doesn't work w/o manually passing the latitude and longitude.
gammastep can retrieve geolocation from geoclue so it works out of the
box for everyone.


MoetaYuko:
one issue that remains is that the user preference doesn't persist. may need to be stored in a file or whatever

end-4:
nice, i'm finding myself using light theme recently :)

PR #466, opened 2024-05-02
darkman integration

### LINKED ISSUE #447: [Feature] darkman integration
**What would you like to be added?**
Use darkman to automatically toggle light/dark color schemes based on geolocation (provided by geoclue).

**How will it help**

1. eye candy
2. browsers can retrieve day/night from dbus provided by darkman and adjust webpage contents accordingly (e.g. activates dark theme provided by github and google).

MoetaYuko:
fixes #447, much easier than I expected thanks to existing infra

needs #459 which introduced geoclue as a hard dependency of gammastep

users can edit `.config/darkman/config.yaml` to disable auto transition

### DIFF
--- a/.config/ags/modules/.miscutils/system.js
+++ b/.config/ags/modules/.miscutils/system.js
@@ -14,8 +14,10 @@ darkMode.connect('changed', ({ value }) => {
     let lightdark = value ? "dark" : "light";
     execAsync([`bash`, `-c`, `mkdir -p ${GLib.get_user_state_dir()}/ags/user && sed -i "1s/.*/${lightdark}/"  ${GLib.get_user_state_dir()}/ags/user/colormode.txt`])
         .then(execAsync(['bash', '-c', `${App.configDir}/scripts/color_generation/switchcolor.sh`]))
+        .then(execAsync(['darkman', 'set', lightdark]))
         .catch(print);
 });
+globalThis['darkMode'] = darkMode;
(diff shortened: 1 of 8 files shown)


MoetaYuko:
@end-4 any comments on this feature? feel free to close if not worth adding.

end-4:
ugh... i'm hesitant to add this since there's already blue light filter and one can always change the brightness if needed
i think i can just give the dark mode toggle an "Auto" mode and make a new user option in case someone wants to enable/disable it by default without digging in hyprland config

MoetaYuko:
darkman has two major advantages:
1. retrieve geolocation-based transition time from geoclue (like gammastep does)
2. run custom user scripts. e.g., I have additional scripts to switch gtk/qt/kvantum themes between light/dark variants

anyway, whoever needs this can cherry-pick in their fork.

clsty:
I actually kinda like this PR... @end-4 maybe reconsider this one? That custom script feature looks great.

end-4:
I consider darkman to be kind of a scripty extra thing I don't need. 

I will not add it to my hyprland execs, but I can accept the AGS part to support it. So I'll make `darkMode` global and have that `darkman set` command run if it's available on the system. Then you can configure darkman if you wish, and I don't have to add an extra daemon I don't care about.

MoetaYuko:
> So I'll make `darkMode` global and have that `darkman set` command run if it's available on the system. Then you can configure darkman if you wish, and I don't have to add an extra daemon I don't care about.
> 
> 

makes sense, thanks

PR #470, opened 2024-05-03
[Bugfix] Update gpt.js

H0mire:
Since a thinking animation is implemented, the "thinking..." string is not necessary anymore. It causes problems displaying answers from chatgpt. My suggestion: Emptying string.


end-4:
i see

PR #473, opened 2024-05-03
[Feature] Added Update-dots.sh script

H0mire:
Automatically updates the Repository and the dots:
Replaces by default .config and .local files in the home directory.
Checks if user has modified files => Asking before replacing modified files. 
You are able to exclude folders from updating.  e.g.  the "custom" folder
Implemented  with Checksums.

### DIFF
--- a/update-dots.sh
+++ b/update-dots.sh
@@ -26,17 +26,21 @@ function file_in_excludes() {
   return 1
 }
 
-
-
-
 # Then check which files have been modified by the user since the last update to preserve user configurations
 modified_files=()
 
 # Find all files in the specified folders and their subfolders
 while IFS= read -r -d '' file; do
+    # If the file is not in the home directory, skip it
+    if [[ ! -f "$HOME/$file" ]]; then
+        continue
+    fi
+
     # Calculate checksums
     base_checksum=$(get_checksum "$base/$file")
     home_checksum=$(get_checksum "$HOME/$file")
+
+
     # Compare checksums and add to modified_files if necessary
     if [[ $base_checksum != $home_checksum ]]; then
         modified_files+=("$file")
@@ -45,6 +49,7 @@ done < <(find "${folders[@]}" -type f -print0)
 
 
 echo "Modified files: ${modified_files[@]}"
+echo "Excluded files and folders: ${excludes[@]}"
 
 # Output all modified files
 if [[ ${#modified_files[@]} -gt 0 ]]; then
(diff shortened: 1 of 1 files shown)


clsty:
Generally this script choose the files under `$base` which does not have a exact same file with md5 hash under `$home`, and prompt whether to sync them, right?

I won't talk about the problem that you did not use `set -e` and either `mkdir -p` or equivalent, but the key point is that:
1. All this can be done by using `rsync` with `--checksum`. You don't have to compare md5sum manually at all.
2. This script only copies the modified files, or not. (Note that you did not ever consider the situation if the destination file does not exists.) It's even meaningless to continue the script after a "n" is given to "Do you want to keep these files untouched? [Y/n]" because it won't do anything anyway, just `exit`.

Though I guess it's possible to implement similar function for `./install.sh`, such as using `--no-overwrite`. I'll work on it when I'm available.

Thanks for your contribution anyway.

Closing.

H0mire:
Thank you for your feedback.

- Correct, i considered rsync, but as you know it would remove the user modified scripts, since the new ones have a different checksum and metadata.
- I also considered the case that git pull has already been called. In that case, the scripts tells the user, that no file has been found that has been configured by the user, and asks if every file should be deleted.
In order to make it independent there must be some cache that holds the checksums created with the install.sh script or the update-dots script. Should I implement that?
- Ok, I will add the possibility to decide for each file independently. But i still will leave a -all option.
- The rest will be fixed.

H0mire:
@clsty I updated the script. Could you please review the script again?

Thanks!

clsty:
The `git clone` part has some major problems.

- I suggest to use a `--depth=1` for `git clone`.
- `find "$temp_folder/$folder" -type f -print0 | while IFS= read -r -d '' file; do` will give you a `$file` var which looks like `./cache/tmp.rhqpeLo6Zw/.config/ags/...`, you surely don't want that `./cache/tmp.rhqpeLo6Zw` thing, else `$HOME/$file` will be messed up.
- It lacks logic to skip modified files.
- After this part is done and temp directory removed, the script does not stop, but continues to copy files from main directory again.

Another problem is that you need to prompt user what this script is for, and give user chance to exit in case they run this by accident.

---

As for the rest part, after reading through it, it seems OK to me, but I don't have a suitable environment to test it, so I can't tell if there're more problems.

Nor have you tested it, right? Else I would not have found out the major problems above.

I suggest you to use it by yourself for a couple of days so you will be able to fix/improve some neglected things, if there is any.

---

@end-4 What do you think about this script?

H0mire:
Good job, great awareness. Makes sense. I tested it, but seems like the wrong way, lol. Thanks, working on it...

H0mire:
Ordered, delivered. Can you check? @end-4  and @clsty

end-4:
> @end-4 What do you think about this script?

@clsty I don't see any problem
Just gonna comment that the grammar/wording is good xd
Since this is a scripting-related PR I'll let you decide when to merge

H0mire:
@clsty It's been a while. Anything that must be improved? Encountered no error since.

clsty:
I can't find out any problem for now, so I'll merge it. Thank you for contribution.

PR #484, opened 2024-05-07
Update rules.conf to fix wps office

H0mire:
Context menus (Right-Click) of wps office applications are treated like windows and therefore tiled with this rule. This leads to an unwanted snapping of the menu next to the active window.
image


end-4:
so that's why...

PR #492, opened 2024-05-11
Update README.md

cosmicwanderer7:
Added Cheat sheet for default key binds

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,15 @@
 </details>
 
 <details> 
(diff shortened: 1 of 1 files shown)


end-4:
i'd rather not show a pic with purple toilet tiles

end-4:
901af41301e0cceb8f30a26ac5c7ea87c39eb898

end-4:
thanks for the idea anyway

PR #515, opened 2024-05-19
Renamed Wolfram to Tungsten in the periodic table

cheyao:
Renamed Wolfram to Tungsten in the periodic table, Tungsten is the more commonly used name. (Even Wikipedia uses Tungsten https://en.wikipedia.org/wiki/Tungsten)

I was trying to find Tungsten on the table but couldn't find it, ended up wasting me 10 minutes. I am submitting this pr to prevent anyone in the future to suffer the same fate as me.


end-4:
fine
basically d76ff6fd875844b6eeb2124ab17566a74d712ed3 reversed lol

PR #527, opened 2024-05-21
fix: truncate if offscreen

Nez21:
Fix a bug in the overview module where the width of window element was not properly truncated.


end-4:
thanks

PR #530, opened 2024-05-23
[bugfix] update dots script

H0mire:
this pull request contains a fix for the bug, which caused that removed or leftover files that were renamed do not get deleted and therefore confused the system.

Further minor improvements:
- Changed Text so it is more clear what the script does
- Changed up coloring a bit
- Added --branch option in git clone.


clsty:
I've noticed this PR and I'll still reserve above a week to let you fully test it before merging.

H0mire:
Ciao @clsty, no errors have occurred for a week, and no bug fixes have been necessary so far. Therefore, it should be safe to merge. Can you verify?

clsty:
Thanks for your contribution.

PR #532, opened 2024-05-24
Adding randomwall and ollama

Soliprem:
- Changes to the switchwall script so that it can also take a positional $1 argument in the form of a wallpaper path
- Added randomwall script
- Added Ollama (with default model llama3:instruct. This could be changed in favour of a smaller model. Requires ollama to be installed and running on the system)
- Tested on my system with no issues

PR #546, opened 2024-05-27
reapply "switchwall: fix transition position for hidpi and multi-monitor"

MoetaYuko:
this was accidentally dropped in #532


end-4:
ah whoops

PR #547, opened 2024-05-27
ags: use system-wide icon theme for systray

MoetaYuko:
The system-wide gtk icon theme can be set by:

    gsettings set org.gnome.desktop.interface icon-theme icon_theme_name

Don't mess up user's preference.

### DIFF
--- a/.config/ags/scss/_bar.scss
+++ b/.config/ags/scss/_bar.scss
@@ -346,7 +346,6 @@ $bar_ws_width_focus_active: 2.045rem;
 .bar-systray-item {
     @include full-rounding;
     @include element_decel;
-    @include symbolic-icon;
     min-height: 1.032rem;
     min-width: 1.032rem;
     font-size: 1.032rem;


end-4:
I don't have Adwaita as my icon theme, but I prefer its symbolic icons. In my defense:
- Adwaita is available almost everywhere, and it's like how i use Material Symbols over system icons for most elements
- I **only** force Adwaita on **symbolic icons**. As Adwaita's symbolic icons are thick, they're **accessible** and **consistent** with Material Symbols.

Though maybe we could make a config option for it if you really wanna change it. Styles are compiled at runtime so it'll probably be easy

MoetaYuko:
I agree that Adwaita is the most accessible icon theme, but it's not complete enough.

For instance, here's my systray w/ Adwaita:
before
and as you can see, keepassxc and kdeconnect icons are indistinguishable from the white background. This is because Adwaita does not provide customized icons for them, and their built-in icons are selected as fallbacks, which are unsuitable for the white background.

On the other hand, they're perfectly fine when using the Papirus-Light icon theme, which is designed for white background:
after

In my personal setting, I use darkman scripts to set the icon theme to Papirus-Light for the light color scheme and Papirus-Dark for the dark color scheme **at runtime** (via the gsettings command mentioned above), which is not possible when it's hardcoded.

end-4:
> (via the gsettings command mentioned above, gtk handles icon reload automatically so ags doesn't need to restart)

yeah but i want to have different icon packs for symbolic and normal icons since the former in my icon pack sucks
i'll implement a config option for separate icon theme for symbolic icons when i have time

MoetaYuko:
> yeah but i want to have different icon packs for symbolic and normal icons since the former in my icon pack suck

Is removing `@include symbolic-icon` from `.bar-systray-item` only acceptable to you? It will resolve my issue anyway

MoetaYuko:
> added that config option

thanks, is it possible to change it from cmdline w/o restarting ags?

end-4:
> thanks, is it possible to change it from cmdline w/o restarting ags?

nope

PR #556, opened 2024-05-29
fixup! ags: config option for symbolic icon theme (#547)

MoetaYuko:


end-4:
what
doesn't it work without the underscore in front? it does for me

MoetaYuko:
> what doesn't it work without the underscore in front? it does for me

not for me:
```
(com.github.Aylur.ags:1693965): Gjs-Console-CRITICAL **: 14:48:10.366: CSS ERROR: Invalid name of pseudo-class at line 8 in /home/xxx/.cache/ags/user/generated/style.css
Gtk.CssProviderError: style.css:9:5Invalid name of pseudo-class
```
can you double check?

MoetaYuko:
meh it worked again

PR #557, opened 2024-05-29
Audio popups also showing up if the volume changes e.g. by keys

myryk31415:
The audio popup now works as in most setups and shows up if the volume changes. It does not if you skip a song etc.

I implemented it for backlight as well, but its not working. The hook doesn't trigger if I change the brightness via keybinds. I assume the service doesnt check for that? Didn't check yet.

### DIFF
--- a/.config/ags/modules/indicators/indicatorvalues.js
+++ b/.config/ags/modules/indicators/indicatorvalues.js
@@ -59,6 +59,7 @@ export default (monitor = 0) => {
         }, 'notify::screen-value'),
         progressSetup: (self) => self.hook(Brightness[monitor], (progress) => {
             const updateValue = Brightness[monitor].screen_value;
+            if (updateValue != progress.value) Indicator.popup(1);
             progress.value = updateValue;
         }, 'notify::screen-value'),
     });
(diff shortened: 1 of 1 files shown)


myryk31415:
One Issue: if you change to headphones, the popup appears because in most cases the volume is different

myryk31415:
The popup now doesn't show up anymore if the device changes. Idk if there is a better way to check tho.

myryk31415:
Should be ready if the feature is desired :)

end-4:
lgtm <- (i often don't say this but i wanna say something different today)

end-4:
anyway thanks, i can remove two keybinds now

myryk31415:
Something didn't work out, for me the Panel now pops up when i pause music as well, I'm gonna have another look at it

myryk31415:
Okay it only happens if you have volume levels over 100%.

(com.github.Aylur.ags:29989): GLib-GObject-CRITICAL **: 20:25:58.111: value "1.113281" of type 'gdouble' is invalid or out of range for property 'fraction' of type 'gdouble'

Any Ideas how to work around that?

myryk31415:
Ah ofc its the progress bar. I'm working on it

PR #563, opened 2024-05-30
ags: use separate symbolic icon theme for different color schemes

MoetaYuko:
One may want to use different icon theme for dark and light backgrounds to ensure readability.

This also fixes a race condition by using `Utils.writeFileSync` to ensure the style file is updated before compiling sass.


MoetaYuko:
Notice that the error I posted at https://github.com/end-4/dots-hyprland/pull/556#issuecomment-2136745848 was caused by the race condition: `_lib_mixins_overrides.scss` had not been created yet when compiling sass.

Issue #549, opened 2024-05-27
[Feature] Respect User Configured XDG Environment Variables


Makrennel:
**What would you like to be added?**
Support for the XDG Base Directories and XDG User Directories environment variables

**How will it help**
It will help for users who have their environment variables set to use different paths besides `$HOME/.config`, `$HOME/.local/share`, and so on, as well as allowing users to save screenshots and screenrecording to user configured directories, rather than forcing use of `$HOME/Videos` and `$HOME/Pictures`

**Extra info**
https://wiki.archlinux.org/title/XDG_Base_Directory
https://wiki.archlinux.org/title/XDG_user_directories

I am willing to work on this myself and PR it if you aren't interested on doing this.

end-4:
am i missing something

Makrennel:
`install.sh` [autocreates](https://github.com/end-4/dots-hyprland/blob/e5339a146b0495cc3ad5610eb9b0d579ac230d04/install.sh#L204) and uses hardcoded paths. I'm not sure what can be done for `~/.local/bin` as the xdg base dirs specification doesn't officially provide a variable for that, although I have seen people use and discuss `XDG_BIN_HOME`.

end-4:
@clsty  👉👈

clsty:
> `install.sh` [autocreates](https://github.com/end-4/dots-hyprland/blob/e5339a146b0495cc3ad5610eb9b0d579ac230d04/install.sh#L204) and uses hardcoded paths. I'm not sure what can be done for `~/.local/bin` as the xdg base dirs specification doesn't officially provide a variable for that, although I have seen people use and discuss `XDG_BIN_HOME`.

Currently the `~/.local/bin` is only for `fuzzel-emoji` and `rubyshot` which are actually "import manually" thingy.

So, just remove the `v rsync -av ".local/bin/" "$HOME/.local/bin/"` line in `install.sh` will do; also the lines in `uninstall.sh`.

For other things, you may give a default value when it's not presented, e.g.
```bash
[[ -z "${XDG_CONFIG_HOME}" ]] && export XDG_CONFIG_HOME=${HOME}/.config

# ...
# Replace all the remainder ~/.config with ${XDG_CONFIG_HOME}
```

@end-4 Sorry I'm not near my home PC recently. Could you please handle this for me? Or leave it to me but wait for some days.

Makrennel:
You can shorten that to

[...]

PR #565, opened 2024-05-30
Make install script obey xdg dirs (#549)

Makrennel:
Updated `install.sh` and `uninstall.sh` to use XDG directories as per #549 
I haven't gotten `update-dots.sh` figured out yet.


Makrennel:
Figured out `update.sh`.... I think

Issue #571, opened 2024-06-02
[Feature] Documentation request


NathanC:
**What would you like to be added?**

An overview of software installed/reasons they were chosen. Basically a high level table breakdown of the core decisions made here and packages used.

**How will it help**

I have no context for a lot of this. I've used Arch extensively, but never tiling windows managers and never waylands. I just went through the hyprland setup and got the baseline working and set up waybar, but for this install setup I can't even tell what status bar is being used because I don't even know the names of any of them.

Basically, if you want this to be useful to newcomers to hyprland, bridging the context gap by an overview of high level decisions made and core software (what status bar, what notification daemon, QT vs GTK etc) would be helpful.

I'm just looking for some more documentation to counter the [curse of knowledge](https://en.wikipedia.org/wiki/Curse_of_knowledge). You have a wiki so seems like it's something in line with what you're trying to do here, and this would have been really helpful to me. Thanks for maintaining this!

end-4:
I honestly totally forget what software is installed, except the ones I mainly work with like 
- Hyprland (the window manager (more technically-accurate term is compositor))
- AGS (a GTK widget system, responsible for the bar)

If you're going to ask me to tell you what a random package on your system is for, I'll probably shrug

end-4:
ughhhhh
So do you want me to make an overview of stuff that were installed by the dotfiles install script or do you want a general utility that explains everything on your system?

NathanC:
lol no, not looking for something to explain everything on my system. 

> AGS (a GTK widget system, responsible for the bar) 

Would have been very useful in the wiki or top level README, along with fuzzle etc, just a brief section on core components/first layer of the dependency tree. By the time you responded, I figured that's what was creating the bar, but it wasn't at all clear at first. Looking for reasonable *entry points* into Googling, that's all.

With the documentation you have and the stated focus on having an accessible system, figured I'd be worth saying what I wished was there as a top-level overview. If you're cool with it, I can add a brief section in the README as I'm digging through this, even just a little bullet point list with stuff like "AGS (a GTK widget system, responsible for the bar) " would be helpful to people.

PR #574, opened 2024-06-02
Update README.md with software choices, intro, and formatting  tweaks

NathanC:
Some changes I would have found helpful, re: https://github.com/end-4/dots-hyprland/issues/571.

lmk if you don't like the formatting changes, I can revert, they're neither here nor there.  Liked your centered header, and wanted to continue that with a sort of "dots" theme

Also copied the wiki intro since it's clean/succinct, and updated a dead branch that's been merged into main.

I like the install script, it's pretty clean having a deps config like this.


end-4:
i like the centered headers, thanks

PR #577, opened 2024-06-03
Substitutions again

### LINKED ISSUE #572: [Issue] Substitutions in user_config broken
- **I have read the [Usage](https://end-4.github.io/dots-hyprland-wiki/en/i-i/02usage) and the [Troubleshooting](https://end-4.github.io/dots-hyprland-wiki/en/i-i/03troubleshooting) pages of the wiki**
- **I have made sure that both my config and system packages are up to date**
- Linux distro: Opensuse Tumbleweed

---
## The issue

Just tried to substitute an app icon and it gave me a wrong config notification. There was a check added a week, ig that broke substitutions. It looks for the specific app name as option in the user_config template, doesn't find it and complains.

myryk31415:
New pull request since I messed up the last one. My bad.

Fixes #572 . The check whether the config option exists doesn't happen for substitutions and regexSubstitutions anymore since we want to be able to add new options there.


end-4:
thanks

PR #580, opened 2024-06-04
Only show 1 digit after decimal point for battry percentage

cheyao:
Keep only one digit after the dots on the battery icon.

Before:
swappy-20240604-180026
After:
swappy-20240604-180038


end-4:
thanks

PR #584, opened 2024-06-05
Improvements for vertical screens

myryk31415:
- Width and height now are correct if the screen is vertical.
- `short` mode for the bar. Only used if the screen has a width smaller than a certain value and currentShellMode is `normal`.

That fixes the logout menu not being in the center of the screen and the bar being cursed on vertical screens.

Issues:
- The treshold for `short` mode was chosen arbitrarily (currently 1800, cause that is smaller than 1920). It works for vertical 1920x1080 screens but might not for other sizes.
- The sidepanels are stretched a lot, but look otherwise normal so ig thats fine.
- Overview is still wider than the screen, one way to fix this would be to switch the amount of columns and rows for vertical screens, but that probably doesn't always work.

For horizontal screens nothing should have changed.
Any feedback and ideas are appreciated :)


myryk31415:
I set the threshold to 1400px so people who have some resolution lower than 1920 don't have the short bar unless necessary.

The overview is still to wide but I dont really wanna fix it rn. The rest are improvements nonetheless and there should be no changes for vertical screens so this can be merged if there are no issues.

end-4:
Good idea, but I think it could be done better
1. Errors:
  ```
  (com.github.Aylur.ags:5589): Gtk-WARNING **: 06:48:04.209: Attempting to add a widget with type Ags_Revealer to a 
container of type Ags_Box, but the widget is already inside a container of type Ags_Box, please remove the widget from its existing container first.
  ```
2. Why make a "short" mode when we can just hide components of the normal bar?
3. I think maybe there's a way to get the font size for a more educated guess whether or not we should declutter the bar

myryk31415:
Oh I didn't even think of that. Guess that's a better solution.

Using the font size and the width of the elements to check whether that's wider than the screen could be possible.
Than it might be possible to hide more and more elements depending on the screen width. That would fix work as a fix for the overview as well. Imma look into it when I have time.

end-4:
we've moved from ags to quickshell so this is no longer valid, and the new quickshell version already has "short" mode for narrow screens
thanks and sorry

myryk31415:
Not like I was actively working on it :D. And if there is a short mode already even better.

PR #587, opened 2024-06-06
Search customizing

myryk31415:
One bugfix: if the blocked website array is empty there is no `-site:` anymore

And added user options to turn search features on and off.
Also there is a new NoResultsButton (maybe it shouldn't be a button lol) so the search result list is never completly empty.

Also I'm wondering whether the Buttons could be changed so they can also be used when pressing enter, so all the conditions in onAccept could be skipped. Might try some stuff tomorrow (its 3:46am)

### DIFF
--- a/.config/ags/modules/overview/windowcontent.js
+++ b/.config/ags/modules/overview/windowcontent.js
@@ -8,7 +8,7 @@ const { execAsync, exec } = Utils;
 import { execAndClose, expandTilde, hasUnterminatedBackslash, couldBeMath, launchCustomCommand, ls } from './miscfunctions.js';
 import {
     CalculationResultButton, CustomCommandButton, DirectoryButton,
-    DesktopEntryButton, ExecuteCommandButton, SearchButton, AiButton
+    DesktopEntryButton, ExecuteCommandButton, SearchButton, AiButton, NoResultButton,
 } from './searchbuttons.js';
 import { checkKeybind } from '../.widgetutils/keybind.js';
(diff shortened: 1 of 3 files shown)


myryk31415:
Now instead of all the checks it just executes the "action" of the first item in the list

myryk31415:
Now the order of results and the actions are just in one place: the `onChange` function for the order of the items, and searchbuttons file for the actions

myryk31415:
This makes further features like maybe a configurable order of the results easier as well.

myryk31415:
Should be good to go now, unless some issue arises from taking all that stuff out?

myryk31415:
1. Are you just busy or are there any problems with the changes? :)

.config/ags/modules/overview/windowcontent.js
```diff
@@ -162,22 +119,22 @@ export const SearchAndWindows = () => {
             _appSearchResults = Applications.query(text);
 
             // Calculate
-            if (couldBeMath(text)) { // Eval on typing is dangerous; this is a small workaround.
+            if (userOptions.search.enableMathResults && couldBeMath(text)) { // Eval on typing is dangerous; this is a small workaround.
```
end-4, line 122: hmm copy pasted comment. my bad

end-4 (approved):
code looks good

myryk31415:
Great :). Have a good day

end-4:
repeating "enable..." options felt weird so i changed those

PR #597, opened 2024-06-10
Adding mute indicator to ags

Soliprem:
It's already present in the hyprland config, but right now it hidden by the indicator. Really small fix. I included the icon in the text, but I guess it could also be added to assets and be handled from there, if it causes issues. On my machine it works fine, though


Soliprem:
I wanted (and tried) to add it using ../.commonwidgets/MaterialIcons, but I couldn't manage to understand it right now, and it's exams season. If it's an easy fix, it'd probably be the best course of action

l4v3nx:
smth like this works fine https://github.com/l4v3nx/dots-hyprland/commit/a3a87a3440081aa112157df5c9141d0095d7e7f1

Soliprem:
I think the taskbar indicator serves the purpose better

PR #601, opened 2024-06-10
Keybind path

myryk31415:
I added the option to change the path to the keybind.conf (since mine is not at the same place).
Also included a check so people get a notification if they entered an incorrect path. `~` in the path now works as well.

The keybindList and keybindListfunc probably be done simpler I just didn't know how.

### DIFF
--- a/.config/ags/modules/.configuration/user_options.js
+++ b/.config/ags/modules/.configuration/user_options.js
@@ -56,8 +56,10 @@ let configOptions = {
             'default': "auto",
         },
     },
-    'configPaths': {
-        'keybinds': "", //custom path to keybinds.conf for the cheatsheet
+    'cheatsheet': {
+        'keybinds': {
+            'configPath': "" // Path to hyprland keybind config file. Leave empty for default (~/.config/hypr/hyprland/keybinds.conf)
+        }
     },
     'gaming': {
         'crosshair': {
--- a/.config/ags/modules/.configuration/user_options.js
+++ b/.config/ags/modules/.configuration/user_options.js
@@ -56,6 +56,9 @@ let configOptions = {
             'default': "auto",
         },
     },
+    'configPaths': {
+        'keybinds': "",
+    },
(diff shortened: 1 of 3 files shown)


.config/ags/modules/.configuration/user_options.js
```diff
@@ -56,6 +56,9 @@ let configOptions = {
             'default': "auto",
         },
     },
+    'configPaths': {
```
end-4: idk why but i don't like this section name

.config/ags/modules/cheatsheet/keybinds.js
```diff
@@ -5,13 +5,22 @@ import Widget from "resource:///com/github/Aylur/ags/widget.js";
...
-const HYPRLAND_KEYBIND_CONFIG_FILE = `${GLib.get_user_config_dir()}/hypr/hyprland/keybinds.conf`;
+const HYPRLAND_KEYBIND_CONFIG_FILE = userOptions.configPaths.keybinds ?
+    userOptions.configPaths.keybinds : `${GLib.get_user_config_dir()}/hypr/hyprland/keybinds.conf`;
 const KEYBIND_SECTIONS_PER_PAGE = 3;
-const keybindList = JSON.parse(
-    Utils.exec(
-        `${App.configDir}/scripts/hyprland/get_keybinds.py --path ${HYPRLAND_KEYBIND_CONFIG_FILE}`,
-    ),
-);
+const keybindListfunc = () => {
```
end-4: ugh let's call this `getKeybindList`?

.config/ags/modules/.configuration/user_options.js
myryk31415: what would you name it? I tried to make it a group where other custom paths fit as well if ever needed

myryk31415:
`'customPaths'`, `'fileLocations'`, `'filePaths'`, `'customDirectorys'` or rather something completely different?

Issue #513, opened 2024-05-19
[Feature] Manage VPN connections


Cu3PO42:
**What would you like to be added?**

I'd love for the control panel on the right to also offer the option to connect to VPNs cofigured via Network Manager. I imagine it looking very similar to the menu already present for Wi-Fi connections, where clicking one would trigger a connection.

**How will it help**

I have several (corporate) VPNs that I might connect to (and disconnect from) multiple times a day. I currently use nm-applet for VPNs, but since Illogical Impulse already covers other networking, it would be nice to centraiize all network config there.

Ghanshyam-shaktawat:
hey greate idea. I also needed something like this as i have stuff i do with vpn. So I create my own version of it. I just create a toggle button on the right side menu next to the toggle wifi button which use the ags networking widget to connect to a praticular vpn connection. The vpn uuid is needed to be used for connection which i stored in userOptions.ags file. If u want i can share the code as a temperarly solution.

Altough about listening the vpn connections on the sidebar like wifi  is not hard to do.  Using nmcli we get get different vpn's and there uuid's which we can used to connect to with the help of ags network widget. I can implementing it and making a pr but i got exams and will be free after 3 June.

Cu3PO42:
You don't even need to invoke `nmcli`, you can directly use the NM library via GIR. It's also wrapped by AGS. What will take some care is that connecting to VPNs can also incur various prompts, for example providing credentials for external PKCS11 devices.

pandalec:
Yeah like GNOME is doing it, e.g. https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/6656

hoovad:
https://github.com/end-4/dots-hyprland/pull/1813
?

mambucodev:
+1

PR #614, opened 2024-06-15
[Feature] VPN Configuration

ArkjuniorK:
Enable VPN configuration from side right and display status icon whenever VPN connected


ArkjuniorK:
Looks like someone has same feature request [#513](https://github.com/end-4/dots-hyprland/issues/513)

alr86:
Change icon to "key" instead of "lock"(like Android)

enessmr:
Put the VPN opener/closer when lmb and configure vpn when rmb in quicktoggles.. the indicator to sideright of bar so it  detects when a VPN is running it shows the key icon

AD-Archer (approved):

end-4:
no longer valid after the transition from ags to quickshell
sorry for never getting to reviewing this and thanks

PR #625, opened 2024-06-17
Increase the idle and suspend time for user convenience

chrollorifat:
3/4 minutes is such a inconvenience as a idle and suspend time. So I increased it to a fair number.

### DIFF
--- a/.config/hypr/hypridle.conf
+++ b/.config/hypr/hypridle.conf
@@ -7,17 +7,17 @@ general {
 }
 
 listener {
-    timeout = 180 # 3mins
+    timeout = 2340 # 39 mins
     on-timeout = $lock_cmd
 }
 
 listener {
-    timeout = 240 # 4mins
+    timeout = 2400 # 40 mins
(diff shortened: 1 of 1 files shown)


end-4:
It's subjective.
You're free to set it on your system, but I do not want these values. Thank you.

chrollorifat:
okay, but why so passive-aggressive, bro?!

end-4:
i figured i'd use proper capitalization/punctuation when refusing in general so clowns wouldn't have a chance to mess with me, but i didn't think it would appear as too serious and interpreted as agressive lol

chrollorifat:
haha...It caught me off-guard..

chrollorifat:
@end-4 btw, is there a telegram or discord where we can discuss/talk about issues before engaging in github?!

end-4:
i don't use telegram
as for discord, (1) idk how much interest there would be and (2) @clsty wouldn't be able to join since he's in china. you can dm me though (`@end_4`)

PR #626, opened 2024-06-17
hypridle: fix monitor not turning off after manual lock

moetayuko:
The dpms off timer is not triggering for me after manually locking the screen via Win+L shortcut, and this commit magically fixes it. It just works, dunno why.

Derived from the example at the bottom of https://wiki.hyprland.org/Hypr-Ecosystem/hypridle/


end-4:
magic indeed

PR #629, opened 2024-06-18
Add Dependency to Build hyprlock-git

GalaXai:
Hyprlock-git won't build without hyprutils, which is first built in line 141 of `install.sh`:
```bash
v yay $hyprland_installflags --asdeps hyprutils-git hyprlang-git hyprcursor-git hyprwayland-scanner-git
```
This occurs after the hyprlock installation.
[Link to line 141 in install.sh](../blob/main/install.sh#L141)

Is there a reason to keep the installation on line 141 after the change?


clsty:
Thanks.

> Is there a reason to keep the installation on line 141 after the change?

Nah I'm just not having enough time to perfect it.

PR #630, opened 2024-06-18
Misc cleanup

moetayuko:


install.sh
clsty, line 1: I don't understand, why add `/usr/local/lib/` things? I can't find all of the same files inside `/usr/lib` after installing those local PKGBUILDs, e.g. `/usr/lib/Gvc-1.0.typelib`.

moetayuko: they're leftovers of `install-ags`

clsty (approved):
That's fine for me.

Though some other things related to AGS needs to be reviewed by @end-4 .

end-4 (commented):
AGS stuff are alright

end-4:
Good
I used gojq because I wanted my bash workspace indicator script to be fast, but now I don't use bash for that anymore so it's nice to use jq instead of gojq to reduce an AUR dependency

moetayuko:
> Good I used gojq because I wanted my bash workspace indicator script to be fast, but now I don't use bash for that anymore so it's nice to use jq instead of gojq to reduce an AUR dependency

you might want to switch to jaq for speed, which is available in Arch extra as well. https://github.com/01mf02/jaq?tab=readme-ov-file#performance

end-4:
iirc i had issues with jaq how it's args were a bit different... i could be wrong and can't be bothered to investigate at this point

moetayuko:
@clsty I just noticed that `illogical-impulse-basic` was not rebuilt by the script because pkgver was not updated. We need to either bump pkgver of these meta-packages when updating, or pass `-f` to makepkg. Which is preferred in your sense?

clsty:
> @clsty I just noticed that `illogical-impulse-basic` was not rebuilt by the script because pkgver was not updated. We need to either bump pkgver of these meta-packages when updating, or pass `-f` to makepkg. Which is preferred in your sense?

Sorry for late response. The former one sounds better to me, though I don't know how to achieve that. PR is welcome.

moetayuko:
> > @clsty I just noticed that `illogical-impulse-basic` was not rebuilt by the script because pkgver was not updated. We need to either bump pkgver of these meta-packages when updating, or pass `-f` to makepkg. Which is preferred in your sense?
> 
> Sorry for late response. The former one sounds better to me, though I don't know how to achieve that. PR is welcome.

Increasing pkgrel every time after updating pkgbuild should be sufficient.

PR #637, opened 2024-06-22
changed cheatsheet to appear only on active workspace

aliqut:
When the cheat sheet is on multiple monitors, using the PgUp and PgDn keys to cycle through the pages doesn't behave as expected. The page switch is sent to the different monitors randomly, and I have to spam the key until it happens to flip the page on the active workspace.

Another possible solution is making the AGS script switch pages on all monitors whenever the keys are pressed.


end-4:
> making the AGS script switch pages on all monitors whenever the keys are pressed

this is better actually, and it's not hard
i'll do it

end-4:
done

PR #653, opened 2024-06-26
fix duplcate line

AmmarHaddadi:
- remove duplicate line
+ add the actual launch command


end-4:
the duplication is intentional
it makes it so that the cheatsheet shows the important binds first to give newcomers a better experience

PR #654, opened 2024-06-27
get_keybinds.py now only checks for "bind="

myryk31415:
I tried to make a submap in my config and ags didn't work anymore since the python script couldn't handle the submap rule. This should fix any configs where there is not only keybind rules in the file. Also works with whitespace in front of the line.

Are there any other rules than "bind" that need to be parsed by the script as well?


end-4:
thanks

PR #673, opened 2024-07-04
Custom wallpaper picker

myryk31415:
This is still WIP, I already opened the pull request to get feedback and discuss issues.

#### What this implements:
A window below the bar that lets you scroll through wallpapers and pick one.
Also makes some changes to the colorgen related scripts so user options `smart` and `popup` can be passed.

#### Stuff that still needs to be done:

- [x] Perfomance: ags startup takes way too long when many pictures are in the wallpaper folder
    - Already way better with loading the pixbuf manually
    - Decoupled loading time from startup
    - Background loading implemented :D
- [x] Scrolling only works on a veeery tiny part at the bottom of the window
- [x] Rounded corners act weird with xray / blur
- [x] Window does not close if clicking elsewhere on screen
- [x] ~Gifs have no thumbnails~
- [x] ~Make gifs distinguishable~
- [ ] some visual polishing
- [x] Wallpaper change on interval is missing
- [x] handle empty folder

#### Things I couldn't get to work
- Gifs won't show up since I can't find an efficient way to load them
- Found no way to make the corners of the images rounded
- (Both of the above work with css background but thats way too resource intensive)


myryk31415:
Also I noticed today that there is a wallpaper service which isn't being used, whats up with that?

end-4:
> Also I noticed today that there is a wallpaper service which isn't being used, whats up with that?

i tried to use ags for wallpaper, but realized that causes a lot of lag and stopped using it

myryk31415:
ah I see. I don't have any issues when ags is running but startup takes like 15 seconds. If I can't fix that its kind of useless

myryk31415:
Did you ever have issues with horizontal Scrollables? If I make the images stack vertically then scrolling works as expected, if I stack them horizontally I can only scroll at the bottom of the window. (Presumably where the scroll bar is.) Scrolling with SHIFT doesn't work either. Also whenever I set `hscroll` to `always` or `automatic`  and the `Scrollable` is the child of a `box` (e.g when using `PopupWindow`) the whole content just disappears and its just an empty window. 

Imma open an issue at ags as well ig maybe they know.

myryk31415:
Lmao if you comment out the whole main.scss scrolling sideways works again. But if I put just a single line of CSS it's broken again. I'll mess around some more maybe I find a workaround. Also with a touchpad it works as well 🤷

myryk31415:
it works now??? Both scrolling with shift and using with PopupWindow.  Even when reverting to an old commit. Still same version of ags. Maybe it was an issue with gtk...

myryk31415:
I didn't even do a system update inbetween. No clue what i changed

end-4:
Since theme change doesn't take effect immediately for GTK apps, I'd rather not have wallpaper automatically change by default

myryk31415:
Okay sure that's reasonable. I'm not at home so I won't be working as much on this in the next months sadly. Also don't have a vertical monitor here, so that's gonna have to wait as well. Maybe I can use the time to read up on how gtk gallery/file apps manage thumbnails.

myryk31415:
You don't happen to know any way to load stuff in the background? I foud Utils.idle() but when I use that to update either the list of filenames or the array of widgets it just does it when starting up.

RN whenever opening the picker first time after restarting ags it takes like 10s to load my 160 pics. Afterwards it works pretty fast. I need to find some way to work around that, e.g. like filemanagers that show a placeholder until the thumbnail loads. Tried some approaches but didn't find anything good yet. Utils.idle() works with setting the css background but that takes up way too much ram (multiple Gbs)...

end-4:
> You don't happen to know any way to load stuff in the background?

i don't

myryk31415:
Whats the default wallpaper folder? Set it to `~/Pictures/wallpapers` as of now.

Also for gifs there is three options:
1) Leave them out entirely
2) Show them without thumbnail, only name
3) Set them as css background, which takes up a lot of resources

Probably 2) is the most reasonable option

end-4:
i say (2) *with the .gif icon*

Pharmaracist:
just use magick to make thumbnail and show it and link it to the file

end-4:
Since we switched from AGS to Quickshell this is no longer valid

Issue #690, opened 2024-07-14
[Issue] overview-tasks-workspace alignment issues


kainzilla:
- I have read the Usage and Troubleshooting pages of the wiki: https://end-4.github.io/dots-hyprland-wiki/en/i-i/03troubleshooting/: Yes
- I have made sure that both my config and system packages are up to date: Yes
- Linux distro: CachyOS

---
## The issue:

I believe I have potential fixes for these two issues and will submit a PR. The two issues observed:

## Issue 1:
- The overview-tasks-workspace boxes do not restrict themselves to the screen shape, and will expand horizontally to fill more space than they are supposed to, causing gaps on the right side of workspace overviews.
- This issue appears to resemble / look like issues #424 and #537, but is occurring for me regardless of scaling settings.
- How large the additional space is can vary greatly, and tends to happen more with smaller overview scale and less "columns" in the overview.
<details>
<summary>Issue 1 Screencap Sample</summary>

![hexpand-issue](https://github.com/user-attachments/assets/90b159d8-92d1-4db9-a6ab-384c7988d1d6)

</details>

## Issue 2:
- When the above issue is fixed, it becomes visible that at some values of `userOptions.overview.scale` where the division isn't favorable, a single pixel can clip on the right and bottom edges of the overview boxes:
<details>
<summary>Issue 2 Screencap Sample</summary>

![edge-issue](https://github.com/user-attachments/assets/7a788d1d-56d3-45a1-a931-e803c766c697)

</details>

<details>
<summary>Sample with Fixes Applied</summary>

![fixed](https://github.com/user-attachments/assets/aa3a429a-f309-4872-8986-44dbc8ad82d2)

</details>

Some details asked for in those other issues in case it helps:

- Multiple monitors: No
- Monitor resolution: 3840x2160 @ 120Hz
- Issue appears with and without scaling:
  - Tested with scaling:
    - Hyprland Scale: 2
    - Hyprland `force_zero_scaling`: true
    - `GDK_SCALE`: 2
    - `userOptions.monitors.scaleMethod` tested with `division` and `gdk`
    - Edge clip could show up at various `userOptions.overview.scale` scales, screencap sample was 0.18
  - Tested without scaling - same issue observed:
    - Hyprland Scale: 1
    - Hyprland `force_zero_scaling`: Unset (false)
    - `GDK_SCALE`: Unset (1)
    - `userOptions.monitors.scaleMethod` tested with `division` and `gdk`
[...]

PR #691, opened 2024-07-14
overview-tasks-workspace alignment fixes

kainzilla:
PR for issue #690

### hpack:

The addition of `hpack` to Widget.Revealer seems to fix the horizontal over-expansion issue.

While testing possible fixes, removing `hexpand` from `overview-tasks-workspace` and setting the size of the children also fixed over-expansion, but during animation (such as switching from overview to search results) the unbounded horizontal expansion would trigger and make the animation look bad.

I couldn't find any animation or sizing issues with the `hpack` solution at any scale or size, including very tiny overviews with few windows and large overviews with many windows.

### overview-tasks-workspace min width/height to fix right and bottom edges:

This might not be the "most correct" fix from a math or simplicity perspective, because there might be values somewhere else not being floored or rounded causing the windows inside overview to sometimes overlap the right and bottom edges.

In spite of that, rounding the multiplication and then adding one single pixel appeared to get it to line up precisely at various scales. I was never able to catch it landing +/- 1px in either incorrect direction, and I peeped a lot of pixels while testing this.

<details>
<summary>Sample screencaps from issue:</summary>

### hpack issue:
hexpand-issue

### overview-tasks-workspace min width/height issue:
edge-issue

### Fixes applied:
fixed

</details>

### DIFF
--- a/.config/ags/modules/overview/overview_hyprland.js
+++ b/.config/ags/modules/overview/overview_hyprland.js
@@ -230,7 +230,7 @@ export default (overviewMonitor = 0) => {
         const widget = Widget.Box({
             className: 'overview-tasks-workspace',
             vpack: 'center',
-            // Adding 1px to minimum width/height to accomodate 1px padding space for contents:
+            // Rounding and adding 1px to minimum width/height to work around scaling inaccuracy:
             css: `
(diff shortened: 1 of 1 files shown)


bThiiago:
this pr needs to be approved

with default code:
image

with this pr:
image

end-4:
`¯\_(ツ)_/¯` works

PR #700, opened 2024-07-19
New Tool for Changing Screen Resolution

nicefaa6waa:
This pull request introduces a new tool to streamline the process of changing screen resolution. Previously, adjusting screen scaling could result in misalignment or issues with AGS widgets, particularly with high-resolution displays. This tool aims to provide a more reliable method for setting resolutions, which addresses the scaling issues experienced by users with high-resolution screens.

### DIFF
--- a/.config/ags/modules/sideleft/tools/changeres.sh
+++ b/.config/ags/modules/sideleft/tools/changeres.sh
@@ -33,9 +33,20 @@ update_resolution_config() {
         exit 1
     fi
 
+    # Extract the resolution and refresh rate from the modeline
+    local resolution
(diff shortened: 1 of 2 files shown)


end-4:
two problems:
- for the resolution it writes something like 1366x768_60 instead of 1366x768@60
- this script isn't related to Arch so don't use its icon and show only on that distro. the entry in quickscripts.js should look something like this:
```js
    {
        icon: 'desktop-symbolic',
        name: 'Change resolution',
        command: `bash .config/ags/modules/sideleft/tools/changeres.sh`,
        enabled: true,
    },
```

nicefaa6waa:
Both of the problems should be fixed now

end-4:
thanks!

PR #715, opened 2024-07-30
Fix for keyboard layout indicator not showing up on all monitors

hossein-mo:
The keyboard layout indicator was only visible on one monitor because there was just one instance of it. I’m not totally sure my fix is the best way to do it, especially since I don't know JavaScript, but it's now visible on all monitors.


end-4:
thanks
can't test but it doesn't break anything and code looks alright

PR #717, opened 2024-07-31
tray: filter out empty tray items

moetayuko:
For unknown reason, an unknown program (even invisible in D-Spy) occasionally populates an empty tray item of the following properties:
```js
TrayItem {
    "category": null,
    "id": null,
    "title": null,
    "status": null,
    "window-id": null,
    "is-menu": null,
    "tooltip-markup": "",
    "icon": "image-missing"
}
```

Add a null check for the item id to workaround the issue.


end-4:
i don't have that weird item, but sure ig

PR #736, opened 2024-08-10
[Feature] Change shell mode per monitor

c-danil0o:
**Multi-monitor improvement.** 

- Users can now change bar shell mode ("normal", "focus", "nothing") per monitor. That way each monitor can have separate bar look. 

I use it to hide the bar on my second monitor while it stays shown on my main monitor.

I plan to improve this by adding bar monitor config to user_options.js so that selected bar look persists between sessions.

### DIFF
--- a/.config/ags/modules/.configuration/user_options.js
+++ b/.config/ags/modules/.configuration/user_options.js
@@ -222,9 +222,10 @@ let configOptions = {
         }
     },
     'bar': {
-        // Array of bar styles for each monitor.
+        // Array of bar modes for each monitor. Hit Ctrl+Alt+Slash to cycle.
+        // Modes: "normal", "focus" (workspace indicator only), "nothing"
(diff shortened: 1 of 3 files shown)


c-danil0o:
- There is now option in userOptions.js to set up default bar look for each monitor.

.config/ags/modules/.configuration/user_options.js
```diff
@@ -221,6 +221,11 @@ let configOptions = {
             'cycleTab': "Ctrl+Tab",
         }
     },
+    'bar': {
+        // Array of bar styles for each monitor.
+        // Example for four monitors: ["normal", "focus", "normal", "nothing"]
+        'monitors': ["normal"]
+    },
```
end-4: let's have this in `bar.modes`, just in case we have another option for bar later on that's also configured per monitor

end-4 (changes requested):
cool

PR #756, opened 2024-08-20
install.sh : Add an option to create config and local backup

shinyPy:
Self explanatory


clsty:
Thank you for the work!

PR #764, opened 2024-08-22
keyboard auto swap on lock

s0mesh1t:
many people do not use English as their native language, and the password is mostly English, and it is inconvenient when the layout is in another language and you need to enter the password in English. It will automatically switch to the English layout when the screen is locked


end-4:
This is a good idea, but I don't think you're supposed to include your own customizations and a direct Russian translation?

PR #796, opened 2024-09-05
Add kitty support

pandalec:
Maybe someone could use it. Added apply_kitty() and a template which will be executed if kitty is installed. It generates a current_theme.conf which must be included within kitty.conf (AFAIK default if a theme is used in kitty). The other changes are related to my linter.

### DIFF
--- a/.config/ags/scripts/templates/kitty/dots-hyprland.conf
+++ b/.config/ags/scripts/templates/kitty/dots-hyprland.conf
@@ -1,0 +1,80 @@
+# vim:ft=kitty
+
+## name:     dots-hyperland
+## author:   pandalec
+## license:  MIT
+## upstream: https://github.com/end-4/dots-hyprland
+## blurb:    Template for generating a kitty theme based on end-4s dots-hyprland
+
+
+
+# The basic colors
+foreground              #$term7 #
+background              #$term0 #
+selection_foreground    #$onSurfaceVariant #
+selection_background    #$surfaceVariant #
+
+# Cursor colors
+cursor                  #$term1 #
+cursor_text_color       #$background #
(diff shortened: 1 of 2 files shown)


Filz0r:
Hello I would appreciate if this patch was applied when ever possible, thanks! :pray:

.config/ags/scripts/templates/kitty/dots-hyprland.conf
```diff
@@ -0,0 +1,80 @@
+# vim:ft=kitty
+
+## name:     dots-hyperland
```
end-4, line 3: hyprland doesn't contain an "e" lol

end-4 (commented):
I tested. It kind of works, but the background is always pitch black. This doesn't match the theme and makes the terminal unreadable when using light theme

enessmr:
Theres a background thing. I need to look at catppuccin theme for kitty.
Background would be like foot, but converted to hex. It reloads the kitty config everytime with sending ghost keys to the terminal for quarter of a second. You could not put your mouse then press a key when you disable dark mode since its so short. Same with others. Generated colors converted to hex, then applied to theme.

nizaralheet:
> I tested. It kind of works, but the background is always pitch black. This doesn't match the theme and makes the terminal unreadable when using light theme
@enessmr 

i think it got kind off solution for that and already made a PR #1123 it could be solved easliy

end-4:
since #1123 is merged, all left for kitty support is to make sure it uses fish

end-4:
now guess which is kitty :p
image

end-4:
anyway it's done so i'll close this. thanks

PR #807, opened 2024-09-10
ags(volumeindicator): detect Headset

refurbishing:
When using a Headset the Volume Widget it will show it as Speakers this PR fixes that.
image


refurbishing:
closing due to switch to quickshell

PR #822, opened 2024-09-16
ags & hypr: make session manager toggleable & add power button keybind

verysillycat:
* pressing the power button on your laptop keyboard will bring up the session manager
* make the session manager toggleable.

**things to point up:**
* the shortcut `XF86PowerOff` will only work if you have the following in `/etc/systemd/logind.conf`
```
PowerKeyIgnoreInhibited=yes
HandlePowerKey=ignore
HandlePowerKeyLongPress=ignore
```
else the system will shutdown immediately clicking it.


refurbishing:
closing due to switch to quickshell

Issue #334, opened 2024-03-15
[Feature] Value indicator popup: Display sink output mute state


yrnmsk:
**What would you like to be added?**

Anything that can help a user recognize that the current sink output is muted (e.g.: headphones/speakers/other bluetooth device).

Right now, it just displays the current volume value regardless of the sink mute state.

**How will it help**

It would help users who frequently toggle mute/unmute their sink outputs for whatever valid reasons they have. Raising/lowering the volume of the sink output doesn't necessarily unmute it if it is muted, so it would be useful if we could somehow display current sink mute state.

**Extra info**

You might already know this since we talked about this already before, but based on the AGS docs, I got mute state (boolean) via `Audio.speaker?.stream?.isMuted`.

The last time I did it on my own, I just displayed 0 both for the number label and the slider value, so it could be improved. For example, an icon that resembles a mute state, or some scss shenanigans that can help visually imply a disabled effect for the slider and the number label, or whatever else.

xxCLIMBINGMANxx:
I have found a bit of a hacky way that seems to provide the necessary functionality. You can add the changes from [this](https://github.com/end-4/dots-hyprland/pull/844) pr and then change the line `if(updateValue === 0) {` to `if(Audio.speaker?.stream?.isMuted) {`. This preserves the volume level display but adds a mute icon for when the sink is muted.

PR #844, opened 2024-10-02
feat: Add volume_off icon for volume indicator

Krayon0:
#334 
Bit of a hacky way of doing it
`include @titlefont` on `osd-value-txt` breaks the icon from appearing for some reason - Not very familiar with scss

image


xYc4t (approved):

Krayon0:
welp forgot this still existed - cba recreating the commit
someone else can remake the PR if they like

PR #868, opened 2024-10-15
Move dwindle config "no gaps when only" to workspace rules

b4shful:
Updated hyprland today and noticed that I was getting an error/red error message at the top of the screen as I had `no_gaps_when_only = 1` uncommented in the dwindle config section of `general.conf`

Turns out that setting has been removed, and the new method to achieve this setting is using workspace rules.

The rules I added were taken from a recent update to the Hyprland default/example config (https://github.com/hyprwm/Hyprland/commit/8e51a36c7fd500e865e34f08ab4dc4331dca59cf). Since there don't seem to be any existing workspace rules, I added a section for them to the rules file, and then added the example code verbatim (with one extra comment line explaining that it replaces the old dwindle config setting).

As the `no_gaps_when_only = 1` is commented out in the current dots-hyprland general.conf file, I have kept the "smart gaps" lines commented in their new position, so this shouldn't change any default functionality. My commit just removes a commented config setting that no longer exists and adds a commented replacement for it.

### DIFF
--- a/.config/hypr/hyprland/general.conf
+++ b/.config/hypr/hyprland/general.conf
@@ -60,7 +60,6 @@ general {
 
 dwindle {
 	preserve_split = true
-        # no_gaps_when_only = 1
 	smart_split = false
 	smart_resizing = false
 }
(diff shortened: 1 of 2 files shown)


b4shful:
Updated again due to https://github.com/hyprwm/Hyprland/commit/dca75db127fedc58fc85ae0e6e47162e3d5d16f9

clsty (approved):

clsty:
This PR only involves fixes to comment and it has direct source from official document so I'll just merge it quickly.

Thank you for contribution.

PR #874, opened 2024-10-17
Added an option to appear tray on mouse hover

SuperSuslik312:
This can be useful if you have a lot of apps in tray
(sorry for recreate pr, shit happens)


end-4:
with the transition from ags to quickshell unfortunately this is no longer valid
sorry and thanks

PR #906, opened 2024-10-29
Update scriptdata/functions (dots installation script)

daUnknownCoder:
Update `rsync` function to skip long unnecessary `.git` and python `site-packages` folders

### DIFF
--- a/scriptdata/functions
+++ b/scriptdata/functions
@@ -76,12 +76,13 @@ function prevent_sudo_or_root(){
 }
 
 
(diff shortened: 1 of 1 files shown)


clsty:
Sorry, I don't think it's a good idea to skip them without giving user option. Closing.

daUnknownCoder:
Oh I forgot this pr even existed cuz it's too old, I don't know much bash, so may have skipped user convenience

PR #912, opened 2024-10-31
move gnome-bluetooth-3.0 and gammastep to ags dependencies

moetayuko:
* our ags configuration uses the bluetooth module so it can't start w/o gnome-bluetooth-3.0
* gammastep is used by ags sidebar and irrelevant to gnome


clsty:
Reasonable changes. I'll merge it.
Thanks for you contribution.

PR #919, opened 2024-11-02
Tab key now switches between APIs

sh1zicus:
https://github.com/user-attachments/assets/88580d26-8a46-4f77-8698-74369e614ea3

### DIFF
--- a/.config/ags/modules/sideleft/apiwidgets.js
+++ b/.config/ags/modules/sideleft/apiwidgets.js
@@ -90,6 +90,12 @@ export const chatEntry = TextView({
             self.placeholderText = (Gemini.key.length > 0 ? getString('Message Gemini...') : getString('Enter Google AI API Key...'));
         }, 'hasKey')
         .on("key-press-event", (widget, event) => {
+            // Переключение по Tab
+            if (event.get_keyval()[1] === Gdk.KEY_Tab) {
+                const nextId = (currentApiId + 1) % APIS.asyncGet().length;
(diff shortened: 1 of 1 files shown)


TwelveNight:
Why is there no return to the previous tab?
just
```js
        if (event.get_keyval()[1] === Gdk.KEY_ISO_Left_Tab) {
          const prevId =
            (currentApiId - 1 + APIS.asyncGet().length) %
            APIS.asyncGet().length;
          switchToTab(prevId);
          return true;
        }
```
🐶

s0mesh1t:
> Why is there no return to the previous tab? just
> 
> ```js
>         if (event.get_keyval()[1] === Gdk.KEY_ISO_Left_Tab) {
>           const prevId =
>             (currentApiId - 1 + APIS.asyncGet().length) %
>             APIS.asyncGet().length;
>           switchToTab(prevId);
>           return true;
>         }
> ```
> 
> 🐶

yeah, I combined both options,
I also want to do this in the left-right panel for centerModules,
but I haven’t figured it out yet

s0mesh1t:
found a problem
when switching to Tools and back to APIs the focus is lost and it stops working

end-4:
binding tab without modifier hurts keyboard-only navigation (idk who cares... but still...)

> found a problem
> when switching to Tools and back to APIs the focus is lost and it stops working

ugh yes this really is mysterious and i probably will rewrite the horrendous implementation of api pages some time

PR #969, opened 2024-11-25
xargs it to wl-copy instead

abdullah-albanna:
piping directly to wl-copy seems to be broken for some reason!
if I pipe anything to wl-copy, it will freeze and not do anything

using like super + v to view and select something from the  history, it will create a new process to copy, but it's not, it's freezing and will never close, over time these processes gets big and a lot of them

using xargs is generally better 

and it of course works

fell free to decline the pr if this is a me problem only


sh1zicus:
I dont have problem 👀

sh1zicus:
> @sh1zicus Archlinux?

CachyOS (Arch) gnome then install dots

Abdullah-Albanna:
> CachyOS (Arch) gnome then install dots

ah, might be an arch problem then, idk

using xargs won't hurt I guess, it's not a breaking change, just more compatibility.. maybe

end-4:
seems like just a you problem
ughhh im closing this now, sorry for not looking at this sooner

PR #977, opened 2024-11-29
Add Chinese translation for 'Feels like'

yuanczx:
## Default
Default

## zh-CN
zh-CN


clsty:
I'm okay with this translation, as a native speaker of Chinese.

nx-smul:
by the way how to add the weather can you help me ?

end-4:
> by the way how to add the weather can you help me ?

It's only a fallback widget for desktops or whatever systems that don't have a battery. There is no simple config option to show both battery and weather widgets. And the weather widget isn't really reliable... I can't recommend it over the battery indicator

And next time for help open a discussion instead of writing in an old issue please

PR #1006, opened 2024-12-18
Fix Layer Rules not being applied

UniOnTheMap:
Changes the layer rules to support the new RegEX rules.


end-4:
i ignored for too long... it's now already fixed
sorry and thanks

Issue #1033, opened 2025-01-04
[Issue] When enabling transparency everything breaks.


gaurav-null:
![image](https://github.com/user-attachments/assets/a218e11b-7a32-476f-841f-2e6583ba5699)
even after reinstalling using the script or running the install.sh it wont fix change

Shar-jeel-Sajid:
Hi Bro!! I also installed this today and faced the same problem. After trying to fix it for hours i finally was able to do it.
All you have to do is go to .config/ags/scss/mani.scss file. You will see an import 
@import 'material'
replace with 
@import './fallback/material'
after doing this just restart widgets with shortcut 
Super + Ctrl + R

Shar-jeel-Sajid:
Hi Bro!! I also installed this today and faced the same problem. After trying to fix it for hours i finally was able to do it.I have also created a pull request for it i don't know when it will be merged. But you can try to fix it yourself its very simple.

All you have to do is go to /home/.config/ags/scss/mani.scss file. You will see an import
@import 'material'
replace with
@import './fallback/material'
after doing this just restart widgets with shortcut
Super + Ctrl + R

Tell me if it worked for you.

gaurav-null:
yup but the colour scheme i think broke, instead of red it now blue. Also i just realised ctrl+c dosnt work in terminal

Marell01:
Yeah the fix works but colour scheme doesn't work now, even after applying a new wallpaper

AlpSha:
Seems like aur package materialyoucolor is not being detected anymore. I'm not saying that this is a perfect solution but here is a workaround.

Open your terminal and cd into ~/.config/ags/scripts/color_generation
Create a venv folder here by using the command:
`python -m venv ./venv`

Then activate it to install required packages (I use fish so):
`source venv/bin/activate.fish`

Now install these packages using pip: **pillow**, **materialyoucolor**
Then we need to tell our colorgen.sh script to source this venv. Paste the following line into the file colorgen.sh, right after line 8.
`source "$XDG_CONFIG_HOME/ags/scripts/color_generation/venv/bin/activate"`

Restart ags and choose a wallpaper. It will fix the issue.

PlexILIghT:
@AlpSha Can it be fix to my #1023 problem? Even tho after installing materialyou to system python, I fixed errors. But ags doesn't change colors accordingly anyway.

AlpSha:
[...]

PR #1034, opened 2025-01-04
Fixes issue#1033 

Shar-jeel-Sajid:
The importing of material in main.scss for ags was wrong wrong fixed it.

### DIFF
--- a/.config/ags/scss/main.scss
+++ b/.config/ags/scss/main.scss
@@ -5,7 +5,7 @@
 *:not(popover) { all: unset; }
 
 // Colors
-@import 'material'; // Material colors
+@import './fallback/material'; // Material colors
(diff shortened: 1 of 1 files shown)


Shar-jeel-Sajid:
This was a serious issue requires urgent merge because it breaks whole css.

Blade04208:
the problem is that this fix breaks material you styling and is more just a temporary patch than anything. should probably cook up an actual fix.

nx-smul:
ya'll maybe missing some dependancy recheck what's missing mine working fine

clsty:
I think this PR is not suitable as a solution, as discussed above.

We'll use uv for python thingy anyway, after applying #1060 .
#1033 should not be a problem hopefully.

Closing.

PR #1039, opened 2025-01-10
ags: fix color generation for materialyoucolor 2.0.10

moetayuko:


clsty:
I've noticed that @end-4 is not maintaining this repository recently. I appreciate his work before a lot.

However, we have 27 PRs and 219 issues currently, and lots of them are about color generation thingy. So, I suggest that we invite @moetayuko as a collaborator of this repo, since it seems that @moetayuko is capable of this and may resolve some problems for now.

What do you two think? @end-4 @moetayuko

end-4:
I'm fine with and would actually appreciate having @moetayuko as another collaborator who can handle AGS-related issues, especially considering my inactivity. @moetayuko, please let me know if you would like this position (ping me please so I get a notification)

---

About this PR, this seems like a trivial but important fix, but unfortunately I cannot test it now as I don't have Hyprland and my dotfiles installed on my system. So I'll trust you two. Thank you for your work!

moetayuko:
@clsty @end-4 Thanks for recognizing my PRs. As I am not an expert in AGS like @end-4 and have limited availability to address the remaining open issues, it is unlikely that I will be able to contribute significantly as a collaborator. While I may continue to submit minor bug fixes (like this one), I am also considering transitioning to an alternative preconfigured setup. Therefore, I must respectfully decline the offer.

---

Regarding this PR, yes I'm sure about its correctness, here's the upstream API change https://github.com/T-Dynamos/materialyoucolor-python/commit/f8a1d61162f19a22daa8eeebee97e76ae0f24e8c#diff-5e804c480f8ec8e161011ea0cdf418215f34090b95ab61e186d14348ff7ec3d4R10

clsty:
Okay, PR accepted. Thanks for your contribution.

PR #1047, opened 2025-01-14
Array boundary Check for Overview

edohusakovic:
This prevents the application overview from throwing an error when multiple monitors are plugged in but only one is enabled.


end-4:
alright... 
i'll add a note so i remember what this is for

PR #1049, opened 2025-01-15
Fix overview widget for multi monitors

### LINKED ISSUE #772: Overview bugged on multi monitor systems
- I have read the Usage and Troubleshooting pages of the wiki: https://end-4.github.io/dots-hyprland-wiki/en/i-i/03troubleshooting/
- I have made sure that both my config and system packages are up to date
- Linux distro: Arch Linux

---
## The issue

Opening the overview whilst having two monitors not the same resolution, in my case one being 1080p and the other being 4K with a scale of 1.5, causes the overview to become oversized and uncentered on the smaller monitor whilst remaining perfect on the larger.

**1920x1080:**
image

**3840x2160:**
image

**Sidenote:**
When attempting to take a "Screen snip" the overview will squash itself on the X axis to what it should be but leaves the Y axis the size it was when it was bugged leading to a weird output as shown below.

image

Toshimichi0915:
Overview opened in the main window.

image

Overview opened in another window.

image

Closes #772


clsty:
This seems cool and should be an important fix. It's working fine on my device.
@end-4 Please considering review the PR when possible, thanks!

notfatdan:
doesn't take into account vertical monitors but it's better than nothing. I wouldn't merge it until vertical monitor support works.

Toshimichi0915:
> doesn't take into account vertical monitors but it's better than nothing. I wouldn't merge it until vertical monitor support works.

Can you describe how it's bugged on vertical monitors.

end-4 (approved):
looks good

end-4:
I can't really immediately see anything inconsistent between width and height... AI tells me the font size calculation might be related, but I can't be sure since @notfatdan didn't really say specifically what's wrong
But it doesn't hurt to merge this now and handle the vertical monitor problem later, so I'll do just that

end-4:
I fixed the above in ecdc219356e0d54ae5e0be94eae7e22a6858ea8b
notfatdan, you can open a new issue if that didn't fix your problem

notfatdan:
sorry i should have added more information when i commented. That was the issue and it's all fixed now, perfect!

PR #1058, opened 2025-01-23
Add screen recording indicator to bar

aliqut:
Adds a red circle to the statusbar when wf-recorder is active. 

Not sure if there's a better way to check if wf-recorder is running other than polling every 2 secs.

### DIFF
--- a/.config/ags/modules/bar/main.js
+++ b/.config/ags/modules/bar/main.js
@@ -1,129 +1,142 @@
 const { Gtk } = imports.gi;
-import Widget from 'resource:///com/github/Aylur/ags/widget.js';
-import Battery from 'resource:///com/github/Aylur/ags/service/battery.js';
+import Widget from "resource:///com/github/Aylur/ags/widget.js";
+import Battery from "resource:///com/github/Aylur/ags/service/battery.js";
 
 import WindowTitle from "./normal/spaceleft.js";
+import RecordingIndicator from "./normal/recording.js";
 import Indicators from "./normal/spaceright.js";
 import Music from "./normal/music.js";
 import System from "./normal/system.js";
 import { enableClickthrough } from "../.widgetutils/clickthrough.js";
 import { RoundedCorner } from "../.commonwidgets/cairo_roundedcorner.js";
-import { currentShellMode } from '../../variables.js';
+import { currentShellMode } from "../../variables.js";
 
 const NormalOptionalWorkspaces = async () => {
+  try {
+    return (await import("./normal/workspaces_hyprland.js")).default();
+  } catch {
     try {
-        return (await import('./normal/workspaces_hyprland.js')).default();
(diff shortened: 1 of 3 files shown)


Chiron8:
Maybe having a script that runs when wf-recorder is started would be more efficient? I might test this soon... Nice idea!

aliqut:
> Maybe having a script that runs when wf-recorder is started would be more efficient? I might test this soon... Nice idea!

I'm not very comfortable with AGS, how would I go about doing this?

Chiron8:
> > Maybe having a script that runs when wf-recorder is started would be more efficient? I might test this soon... Nice idea!
> 
> 
> 
> I'm not very comfortable with AGS, how would I go about doing this? 

What I would probably do is add a line to the script that runs wf-recorder that changes an environment variable. Then When the ags bar updates (e.g. when the ram and cpu usage updates) check the value of the global variable? It should be quite simple to implement and shouldn't reduce performance.

Chiron8 (changes requested):
Sorry for not reviewing properly! I'll try and add a script soon. I feel polling the status of the screen recorder is quite inefficient.

Chiron8:
Okay, I've requested some changes to the service file. I tried to clean up the code a little and made a few changes which could slightly increase the efficiency. Unfortunately, I couldn't find a way to not have the code poll pgrep. I'm not great at javascript or CSS but it seems to be working well on my computer. It's my first time directly editing a pr so let me know if i've done anything wrong!

aliqut:
> > Okay, I've requested some changes to the service file. I tried to clean up the code a little and made a few changes which could slightly increase the efficiency. Unfortunately, I couldn't find a way to not have the code poll pgrep. I'm not great at javascript or CSS but it seems to be working well on my computer. It's my first time directly editing a pr so let me know if i've done anything wrong!
> 
> it's been more than 3 weeks, I'm trying to help you

Hi,

I've been quite busy, sorry for not responding at all. The code changes look alright, I can't approve them myself though. Need to wait for @end-4

clsty:
> Okay, I've requested some changes to the service file. I tried to clean up the code a little and made a few changes which could slightly increase the efficiency. Unfortunately, I couldn't find a way to not have the code poll pgrep. I'm not great at javascript or CSS but it seems to be working well on my computer. It's my first time directly editing a pr so let me know if i've done anything wrong!

I'm not seeing any editing on this PR, and I think end-4 can not see it either.
Since we can not see it, there's no way we could possibly approve it even if we would like to.

Chiron8:
> I'm not seeing any editing on this PR, and I think end-4 can not see it either.
> 
> Since we can not see it, there's no way we could possibly approve it even if we would like to.

Sorry about that. I'll try to fix that as soon as possible!

end-4:
unfortunately no longer valid after the transition from ags to quickshell, so i'm closing

PR #1060, opened 2025-01-24
feat: use uv to manage dependencies

### LINKED ISSUE #1017: [Feature] Use virtual env / Use less AUR packages
Currently there are 210 issues opened, and I've noticed that lots of them are about package installation failure or non-existence, often related to python package version thingy.

For myself, I've successfully installed the repo on two of my computers, but one of them break after a rolling update months ago (and I don't have time to deal with it so I just switch to waybar config temporarily).

I'm not good at python, but I know that python has some virtual environment thingy to deal with package isolation, which may make this repo easier, more stable and cleaner to install on various computers (and even on other distros).

(I'm somehow busy but I'll have a look at this question eventually.)

ZeyadMoustafaKamal:
closes #1017 

As mentioned in #1017 many users have reported that they had a problem with the installer and basically creating a virtual environment will fix everything. We will use `uv` since it's the new way for how stuff works in python.

#### what we will do
- [x] Modify the installer to install `uv` along with the other tools. And also creating the virtual environment itself. It should be in `$XDG_STATE_DIR`.
- [x] Add a `requirements.in` file that contains dependencies.
- [x]  Install all needed dependencies.
- [x] Modify all scripts to use the new python interpreter.
- [ ] Finally add a documentation that should explain how to deal with the new thing.


ZeyadMoustafaKamal:
For this part https://github.com/end-4/dots-hyprland/blob/41520aebc6f0bd5fe4e10e32e08f4817bce321c0/install.sh#L136-L143 

As you can see. All packages in the mentioned [PKGBUILD](https://github.com/end-4/dots-hyprland/blob/main/arch-packages/illogical-impulse-pymyc-aur/PKGBUILD) will be installed by `uv` except for `grandience`. Because it doesn't exists in the python package registry and the repo doesn't have a `setup.py` or a `pyproject.toml`. So we will leave it for yay. But as the problem mentioned [here](https://github.com/end-4/dots-hyprland/issues/428#issuecomment-2081690658) will no longer be an issue because we are installing everything in a virtual environment I think that we should basically remove this part. Is there anything that would force us to keep this part ?

clsty:
> Because it doesn't exists in the python package registry and the repo doesn't have a `setup.py` or a `pyproject.toml`. So we will leave it for yay.

Because that some issues are related to `gradience` too, I hope we can remove the use of yay for `gradience`.

`setup.py` or `pyproject.toml` is not existed, however, maybe(?) you can fork the [gradience](https://github.com/GradienceTeam/Gradience) repo and create the required files by understanding and converting manually its provided installation process, i.e. using meson:
- https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=gradience
- https://github.com/GradienceTeam/Gradience/blob/main/meson.build

I'm not sure whether this method will work. But we'd better avoid `yay` on `gradience` anyway.

> But as the problem mentioned [here](https://github.com/end-4/dots-hyprland/issues/428#issuecomment-2081690658) will no longer be an issue because we are installing everything in a virtual environment I think that we should basically remove this part. Is there anything that would force us to keep this part ?

We need a `gradience` as `gradience-cli` is used [here](https://github.com/end-4/dots-hyprland/blob/41520aebc6f0bd5fe4e10e32e08f4817bce321c0/.config/ags/scripts/color_generation/applycolor.sh#L152C5-L152C18), which is the only reason. The part of this bash script you quoted is not necessary and should be removed once `gradience` is available using an alternative installation process, such as using `uv`.

ZeyadMoustafaKamal:
> setup.py or pyproject.toml is not existed, however, maybe(?) you can fork the [gradience](https://github.com/GradienceTeam/Gradience) repo and create the required files by understanding and converting manually its provided installation process which uses meson:

Yeah I think that I will give it a try.

ZeyadMoustafaKamal:
@clsty Finally I got `gradience` to work in a fresh arch install inside a virtual environment. I didn't fully test it, I just typed `gradience-cli --help` and it worked. As I am not sure what this is doing in the first place. So I would like that you go and test it yourself. Go to [my forked repo](https://github.com/ZeyadMoustafaKamal/Gradience) and clone the branch I am working on in a fresh arch installation. And make sure that you have `uv`, `git`, and `meson` installed and then run these commands in the root directory of `gradience`

```
uv venv --prompt .venv -p 3.12
source .venv/bin/activate
bash install_deps.sh
uv pip install -r requirements.txt
meson setup build --prefix=$VIRTUAL_ENV
meson compile -C build
meson install -C build
```
Also you may get some error when you run the `meson setup` command. If you got it then just `rm` the `build` directory and then try again.

Then to run `gradience-cli` use this command

`GSETTINGS_SCHEMA_DIR=$VIRTUAL_ENV/share/glib-2.0/schemas gradience-cli`

You can export it as an environment variable or we might move the files there to the global path where `gradience` (or glib) searches to schemas by default. I would recommend the first approach as we want to isolate everything.

clsty:
I do not have enough space for a fresh arch install currently so I just did a quick test on my own platform. It stuck here:
```
❯ uv pip install -r requirements.txt
  × No solution found when resolving dependencies:
  ╰─▶ Because pyjobject was not found in the package registry and you require pyjobject, we can
      conclude that your requirements are unsatisfiable.
```

> As I am not sure what this is doing in the first place.

As I mentioned above, `gradience-cli` is used in [this file](https://github.com/end-4/dots-hyprland/blob/41520aebc6f0bd5fe4e10e32e08f4817bce321c0/.config/ags/scripts/color_generation/applycolor.sh#L152C5-L152C18). Specifically, when you press `Super+Ctrl+T`, you are prompted to select a wallpaper and then `gradience-cli` will be involved to generate a color scheme accordingly.

ZeyadMoustafaKamal:
OK I think that we are reaching to a good point here. Finally I have modified the install script to install all needed libraries in a virtual environment. And now all I need to test it which is what I am doing right now. Also I have fixed the `gradience` issue you have told me about, you may try it again now.

As you can see in some files [like this](https://github.com/end-4/dots-hyprland/blob/main/.config/ags/scripts/wayland-idle-inhibitor.py) that has a shebang. Like do we really need it ? It's kinda hard and problematic to use the new python interpreter. Like it should exists in `~/.local/state/ags/.venv/bin/python` and I can't point out to the path in the shebang. The only way I can fix this is by using a relative path in the shebang. Which is a problem since we should execute the files from specific directory and this will cause tons of problems. Or we can create a wrapper script to fix this. Or we might just not use a shebang in the first place. I don't see any sort of benefit from using the shebang in this situation. So why not remove it ? what you think ?

clsty:
According to [document](https://www.gnu.org/software/coreutils/manual/html_node/env-invocation.html), it's possible to point out a path like `~/.local/state/ags/.venv/bin/python` in shebang, using a little trick. Firstly make sure that global environment variable `ILLOGICAL_IMPULSE_VIRTUAL_ENV=$HOME/.local/state/ags/.venv` exists, and then as the shebang:
```sh
#!/usr/bin/env -S\_/bin/sh\_-xc\_"exec\_\$ILLOGICAL_IMPULSE_VIRTUAL_ENV/bin/python\_-E\_"\$0"\_"\$@""$ ./myscript.py
```
(Edit `myscript.py` to the script filename)

Other links maybe related:
- [blog post](https://zameermanji.com/blog/2022/9/17/specifying-a-relative-interpreter-in-a-shebang/)
- [stackoverflow question](https://stackoverflow.com/questions/44076804/what-shebang-to-use-for-python-scripts-run-under-a-pyenv-virtualenv)
------
> Or we might just not use a shebang in the first place. I don't see any sort of benefit from using the shebang in this situation. So why not remove it ? what you think ?

Well, if you are about to remove shebang, it means that you need to modify all the codes that involved the scripts, to use the virtual environment instead, right? For example in `~/.config/ags/modules/sideright/quicktoggles.js`:
```js
215:        if (self.attribute.enabled) Utils.execAsync(['bash', '-c', `pidof wayland-idle-inhibitor.py || ${App.configDir}/scripts/wayland-idle-inhibitor.py`]).catch(print)
216:        else Utils.execAsync('pkill -f wayland-idle-inhibitor.py').catch(print);
221:        self.attribute.enabled = !!exec('pidof wayland-idle-inhibitor.py');
```

See, it's not that easy, right? `pidof wayland-idle-inhibitor.py` and `pkill -f wayland-idle-inhibitor.py` may not work if you are not executing `wayland-idle-inhibitor.py` directly from shell, i.e. using shebang.

I'd rather just modify a single shebang in `wayland-idle-inhibitor.py` instead:
```
#!/usr/bin/env -S\_/bin/sh\_-xc\_"exec\_\$ILLOGICAL_IMPULSE_VIRTUAL_ENV/bin/python\_-E\_"\$0"\_"\$@""$ ./wayland-idle-inhibitor.py
```

scriptdata/dependencies.conf
clsty, line 1:
```
ibportal-gtk4
pygobject
```
These packages do not exist.

ZeyadMoustafaKamal: Fixed

ZeyadMoustafaKamal:
OK I finished everything in my local machine and also I have installed it in a fresh arch installation (I hope that you can try it) I only have one problem, ags is not working for some reason. I have `pkill`ed it and ran it again and this is all I am getting. I will push changes right now so you can try it and reproduce the problem. Like maybe the way I am creating the VM is the problem ?

image

clsty:
Ok, I think I've got it: the shebang should be:
```
#!/usr/bin/env -vS\_/bin/sh\_-xc\_"source\_\$(eval\_echo\_\$ILLOGICAL_IMPULSE_VIRTUAL_ENV)/bin/activate&&exec\_python\_-E\_"\$0"\_"\$@""
```

.config/ags/scripts/color_generation/generate_colors_material.py
clsty, line 1: Please modify the shebang to:
```
#!/usr/bin/env -S\_/bin/sh\_-xc\_"source\_\$(eval\_echo\_\$ILLOGICAL_IMPULSE_VIRTUAL_ENV)/bin/activate&&exec\_python\_-E\_"\$0"\_"\$@""
```

.config/ags/scripts/hyprland/get_keybinds.py
clsty, line 1: Please modify the shebang to:
```
#!/usr/bin/env -S\_/bin/sh\_-xc\_"source\_\$(eval\_echo\_\$ILLOGICAL_IMPULSE_VIRTUAL_ENV)/bin/activate&&exec\_python\_-E\_"\$0"\_"\$@""
```

.config/ags/scripts/wayland-idle-inhibitor.py
clsty, line 1: Please modify the shebang to:
```
#!/usr/bin/env -S\_/bin/sh\_-xc\_"source\_\$(eval\_echo\_\$ILLOGICAL_IMPULSE_VIRTUAL_ENV)/bin/activate&&exec\_python\_-E\_"\$0"\_"\$@""
```

arch-packages/illogical-impulse-python/PKGBUILD
clsty, line 1: Please bring back this file but modify its content to:
```
pkgname=illogical-impulse-python
pkgver=1.1
pkgrel=1
pkgdesc='Illogical Impulse Python Dependencies'
arch=(any)
license=(None)
depends=(
    gtk4
    libadwaita
    libsoup3
    libportal-gtk4
    blueprint-compiler
    gobject-introspection
    sassc
)
```

install.sh
```diff
@@ -116,7 +116,7 @@ install-local-pkgbuild() {
 }
 
 # Install core dependencies from the meta-packages
-metapkgs=(./arch-packages/illogical-impulse-{audio,backlight,basic,fonts-themes,gnome,gtk,portal,python,screencapture,widgets})
```
clsty, line 119: Please restore `python` in `metapkgs` here.

scriptdata/dependencies.conf
clsty, line 1: Please remove this part as we should use PKGBUILD for Arch packages (in `illogical-impulse-python`), not this file.

install.sh
```diff
@@ -130,18 +130,25 @@ for i in "${metapkgs[@]}"; do
...
+gradience_dir=/tmp/gradience
+x git clone https://github.com/ZeyadMoustafaKamal/Gradience.git $gradience_dir
+x cd $gradience_dir
+x git submodule update --init --recursive
+x uv pip install -r requirements.txt
+x meson setup build --prefix=$VIRTUAL_ENV
+x meson compile -C build
+x meson install -C build
+x cd -
+x deactivate # We don't need the virtual environment anymore
```
clsty, line 154: Please move this part into the function `install-uv` in [scriptdata/installers](https://github.com/end-4/dots-hyprland/pull/1060/files#diff-cb15db5013d2d04380e55e08f3714fc506a03841b7af7164855d5772f8300241), and use the following code here instead:
```sh
showfun install-uv
v install-uv
```

clsty (changes requested):

scriptdata/dependencies.conf
ZeyadMoustafaKamal: Ok so whey this file exists ? when I have installed the dots in the VM it installed stuff in `scriptdata/dependencies.conf` pretty fine so first why not use this file and second what is the benefit from using PKGBUILD ?

install.sh
ZeyadMoustafaKamal: but this part is about installing `gradience` not `uv`. I think that it should be part of `install.sh`.
```
x meson compile -C build
x meson install -C build
x cd -
x deactivate # We don't need the virtual environment anymore
```

clsty:
Oh, I forgot to tell you that I've done a quick test and it seems good but just need some modifications as mentioned above. After that it's very likely I'll merge this PR, if things goes as expected.

scriptdata/dependencies.conf
clsty: See https://end-4.github.io/dots-hyprland-wiki/en/dev/inst-script/

ZeyadMoustafaKamal: I see. Fixed now

install.sh
clsty: It should be part of `install.sh`, but this script generally allows people to choose from apply-or-skip for each step. So we should call it from a function and allow user to check this function and consider whether to skip it.

The name of function is not very important, if you think `install-uv` is not suitable, you may use `install-python-related`. Just ensure consistence.

clsty (approved):

ZeyadMoustafaKamal:
OK thanks for these words.

ZeyadMoustafaKamal:
Also I will do the documentation tomorrow. I was about to forget about it.

clsty:
> Also I will do the documentation tomorrow. I was about to forget about it.

Welcome. Take your time though, no hurry.

---

And for the fork for gradience, it seems that the final result is modification on its dependency list `requirements.txt`, as shown [here](https://github.com/GradienceTeam/Gradience/compare/main...ZeyadMoustafaKamal:Gradience:feat/setup.py). I'm not sure is this one necessary but anyway I've copied this repo to my account [here](https://github.com/clsty/ii-gradience), so that you're free to delete that fork now. @ZeyadMoustafaKamal 

And for @end-4 , since this is your project in the end, I recommend you to fork or copy the [repo](https://github.com/clsty/ii-gradience) to your account. (The original Gradience repo is archived and may probably be deleted one day.)

end-4:
Thanks
@clsty, I see you sent me a repo transfer request. It's invalid, so I forked it and it's now available at [end-4/ii-gradience](https://github.com/end-4/ii-gradience). I also sent a collaborator invite for that repo

ZeyadMoustafaKamal:
OK I still didn't create the `setup.py` because I took a a lot of time to figure out how to get `meson` to work so I decided that I will just make something that just works and then I will create `setup.py` in the future. And if I found that `meson` causes many problems I will edit the files that depends on `meson` so we don't need it anymore (we can do this actually). I will still need to learn about how to make a python package so just don't expect this in the near future.

PR #1062, opened 2025-01-27
Window opacity in hyprland/general.conf

Chiron8:
Added active_opacity, inactive_opacity and fullscreen_opacity with defaults as 1. Just wanted to make configuring opacity easier as it is a very popular configuration!

### DIFF
--- a/.config/hypr/hyprland/general.conf
+++ b/.config/hypr/hyprland/general.conf
@@ -90,6 +90,11 @@ decoration {
         render_power = 4
         color = rgba(0000002A)
     }
+
+    # Window Opacities
+    active_opacity = 1
+    inactive_opacity = 1
+    fullscreen_opacity = 1
     
     # Shader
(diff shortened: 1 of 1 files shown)


.config/hypr/hyprland/general.conf
clsty, line 1: Better as comment:
```
    # Window Opacities
    # active_opacity = 1
    # inactive_opacity = 1
    # fullscreen_opacity = 1
```

clsty (changes requested):

.config/hypr/hyprland/general.conf
Chiron8: Yeah makes sense.

nx-smul:
i have implemented this on with  transparency option in ags in my code https://github.com/nx-smul/dots-hyprland/blob/8c40c5ffd40ece8a1896e3dc4325220729a68258/.config/ags/modules/sideright/centermodules/configure.js#L172

PR #1068, opened 2025-01-30
italian translation

fabio-garavini:


clsty:
Thank you for contribution!

Issue #1080, opened 2025-02-06
[Issue] Switch to stable version for Hyprland packages


clsty:
- I have read the Usage and Troubleshooting pages of the wiki: https://end-4.github.io/dots-hyprland-wiki/en/i-i/04troubleshooting/
- I have made sure that both my config and system packages are up to date
- Linux distro: Arch Linux

---
## The issue

We'd better switch the packages to stable version from Arch Linux, not the `*-git` ones from AUR.

## Importance
This solves some compiling problems that some users (include myself) has encountered.

examples:
- I've encountered `libhyprutils.so.3: cannot open shared object file: No such file or directory` issue when launching Hyprland (from tty). And I can not resolve it after spending hours debugging, such as running commands from [Hyprland#9202 comment](https://github.com/hyprwm/Hyprland/issues/9202#issuecomment-2621759143).
- I've managed to launch Hyprland anyway by forcely copy `libhyprutils.so.4` to `libhyprutils.so.3` but xwayland stops working after that.
- Some of the issues of this project also complain about `hypr*-git` installation problem, e.g. #1041

## Benefits
- Solves problems above.
- Much faster installation speed. (It saves above 1 hour for my laptop.)
- Much stabler installation result and running environment.
- Simpler install scripts.
- Less dependencies from AUR. (#1017)

## Solution 
@end-4 If you agree with me, I'll replace all the hyprland related `*-git` packages to the stable Arch packages, i.e.
```
hyprutils hyprlang hypridle hyprland-qt-support hyprland-qtutils hyprlock xdg-desktop-portal-hyprland hyprcursor hyprwayland-scanner hyprland
```
which you can install with `sudo pacman -S `.
See PR #1081.
P.S. I've tested this and it works well with the current dots-hyprland configurations.

end-4:
Considering I don't actively use my dotfiles and tweak it to use new features, this shouldn't hurt. So sure

nx-smul:
i have added caotic aur to not compile it from aur

clsty:
@nx-smul idk why but some users here also use caotic-aur and still have the problem... https://github.com/hyprwm/Hyprland/issues/9202

daUnknownCoder:
> Much faster installation speed. (It saves above 1 hour for my laptop.)

major reason because the new merge saved me 4 hours of installation

PR #1081, opened 2025-02-06
Switch to stable hyprland related packages (#1080)

clsty:
See #1080 for details.


end-4:
Since I'm fine with this change, please merge once you think this PR is ready

Issue #1070, opened 2025-01-31
Any option to shrink calendar in quick settings


angelomds42:
**What would you like to be added?**
We should have an option to disable or at least shrink the calendar from the quick settings panel 

**How will it help**
It would be useful for small monitors, in my case when I use the notebook's own screen the calendar takes up more than half of the quick panel, leaving almost no space for notifications, Wi-Fi, etc.

**Screenshot**
![Image](https://github.com/user-attachments/assets/41fc1a79-c865-4a41-8a5a-ac2653abc6d6)

Pharmaracist:
make reavealer with button in sideright.js file then add the button wherever u want

angelomds42:
> make reavealer with button in sideright.js file then add the button wherever u want

I created a button to hide the calendar, but if I use that button twice, the calendar stops showing the days, months, etc.

like:
1- Expand the calendar for the first time (it works correctly).
2- Collapse the calendar.
3- Expand the calendar for the second time (days and months are not displayed).

Pharmaracist:
Try making the revealer functionality inside the side right file after every module is fully loaded

Pharmaracist:
Try making the revealer functionality inside the side right file after every module is fully loaded and use revealers since they are already there and give better animations

Pharmaracist:
if you cannot figure it out yet check its implementation in https://github.com/pharmaracist/hyprnoon 

https://github.com/user-attachments/assets/36e3f6c3-c7d2-4169-8100-41f1cf225a3c

end-4:
done

https://github.com/user-attachments/assets/ba284ba7-21f5-4722-8e47-92c0ef59ab19

end-4:
now you can set `sidebar.calendar.expandByDefault` in user options to make it collapsed on launch

Issue #1029, opened 2024-12-29
[Feature] Get Module Firewall in sideright.js


haoireal:
I am trying to write a module about firewalld toggle in quicktoggles.js
when I import it in sideright.js it makes ags error. I need help. 

```javascript
// cCustom moduel for firewalld
export const ModuleFirewall = (props = {}) =>
  Widget.Button({
    className: "txt-small sidebar-iconbutton",
    tooltipText: "FirewallD | Right-click to configure",
    onClicked: () => {
      execAsync("systemctl is-active firewalld")
        .then((status) => {
          if (status.trim() === "active") {
            exec("sudo systemctl stop firewalld").catch(print);
          } else {
            exec("sudo systemctl start firewalld").catch(print);
          }
        })
        .catch((error) => {
          print(`Error checking firewalld status: ${error}`);
        });
    },
    onSecondaryClickRelease: () => {
      execAsync(["bash", "-c", "firewalld-config-tool"]).catch(print);
      closeEverything();
    },
    child: FirewallIndicator(),
    setup: (self) => {
      setupCursorHover(self);

      // Cập nhật trạng thái của firewalld
      const updateFirewallStatus = () => {
        execAsync("systemctl is-active firewalld")
          .then((status) => {
            const isActive = status.trim() === "active";
            self.toggleClassName("sidebar-button-active", isActive);
            self.tooltipText = isActive
              ? "FirewallD Active"
              : "FirewallD Inactive";
          })
          .catch((error) => {
            print(`Error updating firewalld status: ${error}`);
          });
      };

      updateFirewallStatus();
      setInterval(updateFirewallStatus, 5000);
    },
    ...props,
  });

haoireal:
I just tweaked the source code a bit and it works. But only when the firewall is turned off. If you want to turn it on, you can't

```javascript
// cCustom moduel for firewalld
export const ModuleFirewall = (props = {}) =>
  Widget.Button({
    className: "txt-small sidebar-iconbutton",
    tooltipText: "Firewalld",
    onClicked: (self) => {
      execAsync("systemctl is-active firewalld")
        .then((status) => {
          const isActive = status.trim() === "active";
          const command = isActive
            ? "pkexec systemctl stop firewalld"
            : "sudo systemctl start firewalld";
          execAsync(command)
            .then(() => {
              self.toggleClassName("sidebar-button-active", !isActive);
            })
            .catch(print);
        })
        .catch(print);
    },
[...]

Issue #1003, opened 2024-12-15
[Feature] Disk usage widget


h0pes:
**What would you like to be added?**
would be nice to have a simple hard disk usage widget on the left of the ram usage one. Just one like `df -h | awk 'NR==2 {printf "%s\n", $4}'` or in % format

**How will it help**
keep control of disk space

**Extra info**
![image](https://github.com/user-attachments/assets/ee4df6fc-46ff-44a9-81c7-6a6079d2d277)
Apply same design and theme colors as the current widgets, nothing fancy

alr86:
The remaining storage is something no one always wants to know or changes in the short term, in my opinion, this feature is not very useful for the general users

enessmr:
I did the same.
But instead of folder i did hard_disk cause it feels correct

end-4:
If I make this I will have it on the sidebar. Disk usage is not something we actively want to know like RAM usage

enessmr:
> Disk usage is not something we actively want to know like RAM usage

I added disk usage to my dotfiles, for android deployment and some large things that cost more than 1gb of disk usage so yeah

clsty:
Closing cuz illogical-impulse on AGS has been deprecated. The latest version is illogical-impulse on Quickshell.

Issue #683, opened 2024-07-08
[Feature] Add an option to toggle from battery to weather


gnussbaum67:
Was wondering if it's possible to toggle the battery to weather in ags?
**What would you like to be added?**

ability to toggle weather and battery module in ags
**How will it help**

**Extra info**

andresparrab:
That would be a nice addition. i always looking for the weather. but cant find in in the panel on my laptop, i think only available in pc mode?

gnussbaum67:
> That would be a nice addition. i always looking for the weather. but cant find in in the panel on my laptop, i think only available in pc mode?

I get around this by changing line 211 in ~/.config/ags/modules/bar/normal/system.js and changing it to "laptop" from "desktop"

fruitsaladchan:
@gnussbaum67 you can get battery and weather like this 

![swappy-20240831-153308](https://github.com/user-attachments/assets/4bd6e198-bb2a-4021-9ee8-d3b9de8f5af5)

by replacing the battery module in ~/.config/ags/modules/bar/normal/system.js with this
https://gist.github.com/fruitsaladchan/a227e71098ef5914795061c73ed33866

PR #1094, opened 2025-02-08
[ Module ] Hyprland gamemode

nx-smul:
this module works exactly like this one https://wiki.hyprland.org/Configuring/Uncommon-tips--tricks/#toggle-animationsbluretc-hotkey 

https://github.com/user-attachments/assets/71e5c38c-7d20-4a59-960c-5b29e1514b8c

### DIFF
--- a/.config/ags/modules/sideright/quicktoggles.js
+++ b/.config/ags/modules/sideright/quicktoggles.js
@@ -203,6 +203,34 @@ export const ModuleRawInput = async (props = {}) => {
     };
 }
 
+export const ModuleGameMode = async (props = {}) => {
+    try {
+        const Hyprland = (await import('resource:///com/github/Aylur/ags/service/hyprland.js')).default;
+        return Widget.Button({
+            className: 'txt-small sidebar-iconbutton',
+            tooltipText: getString('Hyprland Game Mode'),
+            onClicked: (button) => {
+                Hyprland.messageAsync('j/getoption animations:enabled')
+                    .then((output) => {
+                        const value = JSON.parse(output)["int"];
+                        if (value == 1) {
+                            execAsync(['bash', '-c', `hyprctl --batch "keyword animations:enabled 0; keyword decoration:shadow:enabled 0; keyword decoration:blur:enabled 0; keyword general:gaps_in 0; keyword general:gaps_out 0; keyword general:border_size 1; keyword decoration:rounding 0; keyword general:allow_tearing 0" & hyprctl reload`]).catch(print);
+                            button.toggleClassName('sidebar-button-active', false);
+                        } else {
+                            execAsync(['bash', '-c', `hyprctl --batch "keyword animations:enabled 1; keyword general:allow_tearing 1" & hyprctl reload`]).catch(print);
+                            button.toggleClassName('sidebar-button-active', true);
+                        }
+                    })
+            },
+            child: MaterialIcon('gamepad', 'norm'),
+            setup: setupCursorHover,
+            ...props,
+        })
+    } catch {
+        return null;
+    };
+}
+
 export const ModuleIdleInhibitor = (props = {}) => Widget.Button({ // TODO: Make this work
     attribute: {
         enabled: false,
--- a/.config/ags/modules/sideright/quicktoggles.js
+++ b/.config/ags/modules/sideright/quicktoggles.js
@@ -214,10 +214,10 @@ export const ModuleGameMode = async (props = {}) => {
                     .then((output) => {
                         const value = JSON.parse(output)["int"];
                         if (value == 1) {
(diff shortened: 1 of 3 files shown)


clsty:
While this module is good for my use case, it occupies a button and some users may not like it since the space there is already limited (there're 6 buttons already).

If we can fully implement the modularization, i.e. provide an easy way for users to customize the buttons and widgets on the AGS bars, I'll be very happy to accept this PR.

However, since end-4 is inactive for functional updates, and I do not master JS nor AGS, I can only count on unknown contributors sending PR for such modularization in the future. This will be a huge step and has been expected early from #271 but still not completed till now. If someone could achieve that (maybe you?), I would definitely put one's name on the [credits page](https://end-4.github.io/dots-hyprland-wiki/en/general/credits/).

clsty:
Also I'm personally curious about the widget/dock/bar(?) containing games popping out at the bottom in the video. Could you please tell me what is that?

nx-smul:
> While this module is good for my use case, it occupies a button and some users (including end-4) may not like it since the space there is already limited (there're 6 buttons already).
> 
> If we can fully implement the modularization, i.e. **providing a relative easy way for users to customize the buttons and widgets on the AGS bars**, I'll be very happy to accept this PR.
> 
> However, since end-4 is inactive for functional updates, and I do not master JS nor AGS, I can only count on unknown contributors sending PR for such modularization in the future. This will be a huge step and has been expected early from #271 but still not completed till now. If someone could achieve that (maybe you?), I would definitely put one's name on the [credits page](https://end-4.github.io/dots-hyprland-wiki/en/general/credits/). It will make lots of wishes come true (or become possible) too, such as #1070, #683, #675, #1003, #1029, etc.

can't implement this on bar. i am not much of a pro to reverse engineer the full code. maybe can add it on the development setting.

nx-smul:
> Also I'm personally curious about the widget/dock/bar(?) containing games popping out at the bottom in the video. Could you please tell me what is that?

it is rofi actually. found this config for sway in a repo. converted it to be usable in hyprland

end-4:
> While this module is good for my use case, it occupies a button and some users (including end-4) may not like it since the space there is already limited (there're 6 buttons already).
> 
> If we can fully implement the modularization, i.e. providing a relative easy way for users to customize the buttons and widgets on the AGS bars, I'll be very happy to accept this PR.

[the button for WARP](https://end-4.github.io/dots-hyprland-wiki/en/i-i/03config/#cloudflare-warp) only shows if you have it installed. i guess game mode button should only show if you have steam (or other launchers, feel free to suggest guys) installed
customization of the buttons i think shouldn't be hard to implement. we already have something like that for the left sidebar. i can do it

.config/ags/modules/sideright/quicktoggles.js
```diff
@@ -203,6 +203,34 @@ export const ModuleRawInput = async (props = {}) => {
...
+            onClicked: (button) => {
+                Hyprland.messageAsync('j/getoption animations:enabled')
+                    .then((output) => {
+                        const value = JSON.parse(output)["int"];
+                        if (value == 1) {
+                            execAsync(['bash', '-c', `hyprctl --batch "keyword animations:enabled 0; keyword decoration:shadow:enabled 0; keyword decoration:blur:enabled 0; keyword general:gaps_in 0; keyword general:gaps_out 0; keyword general:border_size 1; keyword decoration:rounding 0; keyword general:allow_tearing 0" & hyprctl reload`]).catch(print);
+                            button.toggleClassName('sidebar-button-active', false);
+                        } else {
+                            execAsync(['bash', '-c', `hyprctl --batch "keyword animations:enabled 1; keyword general:allow_tearing 1" & hyprctl reload`]).catch(print);
+                            button.toggleClassName('sidebar-button-active', true);
```
end-4: why do we disallow tearing when enabling game mode? doesn't tearing reduce latency?

nx-smul: i don't know why but the toggle option works revers have to fix it

nx-smul: also tearingg option is a mistake

end-4:
> `make game mode toggle work more consistently`

i tried doing something to make it more consistent but after a while i find it not that consistent... @nx-smul can you explain why you had the `& hyprctl reload` in the first place? i thought its kinda weird because reloading the config would just make hyprland enable animations n' stuff again so i removed it...

end-4:
i'm gonna note that i made invert colors button no longer show by default. it was there because i thought it's cool, but it's not very practical

end-4:
okay i gave the script on the wiki page another read. i think you missed the part that the script exits if it already toggles game mode. my commit fixed it
i think the inconsistency is caused by how ags, because the script on the wiki page just works normally

end-4:
should be alright... i'll merge this now, thanks

nx-smul:
> should be alright... i'll merge this now, thanks

thank you for fixing the code

enessmr:
Huh, I did a similar setup to nx-smul.

I uninstalled steam months ago... I stopped gaming now

end-4:
Today I learned I cannot block a person from commenting on my repository if it isn't in an organization. I am not creaing an organization and will cry more.

PR #1114, opened 2025-02-12
Appearence Enhancements 

Pharmaracist:


Pharmaracist:
@clsty

clsty:
I've noticed this PR and I do think that it's cool, but @end-4 is the one making decision on AGS related things.

Pharmaracist:
in minor tweaks i removed the comments

Pharmaracist:
image
this is the gowall stuff 
still need abit of work but its working and very fast

Pharmaracist:
when i added rofi support i added full rofi setup for just fall back like bluetooth , bookmarks storage area file browser and pdf container its just the keybinds i left it for you to do it since my keybinds are completely different + i need help with foot i really dont know it but it seems to be the one having hot relaod on colorchange and mode change so thats important

Issue #1050, opened 2025-01-16
[Question] how to correctly replace foot with kitty


Yatagarashy:
how to correctly replace foot with kitty, otherwise I changed it and colorschema doesn’t work for me on it

![Image](https://github.com/user-attachments/assets/6b66aa14-af1e-441e-bf72-6a386c78477e)

nizaralheet:
It is just about the .config/ags/scripts/templates/terminal/sequences.txt file , I am not sure how to do it but i think it could be edited some who to be better working on kitty

PR #1123, opened 2025-02-17
Update sequences.txt to make it work on most fo termials 

nizaralheet:
just made the theme work on alacritty +kitty and some other terminals , 
I was having the same issue as https://github.com/end-4/dots-hyprland/issues/1050 so turned out it is just have to add another sequence to it and now it is more compatible with other terminal


Pharmaracist:
i really appreciate the work done for this but isnt switching to matugen really solve the problem

nizaralheet:
I really don't know / didn't try to do that , like i just saw how the colors apply to the terminals using fish + for me it just works why adding more complexity to simple thing  just KISS

dannyusername:
How does this work in alacritty? I only know how to use toml color schemes with it. Not sure how to get it to use ANSI escape sequences

nizaralheet:
a fish function run this 
`cat ~/.cache/ags/user/generated/terminal/sequences.txt`
which configure terminal colors , if want to try copy the template file to you config apply a wallpaper then open alacritty run the command above

end-4:
this works. im too lazy to review the matugen pr for now so i'm merging this
thanks

g1sbi:
> a fish function run this `cat ~/.cache/ags/user/generated/terminal/sequences.txt` which configure terminal colors , if want to try copy the template file to you config apply a wallpaper then open alacritty run the command above

Used this same method in ghostty and I can confirm it works.

However I had to suppress my `powerlevel10k` warning as running the command outputs an empty string.

PR #1131, opened 2025-02-23
ags: fix parsing `ddcutil detect --brief`

moetayuko:
Recent update introduced a new `drm_connector_id` line, so it's no longer feasible to hardcode the line numbers.


clsty:
@end-4 Please take a look at this PR if possible.

end-4 (commented):
so it just checks for the vorrect line instead of assuming line 1 and line 3?
looks good to me

end-4:
thanks

PR #1133, opened 2025-02-24
fix: correct desktop icon name

cristoxdxd:
Correct icon `desktop_symbolic` for change screen resolution script


clsty:
idk but the name seems should be `desktop-symbolic` cuz the icon has a file [here](https://github.com/end-4/dots-hyprland/blob/main/.config%2Fags%2Fassets%2Ficons%2Fdesktop-symbolic.svg).

cristoxdxd:
It's strange, I did a fresh installation, and the SVG has an underscore instead of the hyphen in the name.

PR #953, opened 2024-11-19
[FEATURE] Adjust volume via music widget

endotrizine:
Scroll up/down the music widget now changes music volume via playerctl instead of switching workspaces

PR #1160, opened 2025-03-11
Fix: Adjust volume only for supported players

endotrizine:
#953 small fix
This improves the `adjust-volume.sh` script by making it detect and adjust the volume of the first available media player that supports `playerctl volume`. Without this fix, volume control may not work correctly if a browser or another non-controllable application is running.

### DIFF
--- a/.config/ags/scripts/music/adjust-volume.sh
+++ b/.config/ags/scripts/music/adjust-volume.sh
@@ -2,14 +2,23 @@
 
 change=$1
 
-current_volume=$(playerctl volume)
+players=$(playerctl -l)
 
-new_volume=$(echo "$current_volume + $change" | bc)
+for player in $players; do
+    if playerctl -p "$player" volume &>/dev/null; then
+        current_volume=$(playerctl -p "$player" volume)
 
-if (( $(echo "$new_volume > 1.0" | bc -l) )); then
-  new_volume=1.0
-elif (( $(echo "$new_volume < 0.0" | bc -l) )); then
-  new_volume=0.0
-fi
+        new_volume=$(echo "$current_volume + $change" | bc)
+        
+        if (( $(echo "$new_volume > 1.0" | bc -l) )); then
+            new_volume=1.0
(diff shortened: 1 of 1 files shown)


enessmr:
Then make a M3 Media player so we could acsess (not ags its literally an app, widget i dont really know)

enessmr:
Or make a widget that hyprland recognizes it as a window so you can remake mpv

Oh wait, GJS with AGS JS with GLib and Gio and Gtk is still limited.

You cant remake mpv or vlc with this js.

endotrizine:
@enessmr 
> Then make a M3 Media player so we could acsess (not ags its literally an app, widget i dont really know)



> Or make a widget that hyprland recognizes it as a window so you can remake mpv
> 
> Oh wait, GJS with AGS JS with GLib and Gio and Gtk is still limited.
> 
> You cant remake mpv or vlc with this js.

I'm not sure if my English is the issue here, but I genuinely don't understand what you're trying to say. What do you even mean?

This is just a fix for my previous PR, where I implemented a small feature for adjusting music volume.

Previously, it wouldn't work if a non-controllable application was running. This fix ensures that the script first checks if playerctl can control the player's volume, and if not, it looks for another supported player.

clsty:
@endotrizine You don't have to reply him as I've already folded the comments as offtopic.

(I remembered that end-4 wanted to ban this user but we can't because only organization have that function. You know what I meant, right?)

---

@end-4 I'd like to merge this PR but you'd better take a look at it. Thx in advance.

end-4:
i realized i can ban a user by blocking them in account settings, not in repo settings. we should be good now

---

ill look at this later today

end-4:
I just tried, somehow it doesn't work for me. This is what I did:
- Open vlc (properly supported) -> this works
- After that open Chrome (not properly supported) -> "chromium.instanceXXXX" comes before "vlc" in lexicographical order and gets picked first -> script tries to adjust chrome's volume, fails but still exits

I'm not sure if Chrome or playerctl or this script is to blame

end-4:
i'm closing this now, thanks anyway

PR #1179, opened 2025-03-19
[BUG] fixed the system update on line 59 

japlic:
its sudo pacman -Syu not  v sudo pacman -Syu


clsty:
https://github.com/end-4/dots-hyprland/blob/b3c554a061a7d876ff2b7d077ad467a657a45b4d/scriptdata/functions#L9

Have you ever tested the script? If it did not gives an error when running that `v` command, you should not consider it as a bug in the first place.

PR #1183, opened 2025-03-24
Pull Request: Fix Window Rules Syntax for v2

spaceofnova:
This PR updates the configuration to use the new v2 syntax exclusively. The old v1 syntax was causing invalid rule errors. All window rules now use windowrulev2 with explicit fields (e.g. class:, title:, floating:) as per the latest Hyprland documentation.

This change fixes the syntax issue and ensures the configuration is compatible with the current Hyprland version.

(I will note, this change was partially aided with ChatGPT reasoning; But this has been tested to be working very well.)

### DIFF
--- a/.config/hypr/hyprland/rules.conf
+++ b/.config/hypr/hyprland/rules.conf
@@ -1,99 +1,51 @@
 # ######## Window rules ########
-windowrule = noblur,.*
-# windowrule = opacity 0.89 override 0.89 override, .* # Applies transparency to EVERY WINDOW
-windowrule = float, ^(blueberry.py)$
-windowrule = float, ^(steam)$
-windowrule = float, ^(guifetch)$ # FlafyDev/guifetch
-windowrulev2 = tile, class:(dev.warp.Warp)
+
+# General rule – disable blur for all windows.
+# (Using "class:.*" as a catch-all field since at least one field is required)
+windowrulev2 = noblur, class:.*
+
+# Uncomment to apply global transparency to all windows:
+# windowrulev2 = opacity 0.89 override 0.89 override, class:.*
+
+# Specific floating windows.
+windowrulev2 = float, class:^(blueberry\.py)$
+windowrulev2 = float, class:^(steam)$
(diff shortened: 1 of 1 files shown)


clsty:
@end-4 Please review this PR when possible. Thx in advance.

.config/hypr/hyprland/rules.conf
```diff
@@ -1,99 +1,51 @@
...
+windowrulev2 = float, title:^(Select a File)(.*)$
+windowrulev2 = float, title:^(Choose wallpaper)(.*)$
+windowrulev2 = float, title:^(Open Folder)(.*)$
+windowrulev2 = float, title:^(Save As)(.*)$
+windowrulev2 = float, title:^(Library)(.*)$
+windowrulev2 = float, title:^(File Upload)(.*)$
+
+# --- Tearing ---
+windowrulev2 = immediate, title:.*\.exe
+windowrulev2 = immediate, class:^(steam_app)$
```
end-4: it comes with an id, like `steam_app_7824873` so we don't want to match word end here

.config/hypr/hyprland/rules.conf
```diff
@@ -1,99 +1,51 @@
...
+windowrulev2 = pin, title:^([Pp]icture[-\s]?[Ii]n[-\s]?[Pp]icture)(.*)$
+
+# Additional dialog rules to force floating.
+windowrulev2 = float, title:^(Open File)(.*)$
+windowrulev2 = float, title:^(Select a File)(.*)$
+windowrulev2 = float, title:^(Choose wallpaper)(.*)$
+windowrulev2 = float, title:^(Open Folder)(.*)$
+windowrulev2 = float, title:^(Save As)(.*)$
+windowrulev2 = float, title:^(Library)(.*)$
+windowrulev2 = float, title:^(File Upload)(.*)$
```
end-4: while we're at it i think it's a good idea to have these right after the center rules

end-4 (commented):
looks good
note that later on we'll have to change windowrulev2 to windowrule when it gets removed but this should work for now

end-4:
thanks

PR #1218, opened 2025-04-09
Update bluetooth.js

inoribea:
When I use it, the following error will occur

``` 
(com.github.Aylur.ags:98720): Gjs-CRITICAL **: 00:06:00.589: JS ERROR: TypeError: self.enabled is undefined
BluetoothDevice/extraSetup/</<@file:///home/inoribea/.config/ags/modules/sideright/centermodules/bluetooth.js:60:38
timeout/<@resource:///com/github/Aylur/ags/utils/timeout.js:14:9
_init/GLib.MainLoop.prototype.runAsync/</<@resource:///org/gnome/gjs/modules/core/overrides/GLib.js:263:34
```

The error may be caused by the undefined enabled property, which may lead to inconsistent object states when setting the property value with a delay using Utils.timeout.

### DIFF
--- a/.config/ags/modules/sideright/centermodules/bluetooth.js
+++ b/.config/ags/modules/sideright/centermodules/bluetooth.js
@@ -57,7 +57,7 @@ const BluetoothDevice = (device) => {
             device.setConnection(newValue);
         },
         extraSetup: (self) => self.hook(device, (self) => {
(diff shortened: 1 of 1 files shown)


end-4:
i guess in theory this is better than having an arbitrary value there
but it's self.enabled.value not self.value

end-4:
no longer valid after the transition from ags to quickshell

PR #1222, opened 2025-04-12
AGS v2 implementation of the Bar module

topazrn:
Things that I need your opinion on:

- Prettier and eslint configuration: I tried my best to follow the AGS v1 codebase's style.
- Is converting js to ts out of scope? Should I put `"allowJs": true` in tsconfig.json?
- I changed the `.config/modules/.*` folders to `.config/modules/core/*` because I can't get TypeScript to work correctly on files in folders starting with a dot.

Please do nitpick my work as I want to deliver quality as well.


end-4:
- Formatting: I don't like the closing brace of an if being on the same line as an else
- Converting js to ts: I think we should. From what I see in the changes so far you still have js code in there, and the main goal is to migrate to agsv2 and not exactly js to ts, so we should allow js if it makes things easier. 
- About moving the folders starting with dots, yeah that's fine, idk how I came up with that shitty convention in the first place lol

Comment about JSX: AGS recommends it, but what's the point? I can only see the disadvantage of losing bracket matching/jumping... I find it bothersome to have two tags for one thing instead of just a pair of brackets

It'll probably take some time for me to review this, because:
- ughhhh exams
- I've been also trying to work on a [quickshell](https://quickshell.outfoxxed.me/) version of my dotfiles (stored in ii-qs branch of this repo) in the past few days and would really like to put my effort into it. I should've announced this in the agsv2 migration issue sooner... This PR isn't entirely useless, but it isn't really my priority at the moment

PR #1237, opened 2025-04-24
Fixes the function getImageViewerApp

TheGreyfeather:
The async function getImageViewerApp return void as it doesn't return anything from the function instead only returning from within the promise.
This fix allows the "Waifu" tab to open image in an external viewer.

I have tested this locally and have verified that it now functions as expected.


end-4:
thanks

PR #1245, opened 2025-04-26
fix: add -A to makepkg

saeziae:
i was trying to install it on archlinuxarm

this patch enables the installation via `install.sh` script on non-x86 devices

i confirmed that all packages can be built under aarch64 without modification


clsty:
LGTM, thanks.

PR #1252, opened 2025-04-28
fix(bluetooth): sync toggle state with device connection

MarcoMadera:
Problem:
When using the Bluetooth toggle in the sidebar, the label (Connected/Paired) updated correctly when the device connection state changed, but the switch itself did not reflect the real status.

How I tested:
I manually connected and disconnected Bluetooth devices using `bluetoothctl connect <ID>`, `bluetoothctl disconnect <ID>`commands, and observed that while the label changed instantly, the toggle did not update unless manually reloading the UI or rebuilding the device list

Fix:
The fix reuses the same enabled state (stored inside self.attribute.enabled) that the label and other widget parts are already observing


end-4:
cleaner than doing an arbitrary timeout for sure, thanks

Issue #1241, opened 2025-04-25
Sync colors to Emacs


YuvanMichaelVivenzi:
### What would you like to be added?

Just like how the wal colors are made and synced to ags. I want them for Emacs too. + Other apps that won't require much effort. 

### How will it help?

better looks and coherence. + Emacs is a popular text editor. In fact, more so. 

### Extra info

Ewal repo. original wal repo with .el files and many more.

clsty:
https://github.com/end-4/dots-hyprland/blob/main/.config%2Fags%2Fscripts%2Fcolor_generation%2Fapplycolor.sh

PR is welcomed if you are able to write the code to apply color scheme for Emacs. Just a little requirement: it should be optional for Emacs users. For example generating an `illogical-impulse.el` for users to include manually in their `init.el`. Certainly not directly overwrite `init.el`.

clsty:
I think the file you need is located under `$HOME/.local/state/ags/scss`.

Since deekwiki is out and this repo is indexed, I asked the AI a question and this may help you.

https://deepwiki.com/search/how-can-i-add-a-applyemacs-fun_829c84e0-b376-4e37-a463-904e1f04715d

end-4:
~~holy shit no more docs writing????~~

YuvanMichaelVivenzi:
yep, found the ```material_colors.scss``` in ```$HOME/.cache/ags/user/generated```


P.S Thx for the link! It really helps!

YuvanMichaelVivenzi:
> ~holy shit no more docs writing????~

It's still better to make your own docs tho. 

Only you would know the real stuff in this. Instead of the AI. And I believe that human-written words are more understandable than AI made flowcharts.

end-4:
yeah i was just kidding of course

YuvanMichaelVivenzi:
@end-4 @clsty , Please give me some clarity on this ... 
I've been able to successfully make a GOOD theme!
But the main part is. It needs a .css file for colors.
I used ```matugen``` for this. 
Since the colors are .scss for your dots, I was wondering if matugen generated  .css would be fine. ?

YuvanMichaelVivenzi:
@end-4 @clsty 
I've also made a discussion on matugen repo at https://github.com/InioX/matugen/discussions/151 so that other matugen users can benefit. And maybe InioX will add this to his templates list.

YuvanMichaelVivenzi:
@end-4 @clsty 

Any final thoughts?

[...]

PR #1255, opened 2025-04-30
Create material-theme.el >> New emacs theme in "Extras/emacs"

YuvanMichaelVivenzi:
Material colors based emacs theme that syncs with wallpaper!

As discussed in #1241 I've made a PR.


end-4:
thank you

PR #1266, opened 2025-05-03
Enables the Color Picker in the left Sidebar

TheGreyfeather:
Enables the color picker in the left sidebar by setting clicked: true
Cleans up unused imports in the colorpicker.js and color.js file


end-4:
`clicked` there is just the initial clicked state, changing it to true does nothing

Issue #1261, opened 2025-05-01
Issue with install script


ky0und:
### Step 1. Before you submit

- [x] I have read the [Troubleshooting](https://end-4.github.io/dots-hyprland-wiki/en/i-i/04troubleshooting/) and [Usage](https://end-4.github.io/dots-hyprland-wiki/en/i-i/02usage/) pages.
- [x] I've successfully updated to the latest version following the [guidance](https://end-4.github.io/dots-hyprland-wiki/en/i-i/01setup/#updating).
- [x] I've successfully updated the system packages to the latest.
- [ ] I've ticked the checkboxes without reading their contents

### Step 2. Quick diagnose info

<details><summary>Quick diagnose</summary>

```
/========================\
| Checking git repo info |
\========================/
 
[===diagnose===] git remote get-url origin
https://github.com/end-4/dots-hyprland.git
[---SUCCESS---]
 
[===diagnose===] git rev-parse HEAD
d72b198ba16b2ba9278cc9b8edecb27551fe39be
[---SUCCESS---]
 
/=================\
| Checking distro |
\=================/
 
[===diagnose===] ii_check_distro
./diagnose: line 34: lsb_release: command not found
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo
[---SUCCESS---]
 
/====================\
| Checking variables |
\====================/
 
[===diagnose===] declare -p XDG_BIN_HOME
./diagnose: line 30: declare: XDG_BIN_HOME: not found
[---EXIT 1---]
 
[===diagnose===] declare -p XDG_CACHE_HOME
./diagnose: line 30: declare: XDG_CACHE_HOME: not found
[---EXIT 1---]
 
[===diagnose===] declare -p XDG_CONFIG_HOME
./diagnose: line 30: declare: XDG_CONFIG_HOME: not found
[---EXIT 1---]
 
[===diagnose===] declare -p XDG_DATA_HOME
[...]

PR #1273, opened 2025-05-04
Replace Gradience with a simpler solution

end-4:
As discussed in https://github.com/end-4/dots-hyprland/issues/1261#issuecomment-2849248001
Reduces the size of the python mess we're having

It seems to be already working with minor tweaks, but I'll clean up stuff in the scripts and test on a new account to make sure it works
@clsty you're able to make changes to this branch right?

### DIFF
--- a/.config/ags/i18n/locales/Default.json
+++ b/.config/ags/i18n/locales/Default.json
@@ -145,7 +145,6 @@
     "Options": "Options",
     "Dark Mode": "Dark Mode",
     "Ya should go to sleep!": "Ya should go to sleep!",
-    "Use Gradience": "Use Gradience",
     "Theme GTK apps using accent color\n(drawback: dark/light mode switching requires restart)": "Theme GTK apps using accent color\n(drawback: dark/light mode switching requires restart)",
     "Scheme styles": "Scheme styles",
     "Vibrant": "Vibrant",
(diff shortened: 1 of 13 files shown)


clsty:
Yep I'm able to modify it.

clsty:
Do we need to install `matugen` or something else? However I did not find any new executable involved in this PR...

end-4:
> Do we need to install `matugen` or something else? However I did not find any new executable involved in this PR...

Matugen is not needed. We we're just using a simple template from Matugen's theme repo and applying colors with our current scripts

end-4:
I'm ready when you are @clsty 
since you commited `Also remove rsync glib-2.0`, should we remove `x ls $XDG_DATA_HOME/glib-2.0/schemas` from the diagnose script?

end-4:
I tested on a new user. Installation runs fine and color is applied properly.

clsty:
> I'm ready when you are @clsty 
> since you commited `Also remove rsync glib-2.0`, should we remove `x ls $XDG_DATA_HOME/glib-2.0/schemas` from the diagnose script?

Yes. we should.

clsty:
> I tested on a new user. Installation runs fine and color is applied properly.
> i can archive `ii-blueprint-compiler` and `ii-gradience` after merging this right?

Yes, and also that `ii-pillow`.

clsty:
I've done testing and it works well. Let's merge it.

PR #1288, opened 2025-05-09
Fix title on workspace change

rllynotfox:
If there aren't any opened windows, title widget displays "workspace 2", but if you switch between opened workspaces without opening any windows, this title doesn't change

In this fix, I added a hook to update this widget also when workspace number changes

(I didn't add hook on top title, because it won't change in this case anyway)


end-4:
thanks

PR #1292, opened 2025-05-13
Feat: switch to video background and colorgen

obsidrielle:
## Main Feature
The functionality to use videos as wallpapers and generate corresponding theme colors using the first frame of the video (captured using ffmpeg).

It seems mpvpaper cannot retain settings, so when switching to video, a startup script is automatically generated that runs with `exec-once`, and when switching to an image, they are deleted.

## Some Issues
- Memory usage: When using mpvpaper, hardware acceleration is enabled by default, but there is still significant performance consumption, especially when using too many displays. Currently, the memory usage on one display is about 450~500MB.

Please let me know if you have any questions or suggestions.


end-4:
This is very cool. I really like how cleanly this is handled. 
I looked through the code and found no problem, and from my testing this works well. But since this is a bash script, @clsty, it would be nice to have your opinion

.config/ags/scripts/color_generation/switchwall.sh
```diff
@@ -3,6 +3,83 @@
...
+
+remove_restore() {
+	if [ -f "$RESTORE_SCRIPT" ]; then
+		rm "$RESTORE_SCRIPT"
+	fi
+
+	# Delete `exec-once = bach $RESTORE_SCRIPT` if exists
+	if [ -f "$EXECS_CONF" ]; then
+		grep -v "exec-once=bash $RESTORE_SCRIPT" "$EXECS_CONF" > "$EXECS_CONF.tmp"
+		mv "$EXECS_CONF.tmp" "$EXECS_CONF"
```
clsty: It's bad idea to edit this file on the fly from my point of view. You'd better edit a separate file like  `mpvwallpaperscript.conf` and include it inside the `hyprland.conf`. So when you overwrite this separate file, even mistake happens, such as a full disk usage, the file of the user remain safe.

clsty:
Good work. Mostly LGTM with only one problem.

clsty (changes requested):

.config/ags/scripts/color_generation/switchwall.sh
mon3stera: How about defining a empty startup script and adding `exec-once = ~/.config/hypr/custom/scripts/__restore_video_wallpaper.sh` to `hyprland/execs.conf`in advance ? 

This way, when switching to an image as wallpaper, it just empties this script without deleting it. My intention is to avoid modifying the original structure.

clsty: Yes, I think overwritting that script file without modifying config file on the fly is also a reasonable solution.

.config/ags/scripts/color_generation/switchwall.sh
```diff
@@ -3,6 +3,56 @@
 XDG_CONFIG_HOME="${XDG_CONFIG_HOME:-$HOME/.config}"
 CONFIG_DIR="$XDG_CONFIG_HOME/ags"
 
+THUMBNAIL_DIR="/tmp/mpvpaper_thumbnails"
+CUSTOM_DIR="$XDG_CONFIG_HOME/hypr/custom"
+RESTORE_SCRIPT_DIR="$CUSTOM_DIR/scripts"
+RESTORE_SCRIPT="$RESTORE_SCRIPT_DIR/__restore_video_wallpaper.sh"
+EXECS_CONF="$CUSTOM_DIR/execs.conf"
```
clsty: I guess the script do not need this line now.

clsty (changes requested):

clsty:
I'm ok with current codes but I do not have to test it. @end-4 May you test it again? If it works well I suggest to merge it.

PR #1295, opened 2025-05-14
Feat: add additional quick scripts for yay in left sidebar

AD-Archer:
Update yay packages
Remove yay orphan packages


clsty:
Thanks for PR. However TBH this PR is not so useful, and even we would merge it, since we are moving to QuickShell it will be completely aborted soon. Closing.

rllynotfox:
@clsty I don't really think that rewire to other library is a reason to not accept requests, when they are already done by other people

clsty:
> @clsty I don't really think that rewire to other library is a reason to not accept requests, when they are already done by other people

It is simply a part of the reason, not a full reason.

PR #1296, opened 2025-05-15
Feat: add option to clear pacman cache to quick scripts

AD-Archer:


clsty:
Thanks for PR. However TBH this PR is not so useful, and even we would merge it, since we are moving to QuickShell it will be completely aborted soon. Closing.

PR #1299, opened 2025-05-17
Fix: Prevent raw HTML rendering in notifications

SamuelLeutner:
## Problem Description

This PR addresses a notification rendering issue observed in environments like dots-hyprland using ags, where certain notifications, particularly those originating from browsers like Brave or applications like WhatsApp Web, were being displayed with raw HTML content. This included visible tags such as `<a href="...">`, leading to a garbled and unintended display that hinders readability.

`[Notification displaying: <a href="https://web.whatsapp.com/">web.whatsapp.com</a> message]`

closed_notification
notification

## Proposed Solution

To resolve this issue robustly and centrally, the logic within the `processNotificationBody` function has been enhanced. This function, responsible for preparing the notification content before display, now includes a step to actively filter and remove potential HTML tags from the notification body.

A regular expression `/<[^>]*>/g` is applied within `processNotificationBody` to strip all HTML tags. This ensures that, regardless of the notification's origin (even if it contains unexpected markup), only the pure and clean text will be passed to the display widgets. This approach is favored as it consolidates the sanitization logic in a single, dedicated function, making the notification handling more predictable and maintainable.

## How to Test

1. Trigger notifications from applications known to sometimes include HTML in their body, such as receiving a message on WhatsApp Web or getting a notification from a website in Brave browser.
2. Verify that the notification popups or list entries display the message content correctly, without any visible HTML tags, showing only the intended text.

## Additional Considerations

* Implementing the HTML filtering within the `processNotificationBody` function provides a centralized and cleaner way to handle potential markup compared to adjusting properties on multiple display widgets.
* This sanitization step enhances the overall robustness and security of the notification system by preventing the rendering of arbitrary HTML that might be included in notification payloads, intentionally or not.
* While the issue was particularly noticeable with Brave and WhatsApp Web, this solution provides a general safeguard against unexpected HTML in notifications from any source.

### DIFF
--- a/.config/ags/modules/.commonwidgets/notification.js
+++ b/.config/ags/modules/.commonwidgets/notification.js
@@ -39,12 +39,12 @@ function guessMessageType(summary) {
 }
 
 function processNotificationBody(body, appEntry) {
-    // Only process Chrome/Chromium notifications
+    let processedBody = body;
     if (appEntry?.toLowerCase().includes('chrome')) {
-        // Remove the first line
-        return body.split('\n\n').slice(1).join('\n\n');
+        processedBody = body.split('\n\n').slice(1).join('\n\n');
     }
-    return body;
+    processedBody = processedBody.replace(/<[^>]*>/g, '');
+    return processedBody;
 }
 
 const getFriendlyNotifTimeString = (timeObject) => {


end-4:
works, thank you

PR #1303, opened 2025-05-19
Add option to ignore certain apps in dock

LOSEARDES77:
I have scratchpads and i didn't want them to appear in the dock so i added this

For example i put 
```jsonc
"ignoredAppsRegex": ["^kitty-dropterm-.$"],
```
so it will ignore all app clases that match kitty-dropterm-.


end-4:
dock is not enabled by default, but fine

PR #1306, opened 2025-05-21
Set sww to video thumbnail when setting video 

sansmoraxz:
This PR addresses the issue of temporarily sww image being visible after logging in (which would likely be a different still image) and producing some jarring effects. Setting it as the sww wallpaper allows for seamless transition.

### DIFF
--- a/.config/ags/scripts/color_generation/switchwall.sh
+++ b/.config/ags/scripts/color_generation/switchwall.sh
@@ -1,16 +1,16 @@
 #!/usr/bin/env bash
 
 XDG_CONFIG_HOME="${XDG_CONFIG_HOME:-$HOME/.config}"
+XDG_CACHE_HOME="${XDG_CACHE_HOME:-$HOME/.cache}"
 CONFIG_DIR="$XDG_CONFIG_HOME/ags"
+CACHE_DIR="$XDG_CACHE_HOME/ags"
 
-THUMBNAIL_DIR="/tmp/mpvpaper_thumbnails"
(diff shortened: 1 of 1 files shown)


obsidrielle:
I submitted the PR for setting videos as wallpaper a few days ago, but I found that currently there's an issue with mpvpaper - it seems to have a memory leak: if the loop setting is enabled, memory usage continuously increases with each loop cycle. After running for two hours, each mpvpaper process consumes nearly 1.5GB of my memory. Can you set it to restart the mpvpaper process at regular intervals (or better solutions)? Otherwise, it's destined to be unsustainable for long-term use.
However, my request is not related to the purpose of your PR, so you're welcome to ignore it if you wish.  👉👈
#GhostNaN/mpvpaper#101

sansmoraxz:
Damn didn't check that. Yeah the memory leak's painful. Lemme see if I can pull off some alternatives.

sansmoraxz:
One possible option I found is to convert the video to a gif, and then use sww on that. But the conversion process with ffmpeg seems too damn slow, especially if the video is quite large.

obsidrielle:
Currently, my solution is to periodically restart the mpvpaper process, for example once every hour.

sansmoraxz:
Huh, gif is lossy and ends up with larger size, webp ends up making sww lag like crazy. apng is too damn big to load.

Looks like all of the options have one problem or another.

end-4 (commented):
Thanks
It's funny how when I asked copilot to adapt video switching to the new quickshell version of the dotfiles it handled this automatically that I didn't notice this issue

PR #1307, opened 2025-05-21
Adds JSDoc Typing to some of the JavaScript Files

TheGreyfeather:
Adds some more of the obvious types to some of the JavaScript functions, as the rewrite continues I hope to add further typing where applicable.


.config/quickshell/modules/common/functions/file_utils.js
```diff
@@ -1,3 +1,8 @@
+/**
+ * Trims fthe File protocol off the input string
```
end-4: fthe

.config/quickshell/modules/common/functions/string_utils.js
```diff
@@ -1,28 +1,56 @@
...
+ * @param { string } url 
+ * @returns { string| null } 
+ */
 function getDomain(url) {
     const match = url.match(/^(?:https?:\/\/)?(?:www\.)?([^\/]+)/);
     return match ? match[1] : null;
 }
 
+/**
+ * Returns the base path of the passed in url or null
```
end-4: "base url" should be better
"path" could be interpreted as file paths

.config/quickshell/modules/common/functions/string_utils.js
```diff
@@ -1,28 +1,56 @@
...
+ * @param { string } url 
+ * @returns { string | null } 
+ */
 function getBaseUrl(url) {
     const match = url.match(/^(https?:\/\/[^\/]+)(\/.*)?$/);
     return match ? match[1] : null;
 }
 
+/**
+ * Escapes single quotes
```
end-4: mention that this is for shell commands, because it's different. 
normally to escape single quote you just do `\'`
while in shell it has to be `'\''`

.config/quickshell/modules/common/functions/string_utils.js
```diff
@@ -1,28 +1,56 @@
...
+ */
 function shellSingleQuoteEscape(str) {
     //  escape single quotes
     return String(str)
         // .replace(/\\/g, '\\\\')
         .replace(/'/g, "'\\''");
 }
 
+/**
+ * Splits markdown Blocks
```
end-4: why is "blocks" capitalized
it would also be nice to mention that it splits into 3 types: text, think, and code blocks

end-4 (commented):
just some small fixes and should be good

TheGreyfeather:
I'll get these changes made. Thank you for the feedback

.config/quickshell/modules/common/functions/string_utils.js
TheGreyfeather: Would hostname be more appropriate especially as hostname aligns more closely with URL specs?

end-4: well not more appropriate for me ¯\_(ツ)_/¯

end-4:
thanks

PR #1311, opened 2025-05-21
Fix ddcutil to fall back to binary serial number

Nakii46:
# Changes:

- Fix ddcutil to fall back to binary serial number if no serial number is found for a display. 
- Ignore duplicated entries for DisplayPort monitors (or any unresponsive monitors)

## Binary serial number fallback
Hyprland reports the display's binary serial number when the normal one is not found, so I adapted the code to act similarly using ddcutil's full output instead of using `--brief`, as this last one won't report the binary serial number.
<details>
<summary>For reference, this is the output of `hyprctl monitors` in my system</summary>
image_2025-05-21_15-47-26

As you can see the serial number is reported as `0x947020B8`.
</details>
<details>
<summary>And this is the output of `ddcutil detect` in my system</summary>
image_2025-05-21_15-47-58

Here the "Serial number" field is empty, while the "Binary serial number" field does report a value: `2490376376 (0x947020b8)` (of course, we would want to grab the last one in the brackets).
</details>
When using the binary serial number as a fallback we can allow the user to change the brightness of the screen even if their monitor/s don't report a proper serial number. 

The only difference between hyprctl's and ddcutil's binary serial number is that the former one makes the whole number uppercase after the leading `0x`, while ddcutil keeps it all lowercase. This is something we can easily account for, though.

## Duplicated DisplayPort monitors
If a monitor is connected via DisplayPort, there is a chance ddcutil will detect this monitor twice but only one of the two will work with DDC.
`ddcutil detect` will report the display as invalid and the last line for that monitor will say "`This monitor does not support DDC/CI. (I2C slave address x37 is unresponsive.)`".

<details>
<summary>This is how it appears in my system</summary>
image_2025-05-21_15-49-41

This is the duplicated entry for the same monitor as before.
</details>

To be fair, though, I must say that this last fix is only needed because I changed the regular expresion for finding displays to accept displays marked as invalid. I did this because when testing sometimes my monitors would turn up as invalid even when I could change the brigntess manually by command (ddcutil), but it is also true that in my last tests for some reason they weren't being reported as invalid anymore and everything looked normal again. 

I think the regexp can be reverted to it's original state with no issue, but I kinda just left it like that just in case.

(Source: https://www.ddcutil.com/faq/#duplicate_displayport)

### DIFF
--- a/.config/ags/services/brightness.js
+++ b/.config/ags/services/brightness.js
@@ -90,27 +90,44 @@ class BrightnessDdcService extends BrightnessServiceBase {
 async function listDdcMonitorsSnBus() {
     let ddcSnBus = {};
     try {
-        const out = await Utils.execAsync('ddcutil detect --brief');
+        // Its' better not to use --brief. This way if a serial number is not
+        // found we can still use the binary serial number as an alternative
+        const out = await Utils.execAsync('ddcutil detect');
         const displays = out.split('\n\n');
         displays.forEach(display => {
-            const reg = /^Display \d+/;
-            if (!reg.test(display))
+            const reg = /[Dd]isplay/;
+            if (!reg.test(display)) {
                 return;
+            }
             const lines = display.split('\n');
             let sn, busNum;
+            let unresponsive = false;
             for (let line of lines) {
                 line = line.trim()
-                if (line.startsWith('Monitor:')) {
-                    sn = line.split(':')[3];
+
+                // Sometimes ddcutils will report a DP monitor twice, one of the
+                // two copies of the monitor will "not support DDC/CI". Just ignore it
+                // See https://www.ddcutil.com/faq/#duplicate_displayport
+                if (line.includes('unresponsive')) {
+                    unresponsive = true;
+                }
+                if (line.startsWith('Serial')) {
+                    sn = line.split(':')[1].trim();
+                    // Sometimes sn can be empty. In this cases let's relay on binary sn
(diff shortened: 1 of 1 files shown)


.config/ags/services/brightness.js
```diff
@@ -90,27 +90,44 @@ class BrightnessDdcService extends BrightnessServiceBase {
...
+                // two copies of the monitor will "not support DDC/CI". Just ignore it
+                // See https://www.ddcutil.com/faq/#duplicate_displayport
+                if (line.includes('unresponsive')) {
+                    unresponsive = true;
+                }
+                if (line.startsWith('Serial')) {
+                    sn = line.split(':')[1].trim();
+                    // Sometimes sn can be empty. In this cases let's relay on binary sn
+                } else if (line.startsWith('Binary') && !sn) {
+                    // Make the serial number upper case except for the leading '0x' since Hyperland
```
end-4, line 118: hyp**E**rland :sob:

end-4 (commented):
lgtm

end-4:
Thanks for contribution along with the very detailed explanation

.config/ags/services/brightness.js
end-4: fixed in b3e339c

Nakii46:
No problem 🫡
Thanks for the merge

.config/ags/services/brightness.js
Nakii46: Oops, just noticed now 😅

PR #1320, opened 2025-05-25
remove redundant null check on music.js

TheGreyfeather:
Removes a redundant null check that is handled one line above
Additionally removes one unused import


end-4:
why is it redundant?
i understand it could be shorter with `?.` but why remove it totally

TheGreyfeather:
On line 138, the statement `if (!mpris) return;` will return void if mpris is null, so the null case is already being handled. Meaning that those `mpris !== null` are always equivalent to true

end-4:
i see

TheGreyfeather:
For additional context, null is considered a falsy value in JavaScript, which is why that if check behaves how it does.
Thank you for merging this in

PR #1329, opened 2025-05-26
fixed gpt (openrouter) service logic

Et3rnos:
Fixed 2 issues with the openrouter GPT agent:

- Openrouter sends SSE comments to keep the connection alive or to simply signal it is thinking
  - Previous behaviour: they were being displayed as text (`: OPENROUTER PROCESSING`)
  - New behaviour: they are ignored

- Stop payloads can contain actual response data
  - Previous behaviour: responses were being truncated since the last payload content was never being written
  - New behaviour: the whole response is properly displayed to the user

I'm not sure if these problems only affected the OpenRouter agent but the fixes won't cause issues with the other agents, that's for sure.

### DIFF
--- a/.config/ags/services/gpt.js
+++ b/.config/ags/services/gpt.js
@@ -76,7 +76,7 @@ installedOllamaModels.forEach(model => {
 // Custom prompt
 const initMessages =
     [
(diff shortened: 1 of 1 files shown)


end-4 (approved):
lgtm, thanks

PR #1336, opened 2025-05-28
make update.sh script

BeshoyEhab:
Make the update script for who want to update in minutes not an hour to rebuild the PKGBUILD again 😐️

I hope it's right

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -822,8 +822,8 @@ if [[ ! -f "$HOME_UPDATE_IGNORE_FILE" && ! -f "$UPDATE_IGNORE_FILE" ]]; then
   echo "  *.log                 # Ignore all .log files"
   echo "  .config/personal/     # Ignore entire directory"
   echo "  secret-config.conf    # Ignore specific file"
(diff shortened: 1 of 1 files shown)


clsty:
I'm kinda busy right now. I'll review it when possible.

update.sh
```diff
@@ -0,0 +1,829 @@
+#!/bin/bash
```
qualk: use `#!/usr/bin/env bash`

clsty:
The PR looks good to me after a quick view and I do not have time to fully check it. However, I can accept it if when running the script it tells users about what
 this script do, and "it's not fully tested, use it at your own risk". And only continue when user type in a "y"/"Y" (yes). An argument can be provided to skip this, such as `--skip-notice`.

Also you'd better use a git command to test whether it's git repo or not. When `.git` folder exists it may still not be a git repo.

BeshoyEhab:
@clsty I make it

clsty:
Good job. Thanks for your contribution.

PR #1344, opened 2025-06-02
Allow fuzzel.ini to be customized

Sneethe:
Instead of overwriting the entire fuzzel.ini on each theme change. Theme changes are made to `fuzzel.theme` which is then imported by `fuzzel.ini`

Rationale:
I like to use vim binds for fuzzel and there wasn't a good way to modify `fuzzel.ini` without making the end-4 update process complicated.

If you're okay with this, I'll make another PR for ii-qs

### DIFF
--- a/.config/ags/scripts/color_generation/applycolor.sh
+++ b/.config/ags/scripts/color_generation/applycolor.sh
@@ -49,19 +49,17 @@ get_light_dark() {
 
 apply_fuzzel() {
   # Check if template exists
-  if [ ! -f "scripts/templates/fuzzel/fuzzel.ini" ]; then
+  if [ ! -f "scripts/templates/fuzzel/fuzzel.theme" ]; then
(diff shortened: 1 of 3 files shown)


end-4:
the idea is good, but let's not leave it in cache dir because cache dir should be safe to delete
please either copy the generated colors back to .config or generate to state dir

end-4:
I'd prefer it if you not modify it in place but modify in .cache and copy back, but since fuzzel isn't something that's usually actively running and reloading changes it should be fine

Thanks for your contribution

end-4:
as for ii-qs branch, it's fairly simple, i can just do it now

Sneethe:
>I'd prefer it if you not modify it in place but modify in .cache and copy back, but since fuzzel isn't something that's usually actively running and reloading changes it should be fine

It won't be an issue in ii-qs haha. matugen looks great

PR #1351, opened 2025-06-03
record.sh fixes

Sneethe:
Even if ~/Videos exists, `xdg-user-dir VIDEOS` requires `xdg-user-dirs-update` to be run first. Some people just mkdir their desired xdg-user-dir directories. And so they won't have $HOME/.config/user-dirs.dirs for `xdg-user-dir VIDEOS` to work. Now instead if there is no user-dirs.dirs match for VIDEOS then $HOME/Videos/ is used instead.

You can now cancel the region selection, which exits the script. Previously if you pressed escape when choosing the recording region, record-script would just record in  full screen. This resulted in me accidentally recording my entire screen for hours.


end-4:
uhh whoops i totally forgot about this
lgtm, thanks

thyeun:
Sorry to ask, how should i turn on or adding record module into the bar? So, that by left click the record icon will start recording, and right click will stop the recording.

PR #1357, opened 2025-06-04
chore: add fish shell compatible setup link

harilvfs:
as fish is a non-posix-compliant shell, it doesn't support bash syntax like bash <(), so i have changed the url to one that fish shell supports.

btw, sorry i haven't tested this yet. please test the script before merging it should work fine, but just in case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@
    - **Prerequisite**: Your system works. That's it. You don't have to reinstall your system!
    - **Automatic**, but guided and transparent, installation for Arch(-based) Linux:
    ```bash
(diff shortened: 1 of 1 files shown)


clsty:
This one need to be tested and I do not have time currently. Also it's not consistent with the command with argument (which provides installation path). I'm fine if it's added as a tip for fish shell, not modifying the original one.

harilvfs:
@clsty i agree with your statement. i will make a tip or separate instruction for this, and the original will remain the same.
thank you, i will test and give you feedback.

harilvfs:
@clsty I have tested this on Fish shell, it works just fine.

image

clsty:
Thanks for contribution.

PR #1276, opened 2025-05-06
illogical-impulse: Quickshell rewrite

end-4:
## Purpose
I'm opening this PR to gather feedback/suggestions on the rewrite, as well as present and allow for discussion of the list of requirements to fulfill before replacing the main AGS version with the Quickshell one. 

## Current state
- I'm already using this on my system. Not yet as feature-complete as the AGS illogical-impulse
- Key differences: 
  - Widgets more similar to Material 3, scrollables can be flicked, animations are less dull
  - Responsiveness: Launcher takes -2 to 0 instead of around 13 frames to open (relative to ShowMeTheKey)
  - Resources: bar startup time halved, uses 100MB more memory compared to the AGS version

## Todo
- These are to be satisfied before making the Quickshell the main one
- Please let me know if I'm missing something
- [ ] Make installation work
- [ ] Cheatsheet
  - [ ] Edit Hyprland config for native descriptions instead of relying on manual comment parsing
  - [ ] Make the widget
- [ ] AI
  - [ ] Handle code syntax highlighting
  - [ ] Handle Latex rendering
- [ ] Music info widget
  - [ ] Also ditch Pywal, since Quickshell has its own color quantizer

## Testing
- It is recommended to create a new user
- Run the install script and then follow instructions at the top of the readme of the branch

(very incomplete list, I'll edit later, I gotta go)

PR #1362, opened 2025-06-05
QS Rewrite: notification: Fix Chromium-based notifs

xBiei:
connected to #1276

### DIFF
--- a/.config/quickshell/modules/common/widgets/NotificationItem.qml
+++ b/.config/quickshell/modules/common/widgets/NotificationItem.qml
@@ -35,6 +35,22 @@ Item { // Notification item area
 
     implicitHeight: background.implicitHeight
 
+    function processNotificationBody(body, appName) {
+        let processedBody = body
+        
+        // Handle Brave/Chrome notifications - remove first line
+        if (appName && appName.toLowerCase().includes('brave')) {
+            const lines = body.split('\n\n')
+            if (lines.length > 1) {
+                processedBody = lines.slice(1).join('\n\n')
+            }
+        }
+        
+        // Remove HTML tags
+        processedBody = processedBody.replace(/<[^>]*>/g, '')
+        return processedBody
+    }
+
     function destroyWithAnimation() {
         root.qmlParent.resetDrag()
         background.anchors.leftMargin = background.anchors.leftMargin; // Break binding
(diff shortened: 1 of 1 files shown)


end-4:
we don't need this though? for what do you need it

image

`notify-send '<a href="https://web.whatsapp.com/">web.whatsapp.com</a>' '<a href="https://web.whatsapp.com/">example.com</a>'`

xBiei:
> we don't need this though? for what do you need it
> 
> 
> 
> image
> 
> 
> 
> `notify-send '<a href="https://web.whatsapp.com/">web.whatsapp.com</a>' '<a href="https://web.whatsapp.com/">example.com</a>'`
> 
> 
> 
> 

The reason is that chromium-based browsers send a link as the first line of the notification which makes the content of the notif unreadable unless expanded, so removing the first line seems like a good solution.
I do have another idea about this, if I remember correctly, the chromium notifs don't have an "open" button and if they did it doesn't work, so we could use the first line content as the action of the button before removing it.

end-4:
> The reason is that chromium-based browsers send a link as the first line of the notification which makes the content of the notif unreadable unless expanded, so removing the first line seems like a good solution.
I do have another idea about this, if I remember correctly, the chromium notifs don't have an "open" button and if they did it doesn't work, so we could use the first line content as the action of the button before removing it.

can you copy and send the notif so at i have something to test/work with?

xBiei:
> can you copy and send the notif so at i have something to test/work with?

I'm currently traveling and my main is at home, I'll be returning tomorrow night.

But for now I found [an example you can check out](https://github.com/dunst-project/dunst/issues/716#issue-619316791).

You can also use [this website](https://davidwalsh.name/demo/notifications-api.php) to send a test notification.

end-4:
shows fine for me?
(although i should change the link color)
image

xBiei:
> shows fine for me? (although i should change the link color)

well what I meant is, the body of the notif should be what's after the link
meaning 
```
example.com
the actual body of the notif
```
so the notif would look like this
image

end-4:
i see, but what your code does is: for brave notifications, it
- removes the first line if it's separated from the rest, and
- destroys all xml tags, including that of links that might originally be clickable in the body

please detect if the first line is separated from the rest and is a link to only remove that line, and allow this processing to be run for any app or include a decent set of chromium browsers to check instead of just brave

xBiei:
> i see, but what your code does is: for brave notifications, it
> 
> * removes the first line if it's separated from the rest, and
> * destroys all xml tags, including that of links that might originally be clickable in the body
> 
> please detect if the first line is separated from the rest and is a link to only remove that line, and allow this processing to be run for any app or include a decent set of chromium browsers to check instead of just brave

This code was initially taken from ii-ags but I edited it to be compatible with qs, and didn't really check all the cases.
But since I'm editing this, I'll do some changes so it would fix the first line issue without messing up other stuff.
Running this code on any app might cause issues so I'm not gonna do that, instead I'll check for the popular chromium-based browsers on linux and include them while keeping the comment above the code.

xBiei:
done

end-4:
thank you for your contribution

PR #1363, opened 2025-06-05
QS Rewrite: bar: Add an option for choosing screens

xBiei:
connected to #1276

### DIFF
--- a/.config/quickshell/modules/bar/Bar.qml
+++ b/.config/quickshell/modules/bar/Bar.qml
@@ -29,7 +29,7 @@ Scope {
     Variants { // For each monitor
         model: {
             const screens = Quickshell.screens;
-            const list = ConfigOptions.bar.screensList;
+            const list = ConfigOptions.bar.screenList;
             if (!list || list.length === 0)
                 return screens;
             return screens.filter(screen => list.includes(screen.name));


end-4:
this should work but might not help with memory consumption
it's better if you filter the `Variants`' model

xBiei:
done

.config/quickshell/modules/bar/Bar.qml
```diff
@@ -25,10 +25,18 @@ Scope {
...
+    property var filteredScreens: {
+        const list = ConfigOptions.bar.screensList;
+        if (!list || list.length === 0)
+            return Quickshell.screens;
+        return Quickshell.screens.filter(screen => list.includes(screen.name));
     }
 
     Variants { // For each monitor
-        model: Quickshell.screens
+        model: bar.filteredScreens
```
end-4: use a Quickshell ScriptModel to prevent recreation of every bar when this changes

.config/quickshell/modules/common/ConfigOptions.qml
```diff
@@ -45,6 +45,10 @@ Singleton {
             property bool alwaysShowSwap: true
             property bool alwaysShowCpu: false
         }
+        property list<string> screensList: [
+            "HDMI-A-1",
+            "DP-1"
+        ]
```
end-4: should be empty by default and in that case take all screens

xBiei: I think commenting the current list would be better so anyone can know what's the expected values for this option

end-4: `property list<string> screensList: [] // List of names, like "eDP-1", find out with 'hyprctl monitors' command`

.config/quickshell/modules/bar/Bar.qml
xBiei: I tried using ScriptModel but the bar is still being recreated which is weird, its exactly the same behavior.
Anyways I'm gonna add it

.config/quickshell/modules/bar/Bar.qml
```diff
@@ -25,10 +25,22 @@ Scope {
...
+            const list = ConfigOptions.bar.screensList;
+            if (!list || list.length === 0)
+                return screens;
+            return screens.filter(screen => list.includes(screen.name));
+        }
     }
 
     Variants { // For each monitor
-        model: Quickshell.screens
+        model: screensModel.values
```
end-4: why not put the ScriptModel in the model directly

xBiei: It literally took me about an hour trying that and it never worked, even tho logging what the current "values" return shows the correct screens, the model in variants always logs a warning `Non list data QVariant(QObject*, 0x7f3d3007e300) assigned to Variants.model, Ignoring.`
I made sure its a list, tried recreating lists then filtering which makes a new list too, but still didn't work.
When I moved the ScriptModel out and used its id as a ref, didn't work.
only the current implementation worked, I'm still digging into what's causing this.

xBiei: after digging some more, looks like `model` property in `Variants` only accepts lists (it was mentioned in the docs but I also checked the source code) which is why `Quickshell.screens` or `screensModel.values` work but using `ScriptModel` directly won't.
I'll check with outfoxxed for now to see if we can find a better solution.

end-4: Quickshell.screens is an ObjectModel, not an array
for an array of it you use .values
so for the filtering take values from Quickshell.screens.values

xBiei: I'm pretty sure `Quickshell.screens` is an array (list), and no, using `.values` doesn't work either.
The screens inside the list are objects

end-4: i was wrong indeed
just ignore that recreation/weird scriptmodel i guess, shouldn't happen that often

xBiei: I sent a message on quickshills discord, if no solution for now then we can just use the previous implementation.
But for now we can just merge it (since you want to merge to main asap) and edit it after we have our response from outfoxxed.

EDIT:
I actually edit the screenslist a lot

PR #1365, opened 2025-06-05
QS Rewrite: bar: add microphone control to util buttons

xBiei:
connected to #1276

### DIFF
--- a/.config/quickshell/modules/bar/UtilButtons.qml
+++ b/.config/quickshell/modules/bar/UtilButtons.qml
@@ -5,6 +5,7 @@ import QtQuick.Layouts
 import Quickshell
 import Quickshell.Io
 import Quickshell.Hyprland
+import Quickshell.Services.Pipewire
 
 Rectangle {
     id: root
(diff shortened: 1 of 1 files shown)


end-4:
this should be configurable
please put it in a loader and have it active only when `bar.utilButtons.enableMicToggle` is true
toggling mic is not something i would do frequently

xBiei:
ok done, but since I've add an option for 1 button, I thought it would be better to add an option for all buttons for more customization.

.config/quickshell/modules/common/ConfigOptions.qml
```diff
@@ -36,6 +36,12 @@ Singleton {
             property bool alwaysShowSwap: true
             property bool alwaysShowCpu: false
         }
+        property QtObject utilButtons: QtObject {
+            property bool showMicToggle: true
+            property bool showKeyboardToggle: true
+            property bool showScreenSnip: true
+            // property bool showColorPicker: true
```
end-4: since it's configurable, no reason to leave color picker commented out

end-4 (commented):
looks good

.config/quickshell/modules/common/ConfigOptions.qml
xBiei: I thought you had another reason for commenting the button even tho it worked for me
anyways, updating now...

end-4:
the loaders need to have `visible` set to false along with `active` when the corresponding option is false, or they'll do this funky 
image

end-4:
okay thanks

PR #1376, opened 2025-06-09
QS Feature: better dock

### LINKED ISSUE #1375: [FR ii-qs] unify pinned apps and running apps in dock

it will be great if we have a dock like this...
_I have some codes right now and it just works as the screenshot, but there's some (critical) bugs, i'm trying to fix it_

### How will it help?

The current dock displays pinned apps and running apps separately, while desktops KDE, GNOME, chromeos, they display these two parts together

### Extra info

Image

0xarch:
The pinned apps are along with running apps. `DockAppButton` and `DockSeparator` are inside a `RowLayout` container (Every `DockAppButton` has a `DockSeparator` with it and we only need the one next to the first-unpinned-running app. I have no better idea how to deal with this as my QML knowledge is really poor)

Closes #1375 if this works fine

### DIFF
--- a/.config/quickshell/modules/dock/DockAppButton.qml
+++ b/.config/quickshell/modules/dock/DockAppButton.qml
@@ -21,13 +21,14 @@ DockButton {
     property real iconSize: 35
     property real countDotWidth: 10
     property real countDotHeight: 4
-    property bool appIsActive: appToplevel.toplevels.find(t => (t.activated == true)) !== undefined
+    property bool appIsActive: !appToplevel.toplevels ? false : appToplevel.toplevels.find(t => (t.activated == true)) !== undefined
+    property DesktopEntry entry: DesktopEntries.byId(modelData.appId)
 
     MouseArea {
         id: mouseArea
         anchors.fill: parent
         hoverEnabled: true
-        acceptedButtons: Qt.NoButton
+        acceptedButtons: Qt.LeftButton | Qt.MiddleButton
         onEntered: {
             appListRoot.lastHoveredButton = appButton
             appListRoot.buttonHovered = true
--- a/.config/quickshell/modules/dock/DockAppButton.qml
+++ b/.config/quickshell/modules/dock/DockAppButton.qml
@@ -38,10 +38,18 @@ DockButton {
                 appListRoot.buttonHovered = false
             }
         }
-    }
-    onClicked: {
-        lastFocused = (lastFocused + 1) % appToplevel.toplevels.length
-        appToplevel.toplevels[lastFocused].activate()
+        onClicked: {
+            if (mouse.button === Qt.MiddleButton ) {
+                appButton.entry.execute();
+                return;
+            }
+            if (modelData.isPinnedApp && !modelData.isRunning) {
+                appButton.entry.execute();
+                return;
+            }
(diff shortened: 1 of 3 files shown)


0xarch:
P.S. no animation for now... this patch just works

end-4:
uhh whoops i also tried to implement this 5c2b12bf96af9ca0cafacf46bbc58c1cb2a5d348
what do you mean by animations?

.config/quickshell/modules/dock/DockAppButton.qml
```diff
@@ -21,13 +21,14 @@ DockButton {
...
-    property bool appIsActive: appToplevel.toplevels.find(t => (t.activated == true)) !== undefined
+    property bool appIsActive: !appToplevel.toplevels ? false : appToplevel.toplevels.find(t => (t.activated == true)) !== undefined
+    property DesktopEntry entry: DesktopEntries.byId(modelData.appId)
 
     MouseArea {
         id: mouseArea
         anchors.fill: parent
         hoverEnabled: true
-        acceptedButtons: Qt.NoButton
+        acceptedButtons: Qt.LeftButton | Qt.MiddleButton
```
end-4, line 31: this MouseArea is accepting buttons so it's hogging click signals from the actual RippleButton with its own MouseArea to trigger ripple anims

.config/quickshell/modules/dock/DockApps.qml
```diff
@@ -39,23 +41,55 @@ Item {
...
+                        id: separator
+                        visible: buttonIndex === pinnedCount
+                    }
+
+                    DockAppButton {
+                        id: appButton
+                        appToplevel: modelData
+                        appListRoot: root
+                    }
+                }
```
end-4, line 92: similar to the DockSeparator, when it's a separator, we shouldn't need an app button

.config/quickshell/modules/dock/DockApps.qml
```diff
@@ -39,23 +41,55 @@ Item {
...
+                implicitHeight: appButton.implicitHeight
+
+                RowLayout {
+
+                    spacing: rowLayout.spacing
+
+                    DockSeparator {
+                        id: separator
+                        visible: buttonIndex === pinnedCount
+                    }
```
end-4, line 85: when an item is not always used, it's best to wrap it in a Loader and set `active` prop to true only when it's used
so something like
```qml
Loader {
    visible: buttonIndex === pinnedCount // In this case, it's also a good idea to hide the loader because you're putting it in a RowLayout
    active: buttonIndex === pinnedCount // This is the real thing you need
    sourceComponent: DockSeparator { // Note that just putting this DockSeparator here is not sufficient. It has to explicitly be set to `sourceComponent` or the Loader won't act on it
        id: separator
    }
}
```

end-4 (commented):
Since I already solved it, the only thing I could do now is give you some feedback
I'm sorry and thank you for your attempt

0xarch:
great lesson for my QML, appreciate

0xarch:
> what do you mean by animations?

there should be a slide-in/out animation for icons, or even separators i think?

end-4:
> > what do you mean by animations?
> 
> there should be a slide-in/out animation for icons, or even separators i think?

i might be able to implement them with listview

PR #1384, opened 2025-06-10
fix(font): Kitty now follows font size from kitty.conf

JxJxxJxJ:
image


JxJxxJxJ:
Don't mind the time command, I was checking if it had any noticeable slowdown and it hadn't

end-4:
isn't that already the default config location
i don't see any difference when running this new command
and the font size is 11 not 14

JxJxxJxJ:
Like, when changing that config file and running kitty with -1 the changes to the font wouldn't get applied in my case. I changed from 11 -> 14 and until I ran kitty explicitly with that config file the terminal kept spawning with size 11

end-4:
`-1` flag = try to open a new window instead of a new instance
your config changes failed to apply because you still had another instance open and the config wasn't reloaded

PR #1385, opened 2025-06-10
Add systemsettings as dependency (KDE desktop config)

JxJxxJxJ:
Otherwise the gear from the right pannel does nothing


end-4:
thanks

PR #1408, opened 2025-06-14
Add terminalApp global for kde Apps

JxJxxJxJ:
This helps with "Open with neovim" in dolphin.

Although it doesn't apply the quickshell theme since it opens kitty and neovim without going through a shell.

I'm not sure what would be the proper solution

### DIFF
--- a/.config/kdeglobals
+++ b/.config/kdeglobals
@@ -137,6 +137,7 @@ ColorScheme=MaterialYouDark
 ColorSchemeHash=3c0cecefbea43cdb8fe3da156e4a106f7384a526
 LastUsedCustomAccentColor=184,117,220
 XftHintStyle=hintslight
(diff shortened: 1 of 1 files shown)


end-4:
running it through fish should apply the colors, though idk if you want that

JxJxxJxJ:
Yeah that's what I've been trying. Do you know if that would be possible with a command? I've been trying and I can't get it to work, I've been trying things like`terminalApplication=kitty -e fish` but couldn't quite get it yet

end-4:
fish needs to have `-c` maybe?
i can't be bothered to figure out because i use everything graphical

JxJxxJxJ:
Yeah all good, -c just opens a terminal that closes immediatly

end-4:
then i will just merge this, thanks

PR #1412, opened 2025-06-15
Fix an error in env file

BeshoyEhab:


clsty:
Thanks. I was using my phone to do that quick fix, sorry for that mistake.

PR #1421, opened 2025-06-15
Added ability to hide/show the top bar on keypress

Fane447:
It's sometimes useful to hide the top bar, such as with the recent game The Alters which crashes when fullscreen so you need to play windowed and have the bar hidden, or with an OLED monitor you don't always want static assets always on the screen.

Tested on a clean install and upgrading existing installs to ensure the default behavior is to show the bar.


miguelfm:
There is a GlobalShortcut in Bar.qml named "barToggle" that permits to toggle de top bar's visibility. 
By using this shortcut anyone can add a new custom keyboard bind in  ~/.config/hypr/custom/keybinds.conf like this:
`unbind = Super+Alt, T`
`bind = SUPER+Alt,T ,exec, hyprctl dispatch global quickshell:barToggle # Toggle top bar visibility`

end-4:
I added it at some point but forgot to base it on this PR
I'm sorry

The default keybind for it is Super+J

PR #1428, opened 2025-06-16
Fixed Terminal Color generation bug for quickshell

hrsvrn:
Recently upgraded from ags to quickshell and the terminal colors werent syncing with wallpaper i put

the source file from ~/.config/zshrc.d/dots-hyprland.zsh wasnt working since it was using the outdated ags generated cache.

Removed the ags one and added the quickshell color gen

Now it works perfectly


end-4:
thanks for your contribution
i don't use zsh so i just neglect it

PR #1431, opened 2025-06-17
Add dolphin as dependency

JxJxxJxJ:
It wasn't being installed by default

### DIFF
--- a/arch-packages/illogical-impulse-kde/PKGBUILD
+++ b/arch-packages/illogical-impulse-kde/PKGBUILD
@@ -10,5 +10,6 @@ depends=(
 	networkmanager
 	plasma-nm
 	polkit-kde-agent
+ 	dolphin
 	systemsettings
 )


JxJxxJxJ:
I mean I just installed in two minimal arch installations and that was the case, not sure if anything is wrong with editing this file in particular

clsty:
`dolphin` is only used here in this repo:
https://github.com/end-4/dots-hyprland/blob/dd2e2a9bca673cbd801c71e21d1cf09884e9e6a7/.config/hypr/hyprland/keybinds.conf#L205

It should not be a dependency, unless you can make it an optional dependency.

JxJxxJxJ:
The dock otherwise starts with no icon by default too

end-4:
uhhhh actually
if we're already installing kde stuff then installing dolphin won't hurt, it should ease the experience a bit :thinking:

end-4:
actually, since we even install a settings app that isn't meant to be used outside of the DE's environment, it would be absurd to not include the file manager because it's actually useful

Issue #1440, opened 2025-06-17
Music Player displays >=2 instances when paused


WoojinJung-04:
### Step 1. Before you submit

- [x] I have read the [Troubleshooting](https://end-4.github.io/dots-hyprland-wiki/en/ii-qs/04troubleshooting/) and [Usage](https://end-4.github.io/dots-hyprland-wiki/en/ii-qs/02usage/) pages.
- [x] I've successfully updated to the latest version following the [guidance](https://end-4.github.io/dots-hyprland-wiki/en/ii-qs/01setup/#updating).
- [x] I've successfully updated the system packages to the latest.
- [ ] I've ticked the checkboxes without reading their contents

### Step 2. Quick diagnose info

Not needed

### Step 3. Describe the issue

![Image](https://github.com/user-attachments/assets/4017ae0d-8bcc-4b5e-a888-c3827cd5b225)

I discovered that the music player application displays >= instances when the browser is stopped. I was using YT Music on Brave Browser to do this. I have confirmed that this is consistent in different browsers such as firefox.

It would be really nice if the instance could be filtered so that when there are >= 2 instances, or even better, is fixed. 

Thank you for your config files btw. They are wonderful.

### Reminder

- [x] I agree that it's usually impossible for others to help me without my logs.

PharmaRacist:
is it wilth all chromium based browsers only ?

WoojinJung-04:
No it was the same with firefox too if I remember correctly

PR #1443, opened 2025-06-17
Update MediaControls.qml to filter "YouTube Music" included players

WoojinJung-04:
## Describe your changes

It filters tracks including "YouTube Music" in the player.trackTitle 

It is a crude fix, but it works. Further implementations should look for more efficient ways of doing this.

### DIFF
--- a/.config/quickshell/modules/mediaControls/MediaControls.qml
+++ b/.config/quickshell/modules/mediaControls/MediaControls.qml
@@ -45,6 +45,9 @@ Scope {
         let filtered = [];
         let used = new Set();
 
+        // Skip unwanted players
+        if (p1.trackTitle.includes("YouTube Music")) continue;
+
         for (let i = 0; i < players.length; ++i) {
             if (used.has(i)) continue;
             let p1 = players[i];


WoojinJung-04:
Closes issue #1440

sansmoraxz:
Mind adding amazon prime music?

image

WoojinJung-04:
I also experimented with isValidPlayer from mpris and it doesn't seem to do anything.

.config/quickshell/modules/mediaControls/MediaControls.qml
```diff
@@ -45,6 +45,9 @@ Scope {
         let filtered = [];
         let used = new Set();
 
+        // Skip unwanted players
+        if (p1.trackTitle.includes("YouTube Music") || p1.trackTitle.includes("Amazon Prime Music")) continue;
+
         for (let i = 0; i < players.length; ++i) {
             if (used.has(i)) continue;
             let p1 = players[i];
```
sansmoraxz: shouldn't your changes be after the declaration?

WoojinJung-04: Oh lol my bad I was too lazy to use git for one line will fix

end-4:
make a `music.blacklist` config option

WoojinJung-04:
okey

end-4:
does the above help? i noticed the time of the dupes are similar. if it works then we now have a more elegant solution

WoojinJung-04:
Yes! It works. Can we additionally add && p2.includes("Music") || p2.includes("Browser") for safety? 

It won't hurt b/c it will only filter out ones including "Music" and are small enough.

sansmoraxz:
> Yes! It works. Can we additionally add && p2.includes("Music") || p2.includes("Browser") for safety?
> 
> It won't hurt b/c it will only filter out ones including "Music" and are small enough.

Some songs have "Music" in their title too.

What about while playing this?
https://www.youtube.com/watch?v=SRXH9AbT280

WoojinJung-04:
It will only remove a player if there's a player that has a smaller than 2 difference and has Music inside the title. We're basically restricting the scope of the filter, not adding it. That's why I said it won't hurt.

WoojinJung-04:
If we need extra safety we could also filter out for ones without an artist

end-4:
1. why are you adding an else if clause that does the same thing as the other if clause
2. p2 is an object, not a string or array and we can't use `.includes`

end-4:
> Yes! It works. Can we additionally add && p2.includes("Music") || p2.includes("Browser") for safety?
> 
> It won't hurt b/c it will only filter out ones including "Music" and are small enough.

If it's already fixed and fine, no need to add a redundant and very specific check
I'll add it if you find any problematic case

PR #1447, opened 2025-06-17
Add loupe (img viewer used by default) as dependency

JxJxxJxJ:


JxJxxJxJ:
Otherwise imgs are opened in the browser after installation by default

end-4:
actually it's not even used now, there's `xdg-open`
i'll just remove that loupe thing

PR #1462, opened 2025-06-19
removed nonexistent script

hyprbased:
your AGS scripts are not available in the current install.


end-4:
i'll just add it back to the quickshell version

PR #1463, opened 2025-06-19
Use hyprsunset as blue light filter

JxJxxJxJ:
Before with gammastep it didn't work for me, now it does without problems

### DIFF
--- a/.config/quickshell/modules/sidebarRight/quickToggles/NightLight.qml
+++ b/.config/quickshell/modules/sidebarRight/quickToggles/NightLight.qml
@@ -9,33 +9,37 @@ QuickToggleButton {
     property bool enabled: false
     toggled: enabled
     buttonIcon: "nightlight"
+
     onClicked: {
         nightLightButton.enabled = !nightLightButton.enabled
         if (enabled) {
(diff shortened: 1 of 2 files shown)


WoojinJung-04:
I tried jamming the button and this works great unlike the previous version

end-4:
gammastep automatically triggers at night. we are not switching without this functionality implemented in a quickshell service

JxJxxJxJ:
Is there anything like an actual quickshell service feature? Or should it be implemented in other way?

JxJxxJxJ:
Gammastep does nothing on my PC, not sure how to debug it, I have it installed but running the command `gammastep` does nothing really

end-4:
> Is there anything like an actual quickshell service feature? Or should it be implemented in other way?

we should implement a quickshell service

end-4:
hyprsunset has been implemented in 574a2a11e76f1c843ab4c7a58f5fec236fecd9d9
I should've based my work off this, but in the end everything that would be brought over is just the listing of the dependency in the pkgbuild, which is a trivial thing I hope you don't mind not getting GitHub's recognition for.

PR #1470, opened 2025-06-19
Bar.qml: Fix round decorators transparency and visibility consistency with bar

Sighthesia:
## Describe your changes
1. Fixed visibility consistency: Corner round decorators now properly hide when `showBackground` is false
2. Fixed transparency consistency: Corner round decorators now use the same transparency value with the main bar

- Before: 
swappy-20250619-202904
swappy-20250619-202442

- After:
swappy-20250619-204217
swappy-20250619-204644

Ready, at least it looks good on the surface.


end-4:
the visibility thing is valid, but the corner opacity fix is solving the solution at the wrong place. the corners' color should already be transparent...

end-4:
i can't figure out how to properly fix it, i'll accept this hack for now, i don't care that much about transparency anyways

PR #1475, opened 2025-06-19
added url bookmarks to searchwidget

hyprbased:
add basic fuzzy url bookmarks via json file.

requires QML_XHR_ALLOW_FILE_READ=1 in the environment to read the json properly

feel free to take it out later when/if you revamp the search functionality and or add multi-browser bookmark syncing via extensions, or better ideas (many better ways to do it)

### DIFF
--- a/.config/myurls.json
+++ b/.config/myurls.json
@@ -1,0 +1,4 @@
+[
+    { "name": "GitHub", "url": "https://github.com" },
+    { "name": "YouTube", "url": "https://youtube.com" }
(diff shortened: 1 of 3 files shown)


hyprbased:
do you have a better method you would like to implement?

end-4:
- we already have something called persistent clipboard, with cliphist
- about the code
  - you unnecessarily reimplemented fuzzy search
  - you included your other irrelevant personal changes

PR #1480, opened 2025-06-20
todo popup script for keybinds

hyprbased:
fuzzel popup saves to todo's json

i made todowidget refresh every 3 seconds when visible so that new items are actually show

hyprland keybind can be 

bind = ,(your fav key), exec, ~/.config/quickshell/hypr/hyprland/scripts/add_todo_fuzzel.sh


hyprbased:
for some reason the json file empties after some time

end-4:
image

hyprbased:
@end-4  wow didnt see that. thanks

Issue #1466, opened 2025-06-19
Dark Mode switcher in Right menu


WoojinJung-04:
### What would you like to be added?

![Image](https://github.com/user-attachments/assets/6a0fed63-4d57-4a5e-828a-6bb622d16789)

Currently, applying light/dark mode can only be done from the welcome menu. It will be really convenient if a dark mode/light mode toggle button was on this menu

end-4:
we already have a dark/light mode switcher button on the bar (see the above pr) so now we don't need this

PR #1481, opened 2025-06-20
feat(modules/bar): add nightmode button

tekeoglan:
## Describe your changes

* Added a `darkmode` property to the `ConfigOptions`  with the following sub-properties:

  * `show`: Determines whether to display the dark mode toggle button
  * `lightImgPath`: Path to the image used in light mode
  * `darkImgPath`: Path to the image used in dark mode
* Created a dark mode toggle button on the bar so users can easily switch between dark and light modes

* Occasionally, `ConfigLoader` fails to load the config file properly, resulting in empty image path values. However, this issue is related to `ConfigLoader`, not this feature—so it's fair to say it's ready.
* As a precaution, image paths are assigned directly from `ConfigOptions` instead of being loaded from the config file.
* This feature was created to address issue #1466.


tekeoglan:
This is how it looks,

https://github.com/user-attachments/assets/b4b8ef04-299c-4fcb-b2f8-048669a7a51f

PharmaRacist:
but no one uses 2 wallpapers only right ? 
we need and array to assaign wallpaper group by names and it should be easy for the user to do so

tekeoglan:
Yes, you are absolutely right. I shared this as a quick fix until a better solution comes along. Alternatively, I could remove the image paths so that it only functions as a dark mode toggle.

PharmaRacist:
image
currently im having this wall selector im thinking of devided imageItem and assaigned stuff by name like the images named x-dark and x-light for examples are grouped sth like that but this is a big hassle

.config/quickshell/modules/bar/UtilButtons.qml
```diff
@@ -10,7 +10,8 @@ import Quickshell.Services.Pipewire
 Item {
     id: root
     property bool borderless: ConfigOptions.bar.borderless
-    implicitWidth: rowLayout.implicitWidth + rowLayout.spacing * 2
+    implicitWidth: rowLayout.implicitWidth + rowLayout.spacing * 16
```
end-4: why

end-4 (commented):
alright

.config/quickshell/modules/bar/UtilButtons.qml
end-4: this allocates too much space for the util buttons, no point

end-4:
thanks

end-4:
ohhhh you put the button outside the row layout so the width was weird, ill fix it

end-4:
529b67d7286ff560148ffce3872dba5fe463048b

Issue #1510, opened 2025-06-24
System Tray broke after Quickshell restart.


Bugraern:
### Step 1. Before you submit

- [x] I have read the [Troubleshooting](https://end-4.github.io/dots-hyprland-wiki/en/ii-qs/04troubleshooting/) and [Usage](https://end-4.github.io/dots-hyprland-wiki/en/ii-qs/02usage/) pages.
- [x] I've successfully updated to the latest version following the [guidance](https://end-4.github.io/dots-hyprland-wiki/en/ii-qs/01setup/#updating).
- [x] I've successfully updated the system packages to the latest.
- [ ] I've ticked the checkboxes without reading their contents

### Step 2. Quick diagnose info

<details><summary>Quick diagnose</summary>

```

/========================\
| Checking git repo info |
\========================/

[===diagnose===] git remote get-url origin
https://github.com/end-4/dots-hyprland
[---SUCCESS---]

[===diagnose===] git rev-parse HEAD
ad462649b0c41b87f91b0c4dafcf434148ca4ea1
[---SUCCESS---]

/=================\
| Checking distro |
\=================/

[===diagnose===] ii_check_distro
LSB Version:	n/a
Distributor ID:	cachyos
Description:	CachyOS
Release:	rolling
Codename:	n/a
[---SUCCESS---]

/====================\
| Checking variables |
\====================/

[===diagnose===] declare -p XDG_CACHE_HOME
declare -x XDG_CACHE_HOME="/home/bupte/.cache"
[---SUCCESS---]

[===diagnose===] declare -p XDG_CONFIG_HOME
declare -x XDG_CONFIG_HOME="/home/bupte/.config"
[---SUCCESS---]

[===diagnose===] declare -p XDG_DATA_HOME
declare -x XDG_DATA_HOME="/home/bupte/.local/share"
[---SUCCESS---]

[===diagnose===] declare -p XDG_STATE_HOME
declare -x XDG_STATE_HOME="/home/bupte/.local/state"
[---SUCCESS---]

[===diagnose===] declare -p ILLOGICAL_IMPULSE_VIRTUAL_ENV
declare -x ILLOGICAL_IMPULSE_VIRTUAL_ENV="~/.local/state/quickshell/.venv"
[---SUCCESS---]

/============================\
| Checking directories/files |
\============================/

[===diagnose===] ls -l /home/bupte/.local/state/quickshell/.venv
total 16
[...]

PR #1520, opened 2025-06-27
feat(modules/bar): add weather bar

tekeoglan:
## Describe your changes
I created a `WeatherBar` component that displays the current temperature. Users can also view detailed weather conditions by hovering over the `WeatherBar`.

Yes, it is ready to use. However, I wasn’t able to test it on different resolutions or dual-monitor setups, as
I don’t have access to those configurations.

https://github.com/user-attachments/assets/d075403c-24e8-450d-a355-6ab7f95480e0


end-4:
I do not like this feature and will only include it if it is seamless and absolutely not change any behavior by default. So two problems:
- Design
  - Your placement of the widget currently misaligns the workspace widget to the left
  - The font does not match the rest of the system (Use instead of Text)
- Functionality
  - The city should be automatically detected. If this is not possible or not reliable, I don't want to have this
  - It should be hidden by default, and you should use a Loader to fully hide *and not load* it, not set its visibility to false in a BarGroup that is still always visible

tekeoglan:
I noted the request, but I’ve already spent a lot of time getting this version to work. I need a breather first—then I’ll come back and try to fix the issues.

tekeoglan:
Hey @end-4, I’ve fixed all the issues except for automatically detecting the location for the weather service. Here are the challenges:

There are two main ways to detect a user's location:
1. Using `GeoClue`, which requires a Google API key (with billing enabled), [here](https://www.linuxfromscratch.org/blfs/view/svn/basicnet/geoclue2.html).
2. Using a web request to an IP-based geolocation service, which isn’t reliable for users on VPNs.

I think the most flexible approach would be to add a section to the `settings` panel where users can choose how they want to provide location data—either by city name or GPS coordinates.

tekeoglan:
I’ve added some preliminary changes. Based on your input, I’ll go ahead and implement the suggested updates.

end-4:
for the two location options
1. i didn't know geoclue broke... it doesn't seem to be broken? we're already using it for nightlight. but if it's not supposed to work then we should move away i guess
2. it's on them for using a vpn. just provide a way to override the city name

let's go with (2), if you can find a reliable service

tekeoglan:
@end-4, I’ve made the necessary changes. It detects the location using `geoclue` by default ([check here](https://doc.qt.io/qt-6/qml-qtpositioning-positionsource.html#start-method)). Earlier, I wasn't able to get this work because the `geoclue` agent wasn’t running—this turned out to be due to a broken path in the Hyprland config, which I’ve also fixed.

There might still be some logical bugs in the GPS service logic, and unfortunately, there isn’t an extensive way to test it thoroughly. The intended logic is as follows:

If the location is initialized correctly, it will be used as—even if GPS fails afterward, the initial location and will remain in use until the next successful location retrieval.

If it fails to initialize (i.e., no location service backend is available), it will fall back to the city name provided in `ConfigOptions`."

tekeoglan:
I didn’t add a section to the settings panel for users to override the city name, because that would also require adding other options (e.g., enableGps, interval), which would clutter the settings panel. Instead, users can modify the config file directly if they're not satisfied with the defaults.

tekeoglan:
> I didn’t add a section to the settings panel for users to override the city name, because that would also require adding other options (e.g., enableGps, interval), which would clutter the settings panel. Instead, users can modify the config file directly if they're not satisfied with the defaults.

If having these settings in the panel would be more convenient and clutter isn’t a major concern, I can add them to the panel as well.

.config/quickshell/modules/bar/Bar.qml
```diff
@@ -307,6 +306,22 @@ Scope {
...
+                Loader {
+                    id: weatherLoader
+                    active: Config.options.bar.weather.show
+                    anchors.left: middleSection.right
+                    anchors.margins: 10
+                    sourceComponent: BarGroup {
+                        implicitHeight: Appearance.sizes.baseBarHeight
+                        height: Appearance.sizes.barHeight
+                        WeatherBar {}
+                    }
```
end-4: don't anchor, properly put in the rowlayout of the right section

.config/quickshell/modules/bar/Bar.qml
```diff
@@ -315,34 +330,33 @@ Scope {
                     anchors.right: parent.right
                     implicitHeight: Appearance.sizes.baseBarHeight
                     height: Appearance.sizes.barHeight
-                    width: (barRoot.width - middleSection.width) / 2
+                    width: barRoot.width - (barLeftSideMouseArea.width + middleSection.width + weatherLoader.width)
```
end-4: after the above this manual change shouldn't be necessary

.config/quickshell/modules/bar/weather/WeatherBar.qml
```diff
@@ -0,0 +1,68 @@
...
+                font.pixelSize: Appearance.font.pixelSize.large
+                color: Appearance.colors.colOnLayer1
+                text: WeatherService.data.temp
+            }
+        }
+    }
+
+    LazyLoader {
+        id: popupLoader
+        active: true
```
end-4: why not set `active` instead of setting `visible`?

.config/quickshell/modules/common/Config.qml
```diff
@@ -122,6 +122,18 @@ Singleton {
...
+                    // gps based location
+                    property bool enableGPS: true
+                    // use if 'enableGPS' is false
+                    property string city: "Istanbul"
+                    // use uscs units
+                    // by default use metric (SI) units
+                    property bool useUSCS: false
+                    // in minutes
+                    property int fetchInterval: 10
+                }
```
end-4: - Leave city blank
- rename `show` to `enable`

.config/quickshell/services/WeatherService.qml
end-4, line 1: since it's already in `services` folder just name the file `Weather.qml`

end-4:
~~Hmm why do we need location again? If i just `curl wttr.in` it seems to be correct~~ right, vpn

.config/quickshell/modules/bar/Bar.qml
tekeoglan: The right section is a MouseArea, and I'm also using a MouseArea inside the WeatherBar, which causes mouse events to not work as expected. If you'd prefer, I can wrap them in a RowLayout instead.

.config/quickshell/modules/bar/weather/WeatherBar.qml
tekeoglan: Destroying and remounting the Popup wouldn’t provide any real benefit. It’s a very lightweight component with a purely visual purpose, so it doesn’t introduce any noticeable overhead.
It should also be interactive—the user shouldn’t have to wait to see the weather information every time they hover over it.

tekeoglan:
@end-4 I’ve implemented a version where WeatherBar and RightMouseArea are placed inside a RowLayout. Which approach do you prefer?
```qml
...
                // barRight
                RowLayout {
                    id: barRightSide
                    implicitHeight: Appearance.sizes.baseBarHeight
                    // implicitWidth: barRoot.width - (barLeftSideMouseArea.width + middleSection.width)
                    anchors.right: parent.right
                    anchors.left: middleSection.right
                    anchors.leftMargin: 10
                    spacing: 10

                    VerticalBarSeparator {
                        visible: Config.options.bar.borderless && Config.options.bar.weather.enable
                    }

                    Loader {
                        id: weatherLoader
                        active: Config.options.bar.weather.enable
                        // anchors.margins: 10
                        sourceComponent: BarGroup {
                            implicitWidth: weatherBar.implicitWidth
                            implicitHeight: weatherBar.implicitHeight

                            WeatherBar {
                                id: weatherBar
                            }
                        }
                    }
                    // refactor of the right mouse area
                    BarRightMouseArea {
                        useShortenedForm: barRoot.useShortenedForm
                        bar: barRoot
                        Layout.fillWidth: true
                        Layout.fillHeight: true
                    }
                }
            }
...
```

end-4:
find this in Bar.qml

```
Item {
    Layout.fillWidth: true
    Layout.fillHeight: true
}
```

put your loader below there

tekeoglan:
@end-4, I made the necessary changes.

.config/quickshell/services/Weather.qml
```diff
@@ -0,0 +1,155 @@
...
+
+    function refineData(data) {
+        let temp = {};
+        temp.uv = data?.current?.uvIndex || 0;
+        temp.humidity = (data?.current?.humidity || 0) + "%";
+        temp.sunrise = data?.astronomy?.sunrise || "0.0";
+        temp.sunset = data?.astronomy?.sunset || "0.0";
+        temp.windDir = data?.current?.winddir16Point || "N";
+        temp.wCode = data?.current?.weatherCode || "113";
+        temp.city = data?.location?.areaName[0].value || "Istanbul";
```
HigherOrderLogic: Why is this default to Istanbul?

tekeoglan: I forgot to update the default value in service. I've fixed [here](https://github.com/end-4/dots-hyprland/pull/1520/commits/8b83b42eed0070f0bb96658f188058c0e63b5d5f).

end-4:
need to disable by default and fix the font size

tekeoglan:
🎉

habibulfauzan:
does this module disable the system tray? usually I can show and minimize spotify to the top right corner.

or is it because my monitor is too small?
image

end-4:
it does not, but the system tray is known to be problematic #1510

habibulfauzan:
okay thanks sir.

may i ask what is the resolution of your monitor sir? because i see many space on the bar. 
and i want to be like that

end-4:
> okay thanks sir.
> 
> may i ask what is the resolution of your monitor sir? because i see many space on the bar. and i want to be like that

1920x1080

PR #1530, opened 2025-06-28
Update color.txt path in switchwall.sh

torhaala:
Update color.txt path in switchwall.sh -

The rest of the codebase uses "$STATE_DIR/user/generated/color.txt", apart from one line in the switchwall script which still uses "$STATE_DIR/user/color.txt".

Assume this is just a typo?


end-4:
not typo, i forgot to update it, thanks

PR #1575, opened 2025-07-05
include possible zed aliases in keybinds.conf

Flamarine:
## Describe your changes

Sometimes third-party Zed packages will rename the `zed` binary to avoid conflicts (as described [here](https://zed.dev/docs/linux#installing-via-a-package-manager)). As far as I know Arch renamed the binary to `zeditor`. This PR simply includes these aliases in `keybinds.conf`.

It's ready. (Unless if you do have more ideas on this, then yeah feedback is OK)


end-4:
thank you

Issue #1254, opened 2025-04-29
Handle deepseek reasoner model


baoliejuan:
### What would you like to be added?

Add handling for deepseek reasoner, or deepseek-r1 model

### How will it help?

Help user use deepseek-r1 model as intended

### Extra info

my setting for the model is 
```
            "deepseek-r1": {
                "name": "DeepSeek-R1",
                "logo_name": "deepseek-symbolic",
                "description": "DeepSeek",
                "base_url": "https://api.deepseek.com/chat/completions",
                "requires_key": true,
                "key_get_url": "https://platform.deepseek.com/api_keys",
                "key_file": "deepseek_key.txt",
                "model": "deepseek-reasoner"
            }

```
for example if I say 'hi' to deepseek-r1, it shows 
>nullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullHey! 👋 How can I help today?

in side panel 

when I use curl to send the api same message, it returns 
`{"id":"d1a5d268-016a-432c-a32d-50fe7836ccb7","object":"chat.completion","created":1745919264,"model":"deepseek-reasoner","choices":[{"index":0,"message":{"role":"assistant","content":"Hello! How can I assist you today?","reasoning_content":"Okay, the user just said \"hi\". I should respond in a friendly and welcoming manner. Maybe say something like \"Hello! How can I assist you today?\" to invite them to ask for help. Keep it open-ended so they feel comfortable to ask anything."},"logprobs":null,"finish_reason":"stop"}],"usage":{"prompt_tokens":12,"completion_tokens":64,"total_tokens":76,"prompt_tokens_details":{"cached_tokens":0},"completion_tokens_details":{"reasoning_tokens":53},"prompt_cache_hit_tokens":0,"prompt_cache_miss_tokens":12},"system_fingerprint":"fp_5417b77867_prod0425fp8"}`

hscissors:
Yes please!

end-4:
I think I've already encountered a similar problem when using deepseek r1 offline... I'll see if there's any null check I can add

end-4:
does that fix it?

baoliejuan:
> does that fix it?

[...]

PR #1576, opened 2025-07-05
fix gpt only shows reasoning content (#1254)

baoliejuan:
## Describe your changes
fix gpt only shows reasoning content (#1254)


end-4:
thank you

PR #1597, opened 2025-07-08
UWSM Support Branch

PhoenixAceVFX:
## Additional Branch for UWSM

I highly recommend making this a separate branch in case UWSM needs more
That and if users are not on a systemd boot then having a normal branch and a UWSM branch lets them pick instead of being forced to something

I'm actually using it as we speak so I believe it is ready and working
UWSM is recommended for systemd boot systems and it also enables alot of things autostarting already like XDG which is nice

My implementation of UWSM configuration *may* be incorrect, I have only tinkered with it a small bit and it works relatively well so far

### DIFF
--- a/.config/hypr/hyprland/keybinds.conf
+++ b/.config/hypr/hyprland/keybinds.conf
@@ -49,7 +49,7 @@ bindl = Alt ,XF86AudioMute, exec, wpctl set-mute @DEFAULT_SOURCE@ toggle # [hidd
 bindl = ,XF86AudioMicMute, exec, wpctl set-mute @DEFAULT_SOURCE@ toggle # [hidden]
 bindld = Super+Alt,M, Toggle mic, exec, wpctl set-mute @DEFAULT_SOURCE@ toggle # [hidden]
 bindd = Ctrl+Super, T, Change wallpaper, exec, ~/.config/quickshell/$qsConfig/scripts/colors/switchwall.sh # Change wallpaper
-bind = Ctrl+Super, R, exec, killall ags agsv1 gjs ydotool qs quickshell; qs -c $qsConfig & # Restart widgets
+bind = Ctrl+Super, R, exec, killall ags agsv1 gjs ydotool qs quickshell; app2unit qs -c $qsConfig # Restart widgets
 
 ##! Utilities
 # Screenshot, Record, OCR, Color picker, Clipboard history
@@ -216,6 +216,7 @@ bind = Super, X, exec, ~/.config/hypr/hyprland/scripts/launch_first_available.sh
 bind = Ctrl+Super, V, exec, ~/.config/hypr/hyprland/scripts/launch_first_available.sh "pavucontrol-qt" "pavucontrol" # Volume mixer
 bind = Super, I, exec, XDG_CURRENT_DESKTOP=gnome ~/.config/hypr/hyprland/scripts/launch_first_available.sh "qs -p ~/.config/quickshell/$qsConfig/settings.qml" "systemsettings" "gnome-control-center" "better-control" # Settings app
 bind = Ctrl+Shift, Escape, exec, ~/.config/hypr/hyprland/scripts/launch_first_available.sh "gnome-system-monitor" "plasma-systemmonitor --page-name Processes" "command -v btop && kitty -1 fish -c btop" # Task manager
+bind = Ctrl+Super, R, exec, killall ags agsv1 gjs ydotool qs quickshell; app2unit qs -c $qsConfig # Restart widgets
 
 # Cursed stuff
 ## Make window not amogus large
--- a/.config/hypr/hyprland/scripts/launch_first_available.sh
+++ b/.config/hypr/hyprland/scripts/launch_first_available.sh
@@ -2,6 +2,6 @@
 for cmd in "$@"; do
     [[ -z "$cmd" ]] && continue
     eval "command -v ${cmd%% *}" >/dev/null 2>&1 || continue
-    eval "$cmd" &
+    eval "uwsm app -- $cmd" &
     exit
 done
--- a/.config/hypr/hyprland/env.conf
+++ b/.config/hypr/hyprland/env.conf
@@ -22,6 +22,10 @@ env = XDG_MENU_PREFIX, plasma-
 
 # ######## Virtual envrionment #########
 env = ILLOGICAL_IMPULSE_VIRTUAL_ENV, ~/.local/state/quickshell/.venv
+env = qsconfig, ii
 
 # ######## Terminal application #########
 env = TERMINAL,kitty -1
+
+env = APP2UNIT_SLICES='a=app-graphical.slice b=background-graphical.slice s=session-graphical.slice'
+env = APP2UNIT_TYPE=service
--- a/.config/hypr/hyprland/execs.conf
+++ b/.config/hypr/hyprland/execs.conf
@@ -1,27 +1,28 @@
 # Bar, wallpaper
-exec-once = ~/.config/hypr/hyprland/scripts/start_geoclue_agent.sh
-exec-once = qs -c $qsConfig &
+exec-once = app2unit ~/.config/hypr/hyprland/scripts/start_geoclue_agent.sh
+exec-once = app2unit qs -c $qsConfig
 
 # Input method
-exec-once = fcitx5
+exec-once = app2unit fcitx5
 
 # Core components (authentication, lock screen, notification daemon)
-exec-once = gnome-keyring-daemon --start --components=secrets
-exec-once = /usr/lib/polkit-kde-authentication-agent-1 || /usr/libexec/polkit-kde-authentication-agent-1  || /usr/lib/polkit-gnome/polkit-gnome-authentication-agent-1 || /usr/libexec/polkit-gnome-authentication-agent-1
-exec-once = hypridle
-exec-once = dbus-update-activation-environment --all
-exec-once = sleep 1 && dbus-update-activation-environment --systemd WAYLAND_DISPLAY XDG_CURRENT_DESKTOP # Some fix idk
-exec-once = hyprpm reload
+exec-once = app2unit gnome-keyring-daemon --start --components=secrets
+
+# For polkit agent, pick the correct binary for your system, e.g.:
+exec-once = app2unit /usr/lib/polkit-gnome/polkit-gnome-authentication-agent-1
+
+exec-once = app2unit hypridle
+exec-once = app2unit dbus-update-activation-environment --all
+exec-once = app2unit bash -c "sleep 1 && dbus-update-activation-environment --systemd WAYLAND_DISPLAY XDG_CURRENT_DESKTOP"
+exec-once = app2unit hyprpm reload
 
 # Audio
-exec-once = easyeffects --gapplication-service
+exec-once = app2unit easyeffects --gapplication-service
 
 # Clipboard: history
-# exec-once = wl-paste --watch cliphist store &
-exec-once = wl-paste --type text --watch cliphist store
-exec-once = wl-paste --type image --watch cliphist store
-exec-once = wl-clip-persist --clipboard regular
+exec-once = app2unit wl-paste --type text --watch cliphist store
+exec-once = app2unit wl-paste --type image --watch cliphist store
+exec-once = app2unit wl-clip-persist --clipboard regular
 
 # Cursor
 exec-once = hyprctl setcursor Bibata-Modern-Classic 24
-


Fane447:
For one off commands like changing volume, notifications etc, it's actually recommended to not use UWSM for them. UWSM is meant to help with long running processes, and for those that are running for milliseconds it's just extra overhead.

For example
`bind = Ctrl+Super, R, exec, uwsm app -- killall ags agsv1 gjs ydotool qs quickshell; qs & # Restart widgets`
You'd want it to instead be
`bind = Ctrl+Super, R, exec, killall ags agsv1 gjs ydotool qs quickshell; uwsm app -- qs & # Restart widgets`
as the killall is a command that'll finish instantly but you'd want quickshell to start back up with uwsm.

It can be useful to modify the App Launcher though like [this](https://github.com/end-4/dots-hyprland/compare/main...Fane447:dots-hyprland:uwsm) (I'm using app2unit, so you'd just need to change that to 'uwsm app') so any program you open is using UWSM

NihonFemStyle:
ah thanks for letting me know
i didn't fully understand where it's needed and where it's not so i'll update the conf for that

NihonFemStyle:
Hopefully i addressed that correctly

end-4:
I just wanna confirm what gippity says is correct to at least have a solid reason to use UWSM (instead of the highly technical, unelaborated benefits commonly given): it should let me shut down Chrome cleanly instead of having it cry about restoring tabs right?

As for having an extra branch, I certainly _can_ have two branches, but if we're being real I'm just gonna neglect it. I have some fallbacks for non-systemd systems (somewhere in the logout screen), but I'm not afraid to break non-systemd experience if it's necessary to give systemd systems a better experience.... Actually, maybe we can provide a `non-systemd-patch.sh` to delete the all `uwsm app --`s? 

And I guess I might need to include an UWSM config later? 

Todos
- [ ] Make a release before merging UWSM support so non-systemd users can find a somewhat reliable point to go from

NihonFemStyle:
@end-4 i have a solution
since uwsm mostly requires the modified execs and keybinds (that is literally all i edited for it to work) you could have the setup ask if they want uwsm or not and then copy the set asked for

this idea i recommend because then you only need one branch and can maintain everything within for both systemd and non-systemd users so everyone can enjoy this still

the reason i put this is because sometimes using bash to patch how you want is a little finicky (i had issues with this with my screenshot script)

Fane447:
Also the keybinds that call scripts directly, for example
`bind = Super, Return, exec, uwsm app -- ~/.config/hypr/hyprland/scripts/launch_first_available.sh "kitty -1" "foot" "alacritty" "wezterm" "konsole" "kgx" "uxterm" "xterm" # Terminal`

Will also run the script in the USWM slice but not the actual terminal app. It needs the uswm part removing from the keybind and the launch_first_available script would need line 4 editing to be   `eval "uwsm app -- $cmd" &`.

NihonFemStyle:
wait really?
wow i didn't know about that

NihonFemStyle:
oh yea that does make more sense, the script doesn't use the uwsm but the application it launches would get the uwsm, thanks for pointing that out

NihonFemStyle:
that should do it @Fane447

.config/hypr/hyprland/keybinds.conf
```diff
@@ -44,26 +44,23 @@ bindl = Alt ,XF86AudioMute, exec, wpctl set-mute @DEFAULT_SOURCE@ toggle # [hidd
...
 bindld = Super+Alt,M, Toggle mic, exec, wpctl set-mute @DEFAULT_SOURCE@ toggle # [hidden]
 bindd = Ctrl+Super, T, Change wallpaper, exec, ~/.config/quickshell/scripts/colors/switchwall.sh # Change wallpaper
-bind = Ctrl+Super, R, exec, killall ags agsv1 gjs ydotool qs quickshell; qs & # Restart widgets
+bind = Ctrl+Super, R, exec, killall ags agsv1 gjs ydotool qs quickshell; uwsm app -- qs & # Restart widgets
 
 ##! Utilities
 # Screenshot, Record, OCR, Color picker, Clipboard history
 bindd = Super, V, Copy clipboard history entry, exec, qs ipc call TEST_ALIVE || pkill fuzzel || cliphist list | fuzzel --match-mode fzf --dmenu | cliphist decode | wl-copy # [hidden] Clipboard history >> clipboard (fallback)
-bindd = Super, Period, Copy an emoji, exec, qs ipc call TEST_ALIVE || pkill fuzzel || ~/.config/hypr/hyprland/scripts/fuzzel-emoji.sh copy # [hidden] Emoji >> clipboard (fallback)
+bindd = Super, Period, Copy an emoji, exec, qs ipc call TEST_ALIVE || pkill fuzzel || uwsm app -- ~/.config/hypr/hyprland/scripts/fuzzel-emoji.sh copy # [hidden] Emoji >> clipboard (fallback)
```
Fane447: UWSM not need here, it's just showing a QS window, and QS is already running through UWSM

.config/hypr/hyprland/keybinds.conf
```diff
@@ -176,10 +173,10 @@ bind = Super+Alt, f12, exec, bash -c 'RANDOM_IMAGE=$(find ~/Pictures -type f | g
...
 
 ##! Session
-bindd = Super, L, Lock, exec, loginctl lock-session # Lock
-bind = Super+Shift, L, exec, loginctl lock-session # [hidden]
-bindld = Super+Shift, L, Suspend system, exec, sleep 0.1 && systemctl suspend || loginctl suspend # Sleep
-bindd = Ctrl+Shift+Alt+Super, Delete, Shutdown, exec, systemctl poweroff || loginctl poweroff # [hidden] Power off
+bindd = Super, L, Lock, exec, uwsm app -- loginctl lock-session # Lock
+bind = Super+Shift, L, exec, uwsm app -- loginctl lock-session # [hidden]
+bindld = Super+Shift, L, Suspend system, exec, uwsm app -- sleep 0.1 && systemctl suspend || loginctl suspend # Sleep
+bindd = Ctrl+Shift+Alt+Super, Delete, Shutdown, exec, uwsm app -- systemctl poweroff || loginctl poweroff # [hidden] Power off
```
Fane447: I'd remove the UWSM here and the above lines of suspend system, and the 2 lock keybinds. All 4 are just sending a command so will complete instantly.

.config/hypr/hyprland/keybinds.conf
```diff
@@ -44,26 +44,23 @@ bindl = Alt ,XF86AudioMute, exec, wpctl set-mute @DEFAULT_SOURCE@ toggle # [hidd
...
-bindld = Ctrl,Print, Screenshot >> clipboard & save, exec, mkdir -p $(xdg-user-dir PICTURES)/Screenshots && grim $(xdg-user-dir PICTURES)/Screenshots/Screenshot_"$(date '+%Y-%m-%d_%H.%M.%S')".png # Screenshot >> clipboard & file
 # Recording stuff
-bindd = Super+Alt, R, Record region (no sound), exec, ~/.config/hypr/hyprland/scripts/record.sh # Record region (no sound)
-bindd = Ctrl+Alt, R, Record screen (no sound), exec, ~/.config/hypr/hyprland/scripts/record.sh --fullscreen # [hidden] Record screen (no sound)
-bindd = Super+Shift+Alt, R, Record screen (with sound), exec, ~/.config/hypr/hyprland/scripts/record.sh --fullscreen-sound # Record screen (with sound)
+bindd = Super+Alt, R, Record region (no sound), exec, uwsm app -- ~/.config/hypr/hyprland/scripts/record.sh # Record region (no sound)
+bindd = Ctrl+Alt, R, Record screen (no sound), exec, uwsm app -- ~/.config/hypr/hyprland/scripts/record.sh --fullscreen # [hidden] Record screen (no sound)
+bindd = Super+Shift+Alt, R, Record screen (with sound), exec, uwsm app -- ~/.config/hypr/hyprland/scripts/record.sh --fullscreen-sound # Record screen (with sound)
```
Fane447: For these recording scripts, the UWSM call would be in the script, before each wf-recorder

Fane447 (commented):
Pointed out a few more places where UWSM shouldn't really be used. Apologies if you feel I am messing you around, I didn't want to dump it all at once.

NihonFemStyle:
all good
if anything it helps

end-4:
> have the setup ask if they want uwsm or not and then copy the set asked for

Kinda similar to having 2 branches, i don't wanna have to modify 2 places at once and would prefer a lazy solution

Also, am I supposed to use the `Hyprland (uwsm-managed)` entry in my display manager? using it makes it take really long to log in and makes logout never finish (at least with SDDM)

And you have not answered my question... To me there's nothing but disadvantages at this point so you gotta convince me...

.config/hypr/hyprland/keybinds.conf
```diff
@@ -30,8 +30,8 @@ bindd = Super, Slash, Toggle cheatsheet, global, quickshell:cheatsheetToggle # T
 bindd = Super, K, Toggle on-screen keyboard, global, quickshell:oskToggle # Toggle on-screen keyboard
 bindd = Super, M, Toggle media controls, global, quickshell:mediaControlsToggle # Toggle media controls
 bindd = Ctrl+Alt, Delete, Toggle session menu, global, quickshell:sessionToggle # Toggle session menu
-bind = Ctrl+Alt, Delete, exec, qs ipc call TEST_ALIVE || pkill wlogout || wlogout -p layer-shell # [hidden] Session menu (fallback)
-bind = Shift+Super+Alt, Slash, exec, qs -p ~/.config/quickshell/welcome.qml # [hidden] Launch welcome app
+bind = Ctrl+Alt, Delete, exec, uwsm app -- wlogout -p layer-shell # [hidden] Session menu (fallback)
+bind = Shift+Super+Alt, Slash, exec, uwsm app -- qs -p ~/.config/quickshell/welcome.qml # [hidden] Launch welcome app
```
end-4: don't remove `qs ipc call TEST_ALIVE`, wlogout is only a fallback when qs is not running

well, i guess this could be pidof as well but we would have to check both `qs` and `quickshell` binary names so idk if this or an ipc call is better

Fane447:
> > have the setup ask if they want uwsm or not and then copy the set asked for
> 
> Kinda similar to having 2 branches, i don't wanna have to modify 2 places at once and would prefer a lazy solution
> 
> Also, am I supposed to use the `Hyprland (uwsm-managed)` entry in my display manager? using it makes it take really long to log in and makes logout never finish (at least with SDDM)
> 
> And you have not answered my question... To me there's nothing but disadvantages at this point so you gotta convince me...

I personally don't think it's worth the headache currently. Yeah it's meant to help with the Chrome tab issue you mentioned but apart from that I don't think it's anything that would be noticeable to the user, and it can cause issues like you found.
Yes it would be the Hyprland (uwsm-managed) option. Login should be the same amount of time, but one difference is that UWSM will autostart your .desktop entries such as in .config/autostart so it's quite likely it's opening a lot more programs than normal.
The logout needs to be done differently, there's a few different ways listed on the Github page, one is using `uwsm stop` but this has an ongoing issue open https://github.com/Vladimir-csp/uwsm/issues/92 where for some people this just causes a blank screen or freezing.

NihonFemStyle:
i do not know the full extent to the benefits for UWSM but i have a list here that i've put together based on my own and some other's experiences with using it

# Session Management Benefits:
- Proper session lifecycle management - UWSM handles session startup, shutdown, and cleanup more reliably than traditional methods
- Better environment variable handling - Ensures proper propagation of environment variables to all applications in the session
- Improved service integration - Better integration with systemd user services and session management
- Clean session termination - Handles graceful shutdown of Wayland compositors and associated services

# Hyprland-Specific Advantages:
- Enhanced stability - UWSM provides more robust session handling, reducing crashes and unexpected behavior
- Better application launching - Improved application startup through proper session context
- Resource management - More efficient handling of system resources and cleanup on session end
- Integration with system services - Better coordination with background services and daemons

# Development and Maintenance Benefits:
- Standardized approach - UWSM provides a consistent way to manage Wayland sessions across different compositors
- Reduced configuration complexity - Simplifies the setup process compared to manual session management
- Better debugging - Improved logging and error handling for session-related issues
- Future-proofing - Designed to work with evolving Wayland ecosystem standards

# User Experience Improvements:
- Faster session startup - Optimized session initialization process
- More reliable desktop environment - Fewer session-related glitches and inconsistencies
- Better multi-monitor support - Improved handling of display configuration during session management
- Enhanced security - Better isolation and permission handling for session components

NihonFemStyle:
the list was made with a pool of about me and a group of about 10 people all experimenting with it for some time
I personally use UWSM because it has been more stable for me than normal launch

in the end, my opinion is to each their own

NihonFemStyle:
> > have the setup ask if they want uwsm or not and then copy the set asked for
> 
> Kinda similar to having 2 branches, i don't wanna have to modify 2 places at once and would prefer a lazy solution
> 
> Also, am I supposed to use the `Hyprland (uwsm-managed)` entry in my display manager? using it makes it take really long to log in and makes logout never finish (at least with SDDM)
> 
> And you have not answered my question... To me there's nothing but disadvantages at this point so you gotta convince me...

i'm surprised you're taking awhile to login, i haven't had that specific issue yet but the logout issue i have had *sometimes* (no clue if it's SDDM specific, i just have command scripts for each WM/DE boot i want as i dont use SDDM or any display manager at that)

Vladimir-csp:
> `bind = Super, Return, exec, uwsm app -- ~/.config/hypr/hyprland/scripts/launch_first_available.sh "kitty -1" "foot" "alacritty" "wezterm" "konsole" "kgx" "uxterm" "xterm" # Terminal`

Just `uwsm app -T`. [Default terminal Spec proposal](https://gitlab.freedesktop.org/terminal-wg/specifications/-/merge_requests/3): 

Also check out [app2unit](https://github.com/Vladimir-csp/app2unit/), a faster alternative to `uwsm app`.

xBiei:
I agree with using uwsm as the main branch without having 2 branches at once, I don't think it's worth the hassle of making 2 branches since non-systemd users are a small minority compared to the other side. But I think making a patch file / a script would be a good idea.
UWSM has its advantages and disadvantages but I honestly think that the advantages are worth it, especially for session management which was annoying without it.

NihonFemStyle:
i could likely just make a fast bash script to replace normal cfg with uwsm cfg or vice versa, depending on which becomes the main

also HOLY CRAP I FORGOT THIS PR EXISTED
Don't mind the various changes i did, i kinda went on the road of maintaining a UWSM fork using app2unit with uwsm completely forgetting this PR was here

NihonFemStyle:
oh oops it closed it when i went to rebase lol

NihonFemStyle:
i rebased it and took another try at using uwsm and app2unit properly
hopefully this is alot cleaner

.config/hypr/hyprland/scripts/launch_first_available.sh
sansmoraxz, line 1: Please don't add here, its just an util.

Would be adding too much of an overhead.

.config/hypr/hyprland/keybinds.conf
```diff
@@ -49,7 +49,7 @@ bindl = Alt ,XF86AudioMute, exec, wpctl set-mute @DEFAULT_SOURCE@ toggle # [hidd
 bindl = ,XF86AudioMicMute, exec, wpctl set-mute @DEFAULT_SOURCE@ toggle # [hidden]
 bindld = Super+Alt,M, Toggle mic, exec, wpctl set-mute @DEFAULT_SOURCE@ toggle # [hidden]
 bindd = Ctrl+Super, T, Change wallpaper, exec, ~/.config/quickshell/$qsConfig/scripts/colors/switchwall.sh # Change wallpaper
-bind = Ctrl+Super, R, exec, killall ags agsv1 gjs ydotool qs quickshell; qs -c $qsConfig & # Restart widgets
+bind = Ctrl+Super, R, exec, killall ags agsv1 gjs ydotool qs quickshell; app2unit qs -c $qsConfig # Restart widgets
```
sansmoraxz, line 52: Also not a good idead here. Better use the daemon mode.

NihonFemStyle:
ok i thought the resolve button would open an editor on mobile

.config/hypr/hyprland/keybinds.conf
NihonFemStyle: edit it freely, im still learning how to use uwsm and app2unit

PR #1599, opened 2025-07-08
Constrains Background Widget to Stop it from Clipping Outside of Screen

0iSuck-at-coding0:
I have a 3440x1440 resolution monitor. I noticed that the widget that shows the clock, clips out of my screen to varying degrees. These are my changes in a nutshell: 

- I kept the original logic to calculate the desired position of the widget (which is the `unconstrainedLeftMargin` and `unconstrainedTopMargin`)

- However, I applied a bounds check with `constrainedLeftMargin` and `constrainedTopMargin`

- The boundary constraints are `Math.max(0, ...)` (ensures widget doesn't go off the left/top edge), `Math.min(..., windowRoot.width - implicitWidth)` (ensures the widget doesn't go off the right edge and `Math.min(..., windowRoot.height - implicitHeight)` (ensure the widget doesn't go off the bottom edge.

- I updated the anchors to use my new constraints 

This fixed the background widget clipping outside of my screen on my (admittedly) edge case of a monitor resolution. I tested this on my laptop with a 1920x1080 screen and the widget seems to be working normally there, so it might work for everyone. However, testing on other resolutions and refinements to this code is always welcome.

### DIFF
--- a/.config/quickshell/modules/backgroundWidgets/BackgroundWidgets.qml
+++ b/.config/quickshell/modules/backgroundWidgets/BackgroundWidgets.qml
@@ -92,11 +92,20 @@ Scope {
                     color: root.colBackground
                     implicitHeight: columnLayout.implicitHeight + verticalPadding * 2
                     implicitWidth: columnLayout.implicitWidth + horizontalPadding * 2
+                    
+                    // Calculate constrained position to keep widget within screen bounds
+                    property real unconstrainedLeftMargin: (root.effectiveCenterX / monitor.scale - implicitWidth / 2)
+                    property real unconstrainedTopMargin: (root.effectiveCenterY / monitor.scale - implicitHeight / 2)
+                    
+                    // Constrain to screen boundaries
+                    property real constrainedLeftMargin: Math.max(0, Math.min(unconstrainedLeftMargin, windowRoot.width - implicitWidth))
+                    property real constrainedTopMargin: Math.max(0, Math.min(unconstrainedTopMargin, windowRoot.height - implicitHeight))
+                    
                     anchors {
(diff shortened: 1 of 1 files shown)


.config/quickshell/modules/backgroundWidgets/BackgroundWidgets.qml
```diff
@@ -92,11 +92,20 @@ Scope {
...
                     implicitHeight: columnLayout.implicitHeight + verticalPadding * 2
                     implicitWidth: columnLayout.implicitWidth + horizontalPadding * 2
+                    
+                    // Calculate constrained position to keep widget within screen bounds
+                    property real unconstrainedLeftMargin: (root.effectiveCenterX / monitor.scale - implicitWidth / 2)
+                    property real unconstrainedTopMargin: (root.effectiveCenterY / monitor.scale - implicitHeight / 2)
+                    
+                    // Constrain to screen boundaries
+                    property real constrainedLeftMargin: Math.max(0, Math.min(unconstrainedLeftMargin, windowRoot.width - implicitWidth))
+                    property real constrainedTopMargin: Math.max(0, Math.min(unconstrainedTopMargin, windowRoot.height - implicitHeight))
```
end-4, line 102: use `- width` instead of `- implicitWidth`, and same for height

0iSuck-at-coding0:
I just updated and I see what you're saying. The extra padding does fix this issue on my setup and this PR isn't necessary anymore. Also thanks for the tip, using `- width` and `- height` makes a lot more sense when doing constraints. I'll go ahead and close this now.

P.S. No problem for replying late, I'm not exactly timely myself.

PR #1613, opened 2025-07-11
feat: add option to ignore conflicting files

obsidrielle:
## Describe your changes

This change enhances the `update.sh` script to better handle conflicts with user-modified configuration files.

When a file conflict occurs during an update, it's often because the user has intentionally made personal modifications that they want to keep. When prompted to resolve a conflict, the user can now select "7) Add to ignore and skip". This action appends the file's path to the update ignore list, preventing it from being overwritten in all future updates.

Ready for review. Feedback is welcome


end-4:
hey @clsty can you review if you have time?

clsty:
LGTM, thanks for contribution.

PR #1615, opened 2025-07-11
Added seekable functionality to media player

adityaprakashgupta:
### Summary
This PR adds seekable playback functionality to the media player component. Users can now interact with a timeline to jump to different parts of the media.

### Features Implemented
- Mouse controls for seeking
- Backend support for updating playback position
- Sync between playback state and UI updates

### Testing
- Manually tested with audio, video files and music streaming services
- Verified seek accuracy

### Notes
- No breaking changes introduced

### DIFF
--- a/.config/quickshell/ii/modules/common/widgets/StyledProgressBar.qml
+++ b/.config/quickshell/ii/modules/common/widgets/StyledProgressBar.qml
@@ -9,8 +9,8 @@ import Quickshell.Widgets
 import Qt5Compat.GraphicalEffects
 
 /**
- * Material 3 progress bar. See https://m3.material.io/components/progress-indicators/overview
- */
+* Material 3 progress bar. See https://m3.material.io/components/progress-indicators/overview
+*/
 ProgressBar {
     id: root
     property real valueBarWidth: 120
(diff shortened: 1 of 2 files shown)


end-4:
a seekable progress bar is called a slider
don't extend the progress bar like that

i also don't really like seeking because in many cases it doesn't work...

adityaprakashgupta:
> a seekable progress bar is called a slider don't extend the progress bar like that
> 
> i also don't really like seeking because in many cases it doesn't work...

I'd like to introduce a **separate implementation** by creating a new file that extends the progress bar into a **seekable slider**.

To keep things flexible and non-intrusive:

- Add an optional config flag: `enable_seek_slider`
- Default behavior remains unchanged (non-seekable)
- Users who want slider/seek functionality can enable it explicitly

This way:

- The core experience is preserved for users who don’t want or trust seeking
- Those who prefer or require seek support can opt in without affecting others

Let me know if you're okay with this approach. I’ll keep it clean, modular, and easy to maintain.

end-4:
No, a slider should be implemented with a real Slider provided by QtQuick. You can fix the current positioning and add more customization to the StyledSlider but don't extend StyledProgressBar to do what a slider does

end-4:
In a5831cf3657dc91e92cf6a9fbf48858343485b1d I made the slider more accurate, but haven't implemented the wavy effect yet

adityaprakashgupta:
DONE: add optional seekable media player using StyledSlider

- Extended StyledSlider to support seek functionality in StyledMediaSlider
- Added toggle option in settings panel to enable/disable seekable player
- Updated config structure to handle the new option

end-4:
- The slider handle should be slightly bigger
- Copying StyledSlider into a new StyledMediaSlider is not the way. A `sperm` option should be introduced in `StyledSlider`

PR #1616, opened 2025-07-11
Pr

mrcxlinux:
in qt6ct config and kde-material-you-colors there were two paths that had "/home/end" which ive changed to ~

Theres something weird, I saw that kde-material-you-colors was checking for a "wallpaper.txt" which did not exist, i assume that the correct path would be " ~/.local/state/quickshell/user/generated/wallpaper/path.txt" but kde-material-you-colors says that the file does not exist for some reason

please correct me if im wrong in any way, I just thought that I would reach out
I can make a separate PR for the qt6ct path if wanted!

best regards,
mrcxlinux

### DIFF
--- a/.config/kde-material-you-colors/config.conf
+++ b/.config/kde-material-you-colors/config.conf
@@ -10,7 +10,7 @@ monitor = 0
 
 # File containing absolute path of an image (Takes precedence over automatic wallpaper detection)
 # Commented by default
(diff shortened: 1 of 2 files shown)


end-4:
aaaaaaaaaa silly me

edit: turns out it doesn't matter because i always feed kde-material-you-color a color in the wrapper script

PR #1618, opened 2025-07-11
Round corner transparency of hug style bar

Sighthesia:
## Describe your changes
This make the round corners of hug style bar have the same transparency with bar, uhh it just a slightly different...
I remerber that it was me introduce this trick before...It seems that it was caused by bug, now the bar.qml is refactored, it is ok to remove it.
Thanks a lot!

I think it's ready


end-4:
right, because now the corner does redraw when the color changes. thanks

PR #1620, opened 2025-07-11
Todo list add editing

PimRijkers:
## Describe your changes
Adds a button to the todo list items, so they can be edited. Works for done items as well.
Shows a fade-in animation when edited, so it's clear which item was edited.
Has the same dialog as 'Add Task', but is called 'Edit Task'. The 'Edit' button is only enabled when the text is edited.

editable_task<br/>
edit_dialog<br/>
edit_dialog_button_enabled

1. Perhaps the 'Add Task' action could have the same fade-in animation to create a more consistent appearance for all actions.
2. I'll add another PR soon which allows for todo list ordering. They'll have overlapping code. Might be easier to merge them at the same time.

### DIFF
--- a/.config/quickshell/modules/sidebarRight/todo/EditingCallbackInfo.qml
+++ b/.config/quickshell/modules/sidebarRight/todo/EditingCallbackInfo.qml
@@ -1,8 +1,7 @@
 import QtQuick 2.15
 
 Item {
-    property int totalIndex // The index for both lists together (as stored in file)
-    property int listIndex // The index in its own unfinished/done list
+    property int index
     property bool done
     property string currentText
 }
--- a/.config/quickshell/modules/sidebarRight/todo/EditingCallbackInfo.qml
+++ b/.config/quickshell/modules/sidebarRight/todo/EditingCallbackInfo.qml
@@ -1,7 +1,0 @@
-import QtQuick 2.15
-
-Item {
-    property int index
-    property bool done
-    property string currentText
-}
--- a/.config/quickshell/modules/sidebarRight/todo/EditingCallbackInfo.qml
+++ b/.config/quickshell/modules/sidebarRight/todo/EditingCallbackInfo.qml
@@ -1,0 +1,8 @@
+import QtQuick 2.15
+
+Item {
+    property int totalIndex // The index for both lists together (as stored in file)
+    property int listIndex // The index in its own unfinished/done list
+    property bool done
+    property string currentText
+}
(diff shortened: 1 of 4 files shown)


end-4 (changes requested):
- I see Todo.editTask() but there is no such function?
- Same thing with the `editingCallback` property of `TaskList`
- Don't attach "originalIndex" to everything, just have a `currentTodoItem` variable that's updated when you click an item
- Also please rename `showAddDialog` to `showDialog` and rename `showEditDialog` to `dialogIsEdit` (change their functionality accordingly of course)

end-4:
my bad i didn't review the full changes, but the 3rd and 4th point still hold

.config/quickshell/ii/services/Todo.qml
```diff
@@ -31,6 +31,17 @@ Singleton {
...
+        const item = {
+            "content": new_desc,
+            "done": done,
+        }
+        root.list[index] = item
+        // Reassign to trigger onListChanged
+        root.list = list.slice(0)
+        todoFileView.setText(JSON.stringify(root.list))
+    }
+
```
end-4: why not just `root.list[index].content = newDesc`
you won't even need to pass `done` and you don't need to construct a new item

end-4:
I really don't like the new animation thing. This was made in the early stages when I didn't know about [ListView](https://doc.qt.io/qt-6/qml-qtquick-listview.html) which should make animations much simpler. I'd rather rewrite the list with that so can you not include the fade in thing?

.config/quickshell/ii/services/Todo.qml
PimRijkers: Good call. I don't have a background in javascript, so I didn't know const could be edited.

PimRijkers:
I've updated all the points, except the 'Don't attach "originalIndex" to everything, just have a currentTodoItem variable that's updated when you click an item' point.

PimRijkers:
I've added a currentTodoItemData to the Todo file instead of the TaskList file, so the TodoWidget doesn't need to know which TaskList is being updated.
The modelData of the TodoItem is used instead of a TodoItem, because the TodoItem won't give direct access to its data, so it will require boilerplate code to access the data. Updating the TodoItem directly by calling methods on it isn't possible either, so there wouldn't have been any benefit to using it over the modelData.

Now that Todo.editTask doesn't require an index anymore. Should the Todo.markDone, Todo.markUnfinished and Todo.deleteItem stop using the index as parameter as well to keep the code consistent? The same goes for the new Todo methods in #1621

PR #1648, opened 2025-07-15
hyprland: add window rules for xdg-portal file dialogs

nahilrasheed:
## Describe your changes
Added new windowrulev2 entries to float and center file picker dialogs with titles like "wants to save" and "wants to open".

eg; drive.google.com wants to save
       drive.google.com wants to open

Yes, it's ready.


end-4:
uhhh sure
what browser is that though? i don't get tiled dialogs for both firefox and chrome

nahilr:
brave

PR #1653, opened 2025-07-15
feat: power-profile switcher in topbar

Ninjdai1:
## Describe your changes

Adds a power profile toggle to the topbar, using PowerProfiles from quickshell and power-profiles-daemon
Disabled by default

I didn't encounter any bug while testing and it's a pretty simple feature.
I don't really like the js `switch`es in the click event, but as it's a 3-options toggle I didn't really see a better way to implement it, please lmk if I should change that.

### DIFF
--- a/.config/quickshell/ii/modules/bar/UtilButtons.qml
+++ b/.config/quickshell/ii/modules/bar/UtilButtons.qml
@@ -104,5 +104,38 @@ Item {
                 }
             }
         }
+
+        Loader {
+            active: Config.options.bar.utilButtons.showPerfProfileToggle
+            visible: Config.options.bar.utilButtons.showPerfProfileToggle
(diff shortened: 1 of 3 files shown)


mooncoffee1:
is still merged? how can we activate it? let us know pls

end-4:
uhh thanks

mooncoffee1:
just a question, the icons are not soo intuitive for the power profiles, how we can edit, to add our icons?

end-4:
edit that UtilButtons file
i don't think adding a config option for that is a good idea because there would be no end to making such options if i made one

Radrahil:
I don't seem to see any change in the icon when I click it. What packages do I need to install for this?

end-4:
power profiles daemon

PR #1656, opened 2025-07-16
Fix monitor offset w/ left-of-center monitors in overview

1cebp:
## Describe your changes
Fixes the overview window having incorrect monitor offsets when a monitor is left-of-center. This manifests as the window positions not being aligned with the correct workspace on that monitor.

Tested on a triple monitor setup and with dual/single monitors.


end-4:
thank you!

PR #1659, opened 2025-07-17
Fix the theme color of the settings page; Update translations

xiaomeng2004:
Fix the theme color settings not working (sorry, I mistakenly deleted the import "root:/services")
Add partial translation


end-4:
Thanks

PR #1660, opened 2025-07-17
Add NVIDIA glitching + new browser

Nouvborne:
Added NVIDIA in env and new browser in keybinds (useless fork but might help a bit)


Nouvborne:
I am too sleepy to modify installer to check for nvidia card maybe later

.config/hypr/hyprland/env.conf
```diff
@@ -20,5 +20,9 @@ env = QT_QPA_PLATFORMTHEME, kde
 # ######## Virtual envrionment #########
 env = ILLOGICAL_IMPULSE_VIRTUAL_ENV, ~/.local/state/quickshell/.venv
 
-# ############ Others #############
+# ######## Uncomment if you use NVIDIA GPU #########
+#env = LIBVA_DRIVER_NAME,nvidia
+#env = __GLX_VENDOR_LIBRARY_NAME,nvidia
+#env = ELECTRON_OZONE_PLATFORM_HINT,autos
```
end-4: this electron one is beneficial to also non nvidia users, works for gh desktop for me, so uncomment it

end-4:
actually
we shouldn't put something to uncomment in the default config file because the user can add overrides/more stuff in the custom folder if they want their system to be update friendly
so i'm just gonna take the electron and librewolf ones

PR #1661, opened 2025-07-17
fix random konachan script not updating wallpaper

77Dignity:
## Describe your changes
There is currently an issue with the random konachan button not updating the wallpaper consistently due to the filename not changing. To fix this a counter has been appended to the filename, resulting in subsequent wallpapers being stored permanently.

Tested for around 30 flawless consecutive generations.
If the deletion of the old files is a desired effect, this could be added as well.


end-4:
This does work but let's make it A-B (by reading the currently used wallpaper in `~/.config/illogical-impulse/config.json`) instead of going up infinitely...

77Dignity:
> This does work but let's make it A-B (by reading the currently used wallpaper in `~/.config/illogical-impulse/config.json`) instead of going up infinitely...

Sure. It now switches between the filenames in a A-B pattern if required instead.
Tested for several generations as well.

end-4:
Thank you!

PR #1673, opened 2025-07-19
Properly fix open with menu in KDE

moetayuko:
## Describe your changes
* Use XDG menu scheme from KDE, which is already implicitly introduced by illogical-impulse-kde.
* Set XDG_MENU_PREFIX globally, so kbuildsycoca6 is automatically invoked by /usr/bin/xdg-mime via the pacman hook. Consequently, the manual call can (and should) be removed.

ready


end-4:
funny, i found that earlier solution in an arch linux thread that also contained your solution and didn't read until the end for this better one
thank you

PR #1681, opened 2025-07-20
feat: add option to ignore certain apps in dock

alyingfish:
same as #1303 but for quickshell.

> I have scratchpads and i didn't want them to appear in the dock so i added this
> 
> For example i put
> 
> ```
> "ignoredAppsRegex": ["^kitty-dropterm-.$"],
> ```
> so it will ignore all app clases that match kitty-dropterm-.

### DIFF
--- a/.config/quickshell/ii/modules/dock/DockApps.qml
+++ b/.config/quickshell/ii/modules/dock/DockApps.qml
@@ -59,8 +59,8 @@ Item {
                 }
 
                 // Ignored apps
-                const ignoredAppRegexes = Config.options?.dock.ignoredAppRegexes ?? [];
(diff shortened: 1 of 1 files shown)


.config/quickshell/ii/modules/common/Config.qml
```diff
@@ -160,6 +160,7 @@ Singleton {
                 property bool hoverToReveal: true // When false, only reveals on empty workspace
                 property list<string> pinnedApps: [ // IDs of pinned entries
                     "org.kde.dolphin", "kitty",]
+                property list<string> ignoredAppsRegex: []
```
end-4: it's a list so call it ignoredAppsRegex<ins>es</ins>

PR #1683, opened 2025-07-20
Do not timeout notification if expiration is 0

Alvin-HZ:
## Describe your changes
Makes notifications not timeout if the expiration timeout is 0
Solves https://github.com/end-4/dots-hyprland/issues/1682

Yes


end-4:
so that's why some notifications don't show up at all... thanks

PR #1695, opened 2025-07-22
fix(ai): add the full received message to rawContent (OpenAi format)

Silzinc:
The messages were not preserved and passed to further calls outside of the reasoning part.

## Describe your changes

Append the newContent string to rawContent, just like it is appened to the regular content.

Pretty much I think.


end-4:
thank you

PR #1706, opened 2025-07-23
Add nvidia fixes + splitting

Nouvborne:
idk what to say, it just works tho it looks crappy. feat: weed.


Nouvborne:
task manager doesnt work by default btw, either need to install plasma monitor separately or change it to btop. ill try to work on it later when i sober up.

clsty:
- We're not handling Nvidia things but leave it to the user instead.
- The install script should be idempotent. Such bash code:
  ```bash
  if lsmod | grep -q '^nvidia'; then
  {
    echo "env = LIBVA_DRIVER_NAME,nvidia"
    echo "env = __GLX_VENDOR_LIBRARY_NAME,nvidia"
  } >> ~/.config/hypr/hyprland/env.conf
  fi
  ```
  will produce multiple snippets inside `~/.config/hypr/hyprland/env.conf` when running multiple times, which should be avoided.

Closing.

PR #1718, opened 2025-07-25
Feature: Hyprlock layout indicator

Saint2999:
## Describe your changes

What’s changed:

- Layout indicator in hyprlock.

Result:
Working layout indicator in hyprlock.

I think it's ready.


Vercixx:
I think it would be better if we move the layout indicator above the clock, or to the corner, and enable fade_on_empty back on

PR #1751, opened 2025-08-01
Settings: Navigation Design Improvment

nil0ft:
## Describe your changes

While testing the current QuickShell settings UI, I felt the navigation area could benefit from a clearer structure.  
This PR focuses on improving the visual clarity and scalability of the navigation area by introducing dividers and adjusting layout spacing.  
The intention is to make the navigation feel more structured, especially if settings are extended with more specific categories in the future.

### Before vs After

| Current Design (Before) | Updated Design (After) |
|-------------------------|------------------------|
| Current Settings | New Settings |

Not yet — I've reverted the accidental deletion of the navigation rail button.  
I'm currently reconsidering the approach to better align with Material 3 guidelines and improve how menu items visually integrate into the layout.  

Feedback and suggestions are appreciated!

### DIFF
--- a/.config/quickshell/ii/modules/common/widgets/NavigationRailButton.qml
+++ b/.config/quickshell/ii/modules/common/widgets/NavigationRailButton.qml
@@ -1,0 +1,148 @@
+import qs.modules.common
+import qs.modules.common.widgets
+import qs.modules.common.functions
+import QtQuick
+import QtQuick.Controls
+import QtQuick.Layouts
+
+TabButton {
+    id: root
+
+    property bool toggled: TabBar.tabBar.currentIndex === TabBar.index
+    property string buttonIcon
+    property string buttonText
+    property bool expanded: false
+    property bool showToggledHighlight: true
+    readonly property real visualWidth: root.expanded ? root.baseSize + 20 + itemText.implicitWidth : root.baseSize
+
+    property real baseSize: 56
+    property real baseHighlightHeight: 32
+    property real highlightCollapsedTopMargin: 8
+    padding: 0
+
+    // The navigation item’s target area always spans the full width of the
+    // nav rail, even if the item container hugs its contents.
+    Layout.fillWidth: true
+    // implicitWidth: contentItem.implicitWidth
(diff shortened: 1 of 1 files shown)


nil0ft:
image

So i reverted the deletion and will now try to improve the Design even more but also align with the Material 3 Guidelines.

end-4:
But what is wrong with that sidebar
You can try to come up with a better separation of the sections, but I don't see why it needs a redesign

nil0ft:
Compared to how polished the rest of the dots setup looks, the settings navigation feels a bit cheap and unpolished.  
It comes across more like plain Material 3 buttons stacked together than an actual settings sidebar like in GNOME Settings.  
I just think there’s some potential to make it look a bit cleaner — or what do you think?

end-4:
I find that sidebar to be the most polished thing I've made 🤔
Like, it has the most animations.... And simplicity does not mean low effort

If that design is "cheap" then your remake is "free"... The spacing is inconsistent and the sidebar is no longer collapsible

nil0ft:
I think there's a misunderstanding — I was specifically referring to the sidebar **inside the settings app**, not the global sidebar:

image

end-4:
There was no misunderstanding

nil0ft:
Alright, fair point — simplicity doesn’t mean low effort, I totally agree with that.

I just felt the collapse feature didn’t add much value, and removing it made things simpler and — in my opinion — better.  
What I mainly changed was how the menu items are presented vertically to better match the overall vibe of the dots.

Anyway, just throwing ideas around. I appreciate your perspective on this!

end-4:
I'm going to note that I didn't come up with the design, I applied what works (nicely in my opinion) from Material 3 docs: https://m3.material.io/components/navigation-rail/overview

nil0ft:
Ok, I see — after checking the docs again, it does seem like the NavigationRail is indeed the best option according to Material 3 guidelines.

We can close this PR if you prefer, since maintaining consistency with Material 3 makes sense, and the dots should stay aligned with the official guidelines anyway.

PR #1753, opened 2025-08-01
Overrideable default terminal app

sansmoraxz:
## Describe your changes

Allow us to override the terminal app. Some of us really don't like kitty.

To use, set the env var `TERMINAL` to terminal of your choice, either in hyprland `env.conf` or via whatever script runs before your desktop is initialized.

Yes

### DIFF
--- a/.config/hypr/hyprland/env.conf
+++ b/.config/hypr/hyprland/env.conf
@@ -24,4 +24,4 @@ env = XDG_MENU_PREFIX, plasma-
 env = ILLOGICAL_IMPULSE_VIRTUAL_ENV, ~/.local/state/quickshell/.venv
 
 # ######## Terminal application #########
-env = TERMINAL, "kitty -1"
+env = TERMINAL,kitty -1
--- a/.config/hypr/hyprland/env.conf
+++ b/.config/hypr/hyprland/env.conf
@@ -22,3 +22,6 @@ env = XDG_MENU_PREFIX, plasma-
 
 # ######## Virtual envrionment #########
 env = ILLOGICAL_IMPULSE_VIRTUAL_ENV, ~/.local/state/quickshell/.venv
+
+# ######## Terminal application #########
+env = TERMINAL, "kitty -1"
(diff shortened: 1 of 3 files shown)


end-4:
The launch first available script needs to be fixed for the case TERMINAL is empty

sansmoraxz:
> The launch first available script needs to be fixed for the case TERMINAL is empty

Well I am addressing it to ensure the env var is configured in `.config/hypr/hyprland/env.conf`. But yeah sure.

end-4:
thanks

sansmoraxz:
Adding one more note here

I have observed some systems without uwsm where the substitution behaviour of bindings is entirely random.

So best to use `uwsm` if overriding though env vars.

end-4:
Random? Maybe on those systems something else also tries to set that env var?

I am still not convinced to use `uwsm`

yasink21:
I can't make this env variable work in the custom/env.conf file 

"env = TERMINAL, footclient" doesn't work for me, I don't uwsm or anything, everything else is running perfectly normal, my workaround is commenting out the terminal binds in the keybinds.conf and overriding it on custom keybinds with my own, I am okay with this so far, but it's not optimal

sansmoraxz:
Check what do you get with:

```
hyprctl binds | grep launch_first_available.sh
```

sansmoraxz:
OK I figured out what was happening.

It was not random. The escaping was not proper. Sending fresh patch.

yasink21:
> Check what do you get with:
> 
> ```
> hyprctl binds | grep launch_first_available.sh
> ```

I did as you asked and the result kind of confused me;

	arg: ~/.config/hypr/hyprland/scripts/launch_first_available.sh "linuxINAL" "footclient"
	arg: ~/.config/hypr/hyprland/scripts/launch_first_available.sh  "linuxINAL" "footclient"
	arg: ~/.config/hypr/hyprland/scripts/launch_first_available.sh "linuxINAL" "footclient"

sansmoraxz:
Yeah looks like TERM is being substituted instead of TERMINAL.

Try above linked patch.

PR #1761, opened 2025-08-02
Added pomodoro timer in sidebarRight, closes #1477

### LINKED ISSUE #1477: Adding Pomodoro

I want to add Pomodoro, I know it's useless but it's a simple cool thing to add like calendar and todo list.

### How will it help?

It will be a cool way to run it instead of setup the alarm every time I want to use this technique

### Extra info

maybe make it loop with notification "It's break time" and "Back to work" or something

nyx-4:
## Describe your changes
I have added a Pomodoro in right sidebar below To Do list as was requested in #1477 by @BeshoyEhab. The stopwatch was also added because its trivial to make from a Pomodoro and I wanted maximum compatibility between To Do and Timer. It sends notification using `notify-send` when the Pomodoro status is changed.

## A quick peek
image

1. The Pomodoro didn't updated when right sidebar was not opened. It was fixed after [Commit 596ae72](https://github.com/end-4/dots-hyprland/commit/596ae72942561457c902542e2556578ec7bf668e) by @end-4 
2. I will add functionality like `laps` in Stopwatch only if requested as I personally don't have use for it.
3. The `Focus duration` and `break duration` fall back to 25 minutes and 5 minutes after reloading. I don't know where to save these two settings for reloading after restart.

other than these two points its almost ready. I am always open to Questions & feedback.

### DIFF
--- a/.config/quickshell/ii/modules/sidebarRight/pomodoro/PomodoroWidget.qml
+++ b/.config/quickshell/ii/modules/sidebarRight/pomodoro/PomodoroWidget.qml
@@ -1,0 +1,356 @@
+import qs
+import qs.services
+import qs.modules.common
+import qs.modules.common.widgets
+import QtQuick
+import QtQuick.Controls
+import QtQuick.Layouts
+
+Item {
+    id: root
+    property int currentTab: 0
+    property var tabButtonList: [
+        {"name": Translation.tr("Pomodoro"), "icon": "timer_play"},
+        {"name": Translation.tr("Stopwatch"), "icon": "timer"}
+    ]
+    property bool showDialog: false
+    property int dialogMargins: 20
+    property int fabSize: 48
+    property int fabMargins: 14
+
+
+    // These are keybinds, make sure to change them.
+    Keys.onPressed: (event) => {
+        if ((event.key === Qt.Key_PageDown || event.key === Qt.Key_PageUp) && event.modifiers === Qt.NoModifier) {
+            if (event.key === Qt.Key_PageDown) {
+                currentTab = Math.min(currentTab + 1, root.tabButtonList.length - 1)
+            } else if (event.key === Qt.Key_PageUp) {
+                currentTab = Math.max(currentTab - 1, 0)
+            }
+            event.accepted = true
+        } else if (event.key === Qt.Key_Space && !showDialog) {
+            // Toggle start/pause with Space key
+            if (currentTab === 0) {
+                Pomodoro.togglePomodoro()
+            } else {
+                Pomodoro.toggleStopwatch()
+            }
+            event.accepted = true
+        } else if (event.key === Qt.Key_R && !showDialog) {
+            // Reset with R key
+            if (currentTab === 0) {
+                Pomodoro.pomodoroReset()
+            } else {
+                Pomodoro.stopwatchReset()
+            }
+            event.accepted = true
+        } else if (event.key === Qt.Key_Escape && showDialog) {
+            showDialog = false
+            event.accepted = true
+        }
+    }
+
+    Timer {
+        id: pomodoroTimer
+        interval: 1000
+        running: Pomodoro.isPomodoroRunning
+        repeat: true
+        onTriggered: Pomodoro.tickSecond()
+    }
+
+    Timer {
+        id: stopwatchTimer
+        interval: 1000
+        running: Pomodoro.isStopwatchRunning
+        repeat: true
+        onTriggered: {
+            Pomodoro.stopwatchTime += 1
+        }
+    }
+
+
+    ColumnLayout {
+        anchors.fill: parent
+        spacing: 0
+
+        TabBar {
+            id: tabBar
+            Layout.fillWidth: true
+            currentIndex: currentTab
+            onCurrentIndexChanged: currentTab = currentIndex
+
+            background: Item {
+                WheelHandler {
+                    onWheel: (event) => {
+                        if (event.angleDelta.y < 0)
+                            tabBar.currentIndex = Math.min(tabBar.currentIndex + 1, root.tabButtonList.length - 1)
+                        else if (event.angleDelta.y > 0)
+                            tabBar.currentIndex = Math.max(tabBar.currentIndex - 1, 0)
+                    }
+                    acceptedDevices: PointerDevice.Mouse | PointerDevice.TouchPad
+                }
+            }
+
+            Repeater {
+                model: root.tabButtonList
+                delegate: SecondaryTabButton {
+                    selected: (index == currentTab)
+                    buttonText: modelData.name
+                    buttonIcon: modelData.icon
+                }
+            }
+        }
+
+        Item { // Tab indicator
+            id: tabIndicator
+            Layout.fillWidth: true
+            height: 3
+            property bool enableIndicatorAnimation: false
+            Connections {
+                target: root
+                function onCurrentTabChanged() {
+                    tabIndicator.enableIndicatorAnimation = true
+                }
+            }
+
+            Rectangle {
+                id: indicator
+                property int tabCount: root.tabButtonList.length
+                property real fullTabSize: root.width / tabCount
+                property real targetWidth: tabBar.contentItem.children[0].children[tabBar.currentIndex].tabContentWidth
+
+                implicitWidth: targetWidth
+                anchors {
+                    top: parent.top
+                    bottom: parent.bottom
+                }
+
+                x: tabBar.currentIndex * fullTabSize + (fullTabSize - targetWidth) / 2
+
+                color: Appearance.colors.colPrimary
+                radius: Appearance.rounding.full
+
+                Behavior on x {
+                    enabled: tabIndicator.enableIndicatorAnimation
+                    animation: Appearance.animation.elementMove.numberAnimation.createObject(this)
+                }
+
+                Behavior on implicitWidth {
+                    enabled: tabIndicator.enableIndicatorAnimation
+                    animation: Appearance.animation.elementMove.numberAnimation.createObject(this)
+                }
+            }
+        }
+
+        Rectangle { // Tabbar bottom border
+            id: tabBarBottomBorder
+            Layout.fillWidth: true
+            height: 1
+            color: Appearance.colors.colOutlineVariant
+        }
+
+        SwipeView {
+            id: swipeView
+            Layout.topMargin: 10
+            Layout.fillWidth: true
+            Layout.fillHeight: true
+            spacing: 10
+            clip: true
+            currentIndex: currentTab
+            onCurrentIndexChanged: {
+                tabIndicator.enableIndicatorAnimation = true
+                currentTab = currentIndex
+            }
+
+            // Pomodoro Timer Tab
+            Item {
+                ColumnLayout {
+                    anchors.horizontalCenter: parent.horizontalCenter
+                    spacing: 18
+
+                    StyledText {
+                        Layout.alignment: Qt.AlignHCenter
+                        text: Pomodoro.timeFormattedPomodoro()
+                        font.pixelSize: 50
+                        color: Appearance.m3colors.m3onSurface
+                    }
+
+                    RowLayout {
+                        Layout.alignment: Qt.AlignHCenter
+                        spacing: 20
+
+                        DialogButton {
+                            buttonText: Pomodoro.isPomodoroRunning ? Translation.tr("Pause") : Translation.tr("Start")
+                            Layout.preferredWidth: 90
+                            Layout.preferredHeight: 35
+                            font.pixelSize: Appearance.font.pixelSize.larger
+                            onClicked: Pomodoro.togglePomodoro()
+                            background: Rectangle {
+                                color: Appearance.m3colors.m3onSecondary
+                                radius: Appearance.rounding.normal
+                                border.color: Appearance.m3colors.m3outline
+                                border.width: 1
+                            }
+                        }
+
+                        StyledText {
+                            Layout.alignment: Qt.AlignHCenter
+                            text: Pomodoro.isPomodoroBreak ? Translation.tr("Break time") : Translation.tr("Focus time")
+                            font.pixelSize: Appearance.font.pixelSize.largest
+                            color: Appearance.m3colors.m3onSurface
+                        }
+
+                        DialogButton {
+                            buttonText: Translation.tr("Reset")
+                            Layout.preferredWidth: 90
+                            Layout.preferredHeight: 35
+                            font.pixelSize: Appearance.font.pixelSize.larger
+                            onClicked: Pomodoro.pomodoroReset()
+                            background: Rectangle {
+                                color: Appearance.m3colors.m3onError
+                                radius: Appearance.rounding.normal
+                                border.color: Appearance.m3colors.m3outline
+                                border.width: 1
+                            }
+                        }
+                    }
+
+                    ColumnLayout {
+                        Layout.alignment: Qt.AlignHCenter
+                        spacing: 0
+
+                        StyledText {
+                            text: Translation.tr("Focus Duration: %1 min").arg(Pomodoro.pomodoroWorkTime / 60)
+                            color: Appearance.m3colors.m3onSurface
+                        }
+                        Slider {
+                            id: workTimeSlider
+                            Layout.fillWidth: true
+                            from: 5
+                            to: 120
+                            stepSize: 1
+                            value: Pomodoro.pomodoroWorkTime / 60
+                            onValueChanged: Pomodoro.pomodoroWorkTime = value * 60
+                            handle: Rectangle {
+                                x: workTimeSlider.leftPadding + workTimeSlider.visualPosition * (workTimeSlider.availableWidth - width)
+                                y: workTimeSlider.topPadding + (workTimeSlider.availableHeight - height) / 2
+                                implicitWidth: 20
+                                implicitHeight: 20
+                                radius: 10
+                                color: Appearance.m3colors.m3onSecondary
+                                border.color: Appearance.m3colors.m3outline
+                            }
+                        }
+
+                        StyledText {
+                            text: Translation.tr("Break Duration: %1 min").arg(Pomodoro.pomodoroBreakTime / 60)
+                            color: Appearance.m3colors.m3onSurface
+                        }
+                        Slider {
+                            id: breakTimeSlider
+                            Layout.fillWidth: true
+                            from: 1
+                            to: 60
+                            stepSize: 1
+                            value: Pomodoro.pomodoroBreakTime / 60
+                            onValueChanged: Pomodoro.pomodoroBreakTime = value * 60
+                            handle: Rectangle {
+                                x: breakTimeSlider.leftPadding + breakTimeSlider.visualPosition * (breakTimeSlider.availableWidth - width)
+                                y: breakTimeSlider.topPadding + (breakTimeSlider.availableHeight - height) / 2
+                                implicitWidth: 20
+                                implicitHeight: 20
+                                radius: 10
+                                color: Appearance.m3colors.m3onSecondary
+                                border.color: Appearance.m3colors.m3outline
+                            }
+                        }
+                    }
+                }
+            }
+
+            // Stopwatch Tab
+            Item {
+                ColumnLayout {
+                    anchors.horizontalCenter: parent.horizontalCenter
+                    spacing: 18
+
+                    StyledText {
+                        Layout.alignment: Qt.AlignHCenter
+                        text: Pomodoro.timeFormattedStopwatch()
+                        font.pixelSize: 50
+                        color: Appearance.m3colors.m3onSurface
+                    }
+
+                    RowLayout {
+                        Layout.alignment: Qt.AlignHCenter
+                        spacing: 20
+
+                        DialogButton {
+                            buttonText: Pomodoro.isStopwatchRunning ? Translation.tr("Pause") : Translation.tr("Start")
+                            Layout.preferredWidth: 90
+                            Layout.preferredHeight: 35
+                            font.pixelSize: Appearance.font.pixelSize.larger
+                            onClicked: Pomodoro.toggleStopwatch()
+                            background: Rectangle {
+                                color: Appearance.m3colors.m3onSecondary
+                                radius: Appearance.rounding.normal
+                                border.color: Appearance.m3colors.m3outline
+                                border.width: 1
+                            }
+                        }
+
+                        StyledText {
+                            Layout.alignment: Qt.AlignHCenter
+                            text: Translation.tr("Stopwatch")
+                            font.pixelSize: Appearance.font.pixelSize.largest
+                            color: Appearance.m3colors.m3onSurface
+                        }
+
+                        DialogButton {
+                            buttonText: Translation.tr("Reset")
+                            Layout.preferredWidth: 90
+                            Layout.preferredHeight: 35
+                            font.pixelSize: Appearance.font.pixelSize.larger
+                            onClicked: Pomodoro.stopwatchReset()
+                            background: Rectangle {
+                                color: Appearance.m3colors.m3onError
+                                radius: Appearance.rounding.normal
+                                border.color: Appearance.m3colors.m3outline
+                                border.width: 1
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    // + FAB
+    StyledRectangularShadow {
+        target: fabButton
+        radius: fabButton.buttonRadius
+        blur: 0.6 * Appearance.sizes.elevationMargin
+    }
+    FloatingActionButton {
+        id: fabButton
+        anchors.right: parent.right
+        anchors.bottom: parent.bottom
+        anchors.rightMargin: root.fabMargins
+        anchors.bottomMargin: root.fabMargins
+
+        onClicked: {
+            if (currentTab === 0) {
+                Pomodoro.togglePomodoro()
+            } else {
+                Pomodoro.toggleStopwatch()
+            }
+        }
+
+        contentItem: MaterialSymbol {
+            text: (currentTab === 0 && Pomodoro.isPomodoroRunning) || (currentTab === 1 && Pomodoro.isStopwatchRunning) ? "pause" : "play_arrow"
+            horizontalAlignment: Text.AlignHCenter
+            iconSize: Appearance.font.pixelSize.huge
+            color: Appearance.m3colors.m3onPrimaryContainer
+        }
+    }
+}
--- a/.config/quickshell/ii/services/Pomodoro.qml
+++ b/.config/quickshell/ii/services/Pomodoro.qml
@@ -1,0 +1,69 @@
+pragma Singleton
+pragma ComponentBehavior: Bound
+
+import qs.modules.common
+import Quickshell
+import Quickshell.Io
+import QtQuick
+
+/**
+ * Simple Pomodoro time manager.
+ */
+Singleton {
+    id: root
+
+    property int pomodoroWorkTime: 25 * 60  // 25 minutes in seconds
+    property int pomodoroBreakTime: 5 * 60  // 5 minutes in seconds
+    property int pomodoroTime: pomodoroWorkTime
+    property bool isPomodoroRunning: false
+    property bool isPomodoroBreak: false
+    property int stopwatchTime: 0
+    property bool isStopwatchRunning: false
+
+    function togglePomodoro() {
+        isPomodoroRunning = !isPomodoroRunning;
+    }
+
+    function toggleStopwatch() {
+        isStopwatchRunning = !isStopwatchRunning;
+    }
+
+    function pomodoroReset() {
+        pomodoroTime = pomodoroWorkTime;
+        isPomodoroRunning = false;
+        isPomodoroBreak = false;
+    }
+
+    function stopwatchReset() {
+        stopwatchTime = 0;
+        isStopwatchRunning = false;
+    }
+
+    function tickSecond() {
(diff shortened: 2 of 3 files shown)


.config/quickshell/ii/services/Pomodoro.qml
```diff
@@ -0,0 +1,69 @@
...
+                Quickshell.execDetached(["bash", "-c", `notify-send "🔴 Pomodoro started!" "Focus for ${Math.floor(pomodoroWorkTime / 60)} minutes."`]);
+            }
+        }
+    }
+
+    function timeFormattedPomodoro() {
+        let minutes = Math.floor(pomodoroTime / 60);
+        let seconds = Math.floor(pomodoroTime % 60);
+        return `${minutes.toString().padStart(2, '0')}:${seconds.toString().padStart(2, '0')}`;
+    }
```
end-4: Why not make this a variable binding?

.config/quickshell/ii/services/Pomodoro.qml
```diff
@@ -0,0 +1,69 @@
...
+    function tickSecond() {
+        if (pomodoroTime > 0) {
+            pomodoroTime--;
+        } else {
+            isPomodoroBreak = !isPomodoroBreak;
+            pomodoroTime = isPomodoroBreak ? pomodoroBreakTime : pomodoroWorkTime;
+            if (isPomodoroBreak) {
+                Quickshell.execDetached(["bash", "-c", `notify-send "☕ Short Break!" "Relax for ${Math.floor(pomodoroBreakTime / 60)} minutes."`]);
+            } else {
+                Quickshell.execDetached(["bash", "-c", `notify-send "🔴 Pomodoro started!" "Focus for ${Math.floor(pomodoroWorkTime / 60)} minutes."`]);
```
end-4: Use Translation.tr() for messages that can benefit from translations

.config/quickshell/ii/modules/sidebarRight/pomodoro/PomodoroWidget.qml
```diff
@@ -0,0 +1,356 @@
...
+
+    Timer {
+        id: stopwatchTimer
+        interval: 1000
+        running: Pomodoro.isStopwatchRunning
+        repeat: true
+        onTriggered: {
+            Pomodoro.stopwatchTime += 1
+        }
+    }
```
end-4: - ~~Do we really need two timers? If not combine into one and have an `if`~~ 
- ~~Move the Timer to the Pomodoro service to make sure it still runs when the sidebar is closed and unloaded~~

Make the service timestamp-based and rely on `DateTime.clock`

.config/quickshell/ii/modules/sidebarRight/pomodoro/PomodoroWidget.qml
```diff
@@ -0,0 +1,356 @@
...
+                                implicitWidth: 20
+                                implicitHeight: 20
+                                radius: 10
+                                color: Appearance.m3colors.m3onSecondary
+                                border.color: Appearance.m3colors.m3outline
+                            }
+                        }
+                    }
+                }
+            }
```
end-4: StyledSlider exists. Use it.

.config/quickshell/ii/modules/sidebarRight/pomodoro/PomodoroWidget.qml
```diff
@@ -0,0 +1,356 @@
...
+                            Layout.preferredHeight: 35
+                            font.pixelSize: Appearance.font.pixelSize.larger
+                            onClicked: Pomodoro.toggleStopwatch()
+                            background: Rectangle {
+                                color: Appearance.m3colors.m3onSecondary
+                                radius: Appearance.rounding.normal
+                                border.color: Appearance.m3colors.m3outline
+                                border.width: 1
+                            }
+                        }
```
end-4: - This is not a dialog so don't use a DialogButton. The correct one is RippleButton
- To reduce visual clutter, don't use a border here

.config/quickshell/ii/modules/sidebarRight/pomodoro/PomodoroWidget.qml
```diff
@@ -0,0 +1,356 @@
...
+                            Layout.preferredHeight: 35
+                            font.pixelSize: Appearance.font.pixelSize.larger
+                            onClicked: Pomodoro.stopwatchReset()
+                            background: Rectangle {
+                                color: Appearance.m3colors.m3onError
+                                radius: Appearance.rounding.normal
+                                border.color: Appearance.m3colors.m3outline
+                                border.width: 1
+                            }
+                        }
```
end-4: Same as above

.config/quickshell/ii/modules/sidebarRight/pomodoro/PomodoroWidget.qml
```diff
@@ -0,0 +1,356 @@
...
+            }
+        }
+
+        contentItem: MaterialSymbol {
+            text: (currentTab === 0 && Pomodoro.isPomodoroRunning) || (currentTab === 1 && Pomodoro.isStopwatchRunning) ? "pause" : "play_arrow"
+            horizontalAlignment: Text.AlignHCenter
+            iconSize: Appearance.font.pixelSize.huge
+            color: Appearance.m3colors.m3onPrimaryContainer
+        }
+    }
```
end-4: There is already a Pause button, and this action is not something to use a FAB for. Remove this

end-4:
## Additional changes to be made
1. Break/work time:
- Use a ConfigSpinBox instead of a slider because minutes are discrete, and time is not something you can "feel" as you drag
- Make the configuration change the options `Config.option.time.pomodoro.work` and `Config.option.time.pomodoro.break` (to be made) instead of just modifying the time in the service

2. Service: Use timestamps instead of counting down (solves question (3))
- Have a variable for timer start
- In each timer tick update timer representation values based on `(current timestamp) - (starting timestamp)`
- Store starting timestamps and whether the timers should be running in `Persistent.states.time.pomodoro.startTime`, `Persistent.states.time.stopwatch.startTime`, `Persistent.states.time.pomodoro.running`, and `Persistent.states.time.stopwatch.running`

## Your questions

> 1. The Pomodoro didn't updated when right sidebar was not opened. It was fixed after 596ae72

It did not "fix" it, it simply provides the option to keep the right sidebar content loaded when closed. The service should be made to work when the UI for it on the sidebar is unloaded, as instructed in the review above

> 2. I will add functionality like laps in Stopwatch only if requested as I personally don't have use for it.
> 3. The Focus duration and break duration fall back to 25 minutes and 5 minutes after reloading. I don't know where to save these two settings for reloading after restart.

The change number (2) above should fix question (3). Question (2) should also be very easy (just have a list of timestamps) but I can do it later

.config/quickshell/ii/modules/sidebarRight/pomodoro/PomodoroWidget.qml
end-4: Use a ConfigSpinBox instead, see comment below

.config/quickshell/ii/modules/sidebarRight/pomodoro/PomodoroWidget.qml
nyx-4: I personally also thought that but just let it there to be similar with To Do, I can also add long break because now we have more UI space.

end-4: If you want to add it then it should be there regardless of UI space

end-4:
I think this is a cleaner layout 
image

.config/quickshell/ii/modules/sidebarRight/pomodoro/PomodoroWidget.qml
nyx-4: I have updated it to use RippleButton, but I didn't find any way to center align the text in a Ripple Button. The text appears left side.

.config/quickshell/ii/services/Pomodoro.qml
nyx-4: Sorry, I am quite new to QuickShell and JavaScript. How to **make a variable binding**?

.config/quickshell/ii/modules/sidebarRight/pomodoro/PomodoroWidget.qml
end-4: try horizontalAlignment, anchors.centerIn, and if those don't work maybe wrap the content item StyledText in an Item with anchors.centerIn

.config/quickshell/ii/services/Pomodoro.qml
end-4: I meant this
```qml
timeFormattedPomodoro: {
    let minutes = Math.floor(pomodoroTime / 60);
    let seconds = Math.floor(pomodoroTime % 60);
    return `${minutes.toString().padStart(2, '0')}:${seconds.toString().padStart(2, '0')}`;
}
```
I might not be 100% correct on the terminology, but you declare it like this and use `timeFormattedPomodoro` as a property instead of a function
Both ways work, but I didn't know declaring a function like you did also works so I've been using the above. This suggestion is just for consistency

nyx-4:
## Describe your changes
1. ~Tried my best to add _variable binding_ although a bit hacky way.~
2. ~Used Translate.tr() method, but still missed some spots.~
3. ~Used Ripple button instead, the center alignment didn't worked.~
4. The stopwatch was untouched, what to do with it? _(Edit: Right now, I am a bit busy. I'll see in some other PR if to work on Stopwatch also.)_
5. ~The ConfigSpinBox was added, see if it needs be changed.~
6. I want to get a notification with bell sound as I mostly leave my Laptop when break. Its an audible way to get back. Which command to use and which bell sound to set default? 
7. **Long break Duration** is a bit long, what should I rename them to be concise? **Long break Time** is okay in length but seems out-of-place, didn't it?
8. **Pause** and **Reset** should be centered, maybe you should try to do them.

## A quick peek
image


Edit 2:

## Stopwatch
1. I will try to make stopwatch use time stamp also, and update the logic accordingly.
2. What should the UI of stopwatch look like? This image did helped me alot.
   > I think this is a cleaner layout image

nyx-4:
## Stopwatch
This is a stopwatch, the UI is very bad. Do you have any ideas for UI? The problem is that the laps are not scroll-able.

## A quick peek
image

.config/quickshell/ii/modules/common/Config.qml
```diff
@@ -253,6 +253,12 @@ Singleton {
                 // https://doc.qt.io/qt-6/qtime.html#toString
                 property string format: "hh:mm"
                 property string dateFormat: "ddd, dd/MM"
+                property JsonObject pomodoro: JsonObject {
+                    property int breaktime: 300
+                    property int cycle: 4
+                    property int focus: 1500
+                    property int longbreak: 1200
+                }
```
end-4: camelCase please

.config/quickshell/ii/services/Pomodoro.qml
```diff
@@ -0,0 +1,95 @@
...
+import Quickshell.Io;
+import QtQuick;
+
+/**
+ * Simple Pomodoro time manager.
+ */
+Singleton {
+    id: root
+
+    // TODO: read these values from a config file.
```
end-4: This TODO is done so remove the comment

.config/quickshell/ii/services/Pomodoro.qml
```diff
@@ -0,0 +1,95 @@
...
+    property int pomodoroFocusTime: Config.options.time.pomodoro.focus
+    property int pomodoroBreakTime: Config.options.time.pomodoro.breaktime
+    property int pomodoroLongBreakTime: Config.options.time.pomodoro.longbreak
+    property int pomodoroLongBreakCycle: Config.options.time.pomodoro.cycle
+
+    property int pomodoroTimeLeft: pomodoroFocusTime
+    property int getPomodoroSecondsLeft: pomodoroFocusTime
+    property int pomodoroTimeStarted: getCurrentTime()  // The time pomodoro was last Resumed
```
end-4: Wouldn't this going to be constantly triggered? Please only do this when necessary

.config/quickshell/ii/modules/sidebarRight/pomodoro/PomodoroWidget.qml
```diff
@@ -166,102 +166,102 @@ Item {
...
-                            Layout.preferredHeight: 35
-                            font.pixelSize: Appearance.font.pixelSize.larger
-                            onClicked: Pomodoro.pomodoroReset()
-                            background: Rectangle {
-                                color: Appearance.m3colors.m3onError
-                                radius: Appearance.rounding.normal
-                                border.color: Appearance.m3colors.m3outline
-                                border.width: 1
+                            RippleButton {
+                                buttonText: Translation.tr("Reset")
```
end-4: To ensure proper center alignment of the text, try using this instead of `buttonText` property
```qml
contentItem: StyledText {
    anchors.centerIn: parent
    horizontalAlignment: Text.AlignHCenter
    text: "Reset"
}
```
if that doesn't work try this
```qml
contentItem: Item {
    anchors.centerIn: parent
    implicitWidth: buttonTextItem.implicitWidth
    implicitHeight: buttonTextItem.implicitHeight
    StyledText {
        anchors.fill: parent
        text: "Reset"
    }
}
```

.config/quickshell/ii/modules/sidebarRight/pomodoro/PomodoroWidget.qml
```diff
@@ -166,102 +166,102 @@ Item {
...
-                                y: workTimeSlider.topPadding + (workTimeSlider.availableHeight - height) / 2
-                                implicitWidth: 20
-                                implicitHeight: 20
-                                radius: 10
-                                color: Appearance.m3colors.m3onSecondary
-                                border.color: Appearance.m3colors.m3outline
+                        spacing: 10
+
+                        ConfigSpinBox {
+                            text: Translation.tr("Focus Duration: ")
```
end-4: Typography: remove the colon
(the space after it is also unnecessary)

.config/quickshell/ii/services/Pomodoro.qml
```diff
@@ -1,69 +1,95 @@
...
+                    Translation.tr(`Relax for %1 minutes.`).arg(Math.floor(pomodoroBreakTime / 60))
+                ])
             } else {
-                Quickshell.execDetached(["bash", "-c", `notify-send "🔴 Pomodoro started!" "Focus for ${Math.floor(pomodoroWorkTime / 60)} minutes."`]);
+                Quickshell.execDetached([
+                    "notify-send",
+                    Translation.tr("🔴 Pomodoro started!"),
+                    Translation.tr(`Focus for %1 minutes.`).arg(Math.floor(pomodoroFocusTime / 60))
+                ])
+                pomodoroCycle += 1
```
end-4: - Add `-a Shell` for app name
- Let's have the title be "Pomodoro" and the body text be the actual content
- Typography: Don't put "." at the end

So something like 

```qml
Quickshell.execDetached([
    "notify-send",
    Translation.tr("Pomodoro"),
    Translation.tr(`🌿 Long break: %1 minutes`).arg(Math.floor(pomodoroLongBreakTime / 60)),
    "-a", "Shell"
])
```

end-4:
image

end-4:
Also for the stopwatch, for the main number display 
- make the `.79` smaller 
- either:
  - grey out hours/minutes if they're 00
  - not include hours

.config/quickshell/ii/modules/sidebarRight/pomodoro/PomodoroWidget.qml
end-4: see my new comment in the most recent review

end-4:
> 6. I want to get a notification with bell sound as I mostly leave my Laptop when break. Its an audible way to get back. Which command to use and which bell sound to set default?

I would rather just have the notification because 
- Sounds can be disturbing and 
- Sound packs on Linux generally feel old

> 7. Long break Duration is a bit long, what should I rename them to be concise? Long break Time is okay in length but seems out-of-place, didn't it?

Have a heading called "Durations" then you can remove that word from each line

nyx-4:
## Known bug / issues
- [x] The laps in Stopwatch are not scrollable. I must have missed something. (Fixed)
- [ ] The `ConfigSpinBox` do not read properly from `Config.options...` although they write perfectly to `Config.options...`
- [x] There is a weird behavior with headers, I've copy/pasted that stuff from `Todo/`. I'll look closely into it. (Not reproduce-able after reboot)
- [x] The keybinding copied from `Todo/` shall also be updated. (Suggest some sane keybinds 🥺) (Fixed: R for Reset, S for Start,Stop and L for laps)


## A quick peek (Pomodoro)
image


## A quick peek (Stopwatch)
image


## Weird peeks
Explanation: The focus in `Circular Progress` and focus in `ConfigSpinBox` are not sync.

image

Explanation: The weird headers.

image

.config/quickshell/ii/modules/sidebarRight/pomodoro/PomodoroWidget.qml
```diff
@@ -52,20 +54,18 @@ Item {
 
     Timer {
         id: pomodoroTimer
-        interval: 1000
-        running: Pomodoro.isPomodoroRunning
+        interval: 200
```
end-4: Why 200? We don't need such precision

.config/quickshell/ii/modules/sidebarRight/pomodoro/PomodoroWidget.qml
```diff
@@ -0,0 +1,494 @@
...
+        id: pomodoroTimer
+        interval: 200
+        running: Config.options.time.pomodoro.running
+        repeat: true
+        onTriggered: Pomodoro.tickSecond()
+    }
+
+    Timer {
+        id: stopwatchTimer
+        interval: 10
```
end-4: Yikes. Let's either have this 1000ms or not have the stopwatch...

.config/quickshell/ii/modules/sidebarRight/pomodoro/PomodoroWidget.qml
```diff
@@ -174,17 +174,19 @@ Item {
                         CircularProgress {
                             Layout.alignment: Qt.AlignHCenter
                             lineWidth: 7
+                            gapAngle: Math.PI / 14
```
end-4: this is in degrees not radians. how does it even work and not have near-0 gap...

end-4 (commented):
- Use [SystemClock](https://quickshell.org/docs/master/types/Quickshell/SystemClock/) and get your way around with bindings instead of manually fetching the time in extremely short intervals

.config/quickshell/ii/modules/sidebarRight/pomodoro/PomodoroWidget.qml
nyx-4: When >= 500, it skips seconds. Like from 24:54 to 24:52 to 24:50. It looks annoying. I will make it 300.

nyx-4: This behavior happens because I'm using Math.floor. Using Math.round wouldn't help either.

## Example
```js
Math.floor(24.99) === 24
Math.floor(24.01) === 24

// Although 24.99 - 24.01 ~= 1
```

end-4: This is why we should make it use SystemClock...

.config/quickshell/ii/modules/sidebarRight/pomodoro/PomodoroWidget.qml
nyx-4: According to **NIST Recommendation**:
> Resolution of 10 ms is common for digital stopwatches, but some devices have 1 ms resolution (0.001 s), or even smaller. For analog stopwatches, a common resolution is 1/5 of a second, or 0.2 s. [^1]


My points for 10ms resolution:
1. Having 10ms resolution for stopwatch is more standard for digital stopwatch.
2. In UI design, _POLA_ proposes that a component of a system should behave in a way that most users will expect it to behave, and therefore not astonish or surprise users. [^2] [^3]
3. The Quickshell's SystemClock is not a good idea for stopwatch/Pomodoro. (3 Problems: Midnight, DST, and Midnight) [^4]


According to Description of [The Problem with Time & Timezones - Computerphile
](https://www.youtube.com/watch?v=-5wpm-gesOY)

> A web app that works out how many seconds ago something happened. How hard can coding that be? Tom Scott explains how time twists and turns like a twisty-turny thing. It's not to be trifled with! [^4]



[^1]: https://tf.nist.gov/general/pdf/2281.pdf
[^2]: https://en.wikipedia.org/wiki/Principle_of_least_astonishment
[^3]: https://en.wikipedia.org/wiki/Convention_over_configuration
[^4]: https://www.youtube.com/watch?v=-5wpm-gesOY

.config/quickshell/ii/modules/sidebarRight/pomodoro/PomodoroWidget.qml
nyx-4: See my reply on next comment. It is nasty to deal with time-zones, day-light saving time and above all mid-night. And I forget new year's eve. Date.now() is standard JS and works fine.


Secondly, this timer is to just update the UI. The backend time is calculated using `getCurrentTimeIn10ms()` and `getCurrentTimeInSeconds()` function in services/Pomodoro.qml.


Thirdly, the 200 precision here means update `getPomodoroSecondsLeft` after every 200ms. It has nothing to do with time or precision of Pomodoro.

.config/quickshell/ii/modules/sidebarRight/pomodoro/PomodoroWidget.qml
nyx-4: I don't know, it just looks OK. I can change it, if requested.
1. Setting to zero will kind of break the UI consistency.
2. Possible values are 10-15. 14 is just a good-enough guess. (Hint: its 10 + lucky number)
3. The gapAngle for Stopwatch is PI / 18 (can be 15-20). so PI / 14 doesn't look too odd. (Yep, thats' a circular dependency)

.config/quickshell/ii/modules/sidebarRight/pomodoro/PomodoroWidget.qml
```diff
@@ -0,0 +1,494 @@
...
+                        RowLayout {
+                            Layout.alignment: Qt.AlignHCenter
+                            spacing: 0
+
+                            ConfigSpinBox {
+                                id: focusSpinBox
+                                spacing: 0
+                                Layout.leftMargin: 0
+                                Layout.rightMargin: 0
+                                value: Config.options.time.pomodoro.focus / 60
```
nyx-4: @end-4 This line always read 1500 (25 min, default value) regardless of what's in `config.json`. This is the only known issue in current implementation. Am I missing something? or is it some issue with Config.

Edit: Whenever quickshell reloads, the values in ConfigSpinBox are always the default one's. But it do write back changes in config.json

nyx-4: ## Triangulating problem
Putting old-school `console.log` to find the source of problem. As seen here, the **Value loaded** is before **Configuration Loaded**. This explains that I am missing something. I should load the value after **Configuration Loaded**.

## Logs
```sh
  INFO: Reloading configuration...
  WARN scene: @modules/bar/ActiveWindow.qml[17:-1]: TypeError: Cannot read property 'monitor' of null
  WARN scene: QML ColumnLayout at @modules/background/Background.qml[200:17]: Cannot anchor to an item that isn't a parent or sibling.
  WARN scene: @modules/background/Background.qml[31:-1]: TypeError: Cannot read property 'monitor' of null
 DEBUG qml: ------   Value loaded   ------
  INFO: Configuration Loaded
 DEBUG qml: [Notifications] File loaded
 DEBUG qml: [To Do] File loaded
```

nyx-4:
I opted to submit this PR because I wanted a Pomodoro myself, and I had some free time over weekend. This was a great experience but I have more pressing things to do (in Python) and my College stuff 😏. I will always remember this as my _first PR_. Thanks!

lash0000 (commented):
```py
 DEBUG qml: [To Do] File loaded
  INFO: Reloading configuration...
 ERROR: Failed to load configuration
 ERROR:   caused by @shell.qml[74:57]: Type SidebarRight unavailable
 ERROR:   caused by @modules/sidebarRight/SidebarRight.qml[179:25]: Type BottomWidgetGroup unavailable
 ERROR:   caused by @modules/sidebarRight/BottomWidgetGroup.qml[246:7]: Type PomodoroWidget unavailable
 ERROR:   caused by @modules/sidebarRight/pomodoro/PomodoroWidget.qml[184:29]: Cannot assign to non-existent property "secondaryColor"
  INFO: Watching additional files picked up in reload for changes...
```

I try it with latest commit from main so I feel that it needs to adjust the `PomodoroWidget.qml` dunno I can't try...

.config/quickshell/ii/modules/sidebarRight/pomodoro/PomodoroWidget.qml
end-4: Uhhhhh I actually tried your implementation, it doesn't really spike the cpu or anything so it's fine I guess
I understand why not using SystemClock makes sense now

.config/quickshell/ii/modules/sidebarRight/pomodoro/PomodoroWidget.qml
end-4: I realized this is because you're basing this off an older version of what I have right now
My fault for changing it and this can be fixed later

nyx-4:
> ```python
>  DEBUG qml: [To Do] File loaded
>   INFO: Reloading configuration...
>  ERROR: Failed to load configuration
>  ERROR:   caused by @shell.qml[74:57]: Type SidebarRight unavailable
>  ERROR:   caused by @modules/sidebarRight/SidebarRight.qml[179:25]: Type BottomWidgetGroup unavailable
>  ERROR:   caused by @modules/sidebarRight/BottomWidgetGroup.qml[246:7]: Type PomodoroWidget unavailable
>  ERROR:   caused by @modules/sidebarRight/pomodoro/PomodoroWidget.qml[184:29]: Cannot assign to non-existent property "secondaryColor"
>   INFO: Watching additional files picked up in reload for changes...
> ```
> 
> I try it with latest commit from main so I feel that it needs to adjust the `PomodoroWidget.qml` dunno I can't try...

In [0708070764cf0a15642b646ae0d21cae0c5fedab](https://github.com/end-4/dots-hyprland/commit/0708070764cf0a15642b646ae0d21cae0c5fedab#diff-d69afc3055598270929d37c9e08268279d0044fcb4f1a3858106418e91df55efL17-R16), some properties are renamed. Replace `primaryColor` with `colPrimary`. It should work _fine_. It's a trivial change.

end-4:
I've temporarily disabled the pomodoro time config because it's kinda mysterious

.config/quickshell/ii/services/Pomodoro.qml
```diff
@@ -49,30 +49,34 @@ Singleton {
...
-            Persistent.states.timer.pomodoro.isBreak = !isBreak
-            Persistent.states.timer.pomodoro.start += timeLeft
-            timeLeft = isBreak ? breakTime : focusTime
-
+        if (getCurrentTimeInSeconds() >= Persistent.states.timer.pomodoro.start + pomodoroLapDuration) {
+            // Reset counts
+            const currentTimeInSeconds = getCurrentTimeInSeconds()
+            Persistent.states.timer.pomodoro.isBreak = !Persistent.states.timer.pomodoro.isBreak
+            Persistent.states.timer.pomodoro.isLongBreak = Persistent.states.timer.pomodoro.isBreak && (pomodoroCycle + 1 == cyclesBeforeLongBreak)
+            Persistent.states.timer.pomodoro.start = currentTimeInSeconds
```
nyx-4: You are declaring a `const currentTimeInSeconds` and only using it once, why bother with declaration at all.

end-4: fixed in ca6463cae8f828a032a4136c489f8347196f256f
it was something i forgot to remove from my debug logging when dealing with persistent state race conditions

end-4:
- I think the UI is ready. It has small problems that I can't be bothered to fix because I won't gonna use it that much
  - The colons of the stopwatch move when numbers change
  - The spinboxes aren't there in the pomodoro tab but for the sake of simplicity and self-control it's probably a good idea to put that in the settings app anyway
- Other problems
  - Buggy stopwatch persistence: If we start the timer, record laps and reload the shell, the overall time goes back to 0 which is okay but somehow the laps are kept. This should be fixed before merging
  - The service uses timestamps in seconds/10ms which is hella weird, but I'm kinda lazy to fix this

.config/quickshell/ii/services/TimerService.qml
```diff
@@ -98,6 +97,7 @@ Singleton {
     function resetPomodoro() {
         Persistent.states.timer.pomodoro.running = false;
         Persistent.states.timer.pomodoro.isBreak = false;
+        Persistent.states.timer.pomodoro.isLongBreak = false;
```
nyx-4: Is there any reason to use `isLongBreak` when it can be deduced from `...pomodoro.cycle`? Similarly about `isPomodoroLongBreak`?

end-4: c2d5d2b61ae4d4905ecc7f61e7dbf90e46557071

end-4:
I made the Start button of the stopwatch clear the laps if there are left
Not ideal but I say acceptable

PR #1770, opened 2025-08-04
quickshell: fix qml null safety and monitor property errors

finjener:
## Describe your changes

I have fixed some type errors and null checks I saw when restarting quickshell, like;
  @modules/bar/ActiveWindow.qml[17:-1]: TypeError: Cannot read property 'monitor' of null
  @modules/bar/ActiveWindow.qml[17:-1]: TypeError: Cannot read property 'monitor' of null
  @modules/background/Background.qml[31:-1]: TypeError: Cannot read property 'monitor' of null

It is ready, no question/feedback needed.

### DIFF
--- a/.local/state/quickshell/user/ai/chats/chat.json
+++ b/.local/state/quickshell/user/ai/chats/chat.json
@@ -1,0 +1,7 @@
+{
+  "messages": [],
+  "created": "2024-01-01T00:00:00Z",
+  "updated": "2024-01-01T00:00:00Z",
+  "title": "Default Chat",
+  "id": "default"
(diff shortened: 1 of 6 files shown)


.local/state/quickshell/user/ai/chats/chat.json
end-4, line 1: don't create this file, make the service not try to load it or shut up when it fails instead

finjener: ok, I deleted the file.

end-4 (approved):
thanks

PR #1775, opened 2025-08-06
Fix hiding background when fullscreen

lunstia:
## Describe your changes

The following video shows the result of the current implementation:

https://github.com/user-attachments/assets/84bd2e65-090d-422a-9da8-66bd89815a27

Current implementation does not handle extra monitors very well.

However, to fix this, I filter through the workspaces for that current monitor, I then filter through the toplevel windows to find any that are fullscreen while making sure that's for the current active workspace.

This way we only care for the active workspace on that monitor and if it has a fullscreen toplevel window.

I am not sure how to detect if a window is transparent as I feel that would need to be taken in account as well as possible suggestions on refactoring the code if it's not very well structured, with the current documentation, I couldn't see any other way to write it.

### DIFF
--- a/.config/quickshell/ii/modules/background/Background.qml
+++ b/.config/quickshell/ii/modules/background/Background.qml
@@ -26,9 +26,10 @@ Variants {
         required property var modelData
 
         // Hide when fullscreen
-        readonly property var monitorHasFullscreen: ToplevelManager.toplevels.values.filter(window=>(window.screens[0].name == monitor.name) && window.fullscreen)[0]
-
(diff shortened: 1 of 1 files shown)


lunstia:
Stupidly overlooked an error when switching to other workspaces

lunstia:
Redid how I went about it and it should be fully working now with the testing I did, there was no way the quickshell documentation was up-to-date yesterday as I got home and found that toplevels were easily listed in HyprlandWorkspace

end-4:
This works, thank you!

PR #1773, opened 2025-08-05
System resource tooltips

finjener:
## Describe your changes

I added hover popups for system resources and upcoming todos like in wheather.

photo_2025-08-06_00-12-28
photo_2025-08-06_00-12-31
photo_2025-08-06_00-12-36
photo_2025-08-06_00-12-39

## Is it ready? Questions/feedback needed?

It is ready, feedback neededte

PR #1771, opened 2025-08-05
Added a small battery popup to show information

Runzelee:
## Describe your changes

Added a small battery popup to show temperature, remaining time and energy rate

image
image

## Is it ready? Questions/feedback needed?
It is ready, no question/feedback needed.

PR #1776, opened 2025-08-06
feat(bar): unified popup handling (follow-up to #1771 and #1773)

Runzelee:
## Describe your changes
This PR builds upon my previous work in #1771 and @finjener's PR #1773, merging both approaches and further refining popup handling for a more polished and consistent user experience.

### Before:
https://github.com/user-attachments/assets/b5fc3be4-7103-40f7-938f-7998891c80be

### After:
https://github.com/user-attachments/assets/de71ef82-5550-423b-8c18-833240169e96

### Key changes:
1. Unified popup handling in `ClockWidget`, `Resource`, `BatteryPopup`, and `WeatherBar` using the `PanelWindow` + `LazyLoader` pattern, allowing all popups to benefit from compositor animations and consistent positioning.

2. Updated `ClockWidget` and `Resource` to use `ColumnLayout` and `RowLayout` instead of plain text where appropriate, adding `MaterialSymbol` icons for visual consistency with the overall desktop style and my `BatteryPopup`.

3. Added `Translation.tr()` with bilingual (Chinese/English) translations to avoid hardcoded strings.

---

This PR provides a unified, more maintainable implementation that supersedes #1771 and complements #1773. 

It is ready. Feedback is welcome.


.config/quickshell/ii/services/Battery.qml
```diff
@@ -21,6 +23,43 @@ Singleton {
...
+                    Battery.temperature = parseInt(text.trim()) / 1000
+                } else {
+                    Battery.temperature = 0
+                }
+            }
+        }
+    }
+
+    Timer {
+        interval: 3000
```
end-4: make this follow `Config.options.resources.updateInterval` (it also defaults to 3000)

.config/quickshell/ii/services/Battery.qml
```diff
@@ -21,6 +23,43 @@ Singleton {
...
+        stdout: StdioCollector {
+            onStreamFinished: {
+                if (text && text.trim() !== "") {
+                    Battery.temperature = parseInt(text.trim()) / 1000
+                } else {
+                    Battery.temperature = 0
+                }
+            }
+        }
+    }
```
end-4: 1. If it's reading a file use FileView
2. Why `thermal_zone0`? Are you sure this is the battery temperature? From [this StackExchange thread](https://unix.stackexchange.com/questions/756041/why-is-the-result-of-cat-sys-class-thermal-thermal-zone0-temp-0) it's not necessarily battery temperature. Unless there is further documentation that says for sure let's just not have this

end-4 (commented):
Needs some refractoring of the `LazyLoader { PanelWindow { ... } }` to be "unified"

end-4:
It... doesn't work? The popups aren't popping up for me
And why are you using PanelWindow and not PopupWindow

Runzelee:
How's that possible??? Works fine on my machine. With PanelWindow it can inherit the compositor animation, but PopupWindow feels kinda stiff and awkward.

Runzelee:
Let me clean up the code a bit first. I'll pull `LazyLoader { PanelWindow { ... } }` out into `StyledPopup.qml` like you suggested. If the popup still doesn't show up on your side, maybe check the logs for anything weird.

.config/quickshell/ii/services/Battery.qml
Runzelee: I know this isn't actually the battery temperature, but on Linux there aren't many ways to read temps and getting the real battery temp is usually impossible. For a more general ‘device temperature’, aside from `cat /sys/class/thermal/thermal_zone0/temp`, the only options are lm-sensors or Intel/AMD specific tools, which users might not even have installed. Still, this reading is super useful for laptop users. From my tests it clearly shows how charging or discharging power affects it, especially charging where I'm seeing +20°C jumps.

Runzelee: anyway,  like you said earlier this value can vary a lot between devices. I'll just drop it for now.

Runzelee:
I added `WlrLayershell.namespace: "quickshell:styledPopup"` and `WlrLayershell.layer: WlrLayer.Overlay` in StyledPopup.
no idea if it'll fix the issue, but worth a try...

end-4:
It appears that was a problem with monitor offset. 
With the following monitor configuration
```ini
monitor=eDP-1,1920x1080@60,300x0,1
```
I get this
image

end-4:
I was gonna say "And that's by design" but I realized "Rich tooltips" should not use inverted colors
M3 Docs: https://m3.material.io/components/tooltips/specs

Question is: Are these tooltips "rich"? Guess it's a yes because there's a lot of info... What's bothering me is the tooltips here are not interactable so it might be more consistent/less misleading to consider it a plain tooltip

Runzelee:
Oh, I see. Personally, I think it's already pretty rich here. Whether something feels ‘rich’ isn't about interactivity, but more about how much info and structure it has. At least in the doc you showed me, the inverted color parts are all short, toast-like messages.

https://m3.material.io/components/tooltips/guidelines#edbe33b2-ace1-42cc-a843-ec3790d6bdf3
"Avoid wrapping text to multiple lines or including many pieces of information"

end-4:
This looks good now, I'm going to merge it
Thank you for your work

PR #1780, opened 2025-08-07
Added pomodoro timer

jiboli:
Pomodoro controls are in the right sidebar, when running, timer shows on bar


end-4:
Problems:
- This includes irrelevant personal changes, including matugen templates and the pointless, schizophrenic removal of AI-related config options 
- Config.qml is not the place to dump pomodoro logic into
- The bottom section of the sidebar is a better place to put this
- There's already #1761. It has gone further and I would like to follow this implementation.

PR #1781, opened 2025-08-07
fix(touchpad): improve scroll speed handling for touchpad in StyledListView & StyledFlickable

Runzelee:
### Background

Previously, in all places where `StyledListView` or `StyledFlickable` were used — such as `AiChat`, `Anime`, `NotificationGroup`, `TaskList`, `SearchWidget`, and even `ContentPage` used in settings (and many others) — scrolling with a touchpad was **EXTREMELY** slow, while mouse wheel scrolling was completely normal.

I tried many approaches to address this issue (e.g., tweaking `maximumFlickVelocity`, etc.), but none of them worked. I suspect this might be a bug in Quickshell itself.

In the end, I decided to implement a (somewhat inelegant?) workaround by adding logic to `StyledListView` and `StyledFlickable` to differentiate between mouse and touchpad inputs and adjust scrolling accordingly.

### Current Changes

I’ve added two new properties to both `StyledListView` and `StyledFlickable`:
`touchpadScrollFactor` and `mouseScrollFactor`.

These help differentiate the scroll speed based on input device, and allow for fine-tuned adjustments per component. All existing animations and behaviors are preserved.

Default values:

```qml
property real touchpadScrollFactor: 100
property real mouseScrollFactor: 50
```

These values feel most comfortable based on testing with my ASUS laptop.
However, there are two exceptions: `Anime` and `AiChat`, which have a lot of content and are vertically long. For these, I used:

```qml
touchpadScrollFactor: 600
mouseScrollFactor: 200
```

So far, other components perform well using the default values.

### Demo

This is tested on my laptop (Model: ASUS TX Air 2025) using its built-in touchpad.
I tried to keep my finger movement speed as consistent as possible.

**Before**

https://github.com/user-attachments/assets/d4e925b6-a4c2-47bd-b4c5-897173dea722

**After**

https://github.com/user-attachments/assets/283ed448-0eb4-4d78-ba7a-dd9a4b3ce939

### Additional Notes

If you're developing new components based on `StyledListView` or `StyledFlickable`, remember to add `touchpadScrollFactor` and `mouseScrollFactor` where necessary — though the default values should already work well in most cases.

As for the natural inertia after lifting fingers from the touchpad (which should result in continued scrolling), there's unfortunately nothing I can do — as it was never implemented in the first place.
However, I’ve kept the long-press + drag scrolling behavior, which still provides smooth and elegant deceleration.

If Qt or Quickshell fixes this in the future, feel free to remove this custom logic entirely.

Tested thoroughly on my ASUS laptop.
Feedback is very welcome — especially from those using other laptops or touchpads.

### DIFF
--- a/.config/quickshell/ii/modules/common/widgets/StyledFlickable.qml
+++ b/.config/quickshell/ii/modules/common/widgets/StyledFlickable.qml
@@ -1,6 +1,24 @@
 import QtQuick
 
 Flickable {
+    id: root
     maximumFlickVelocity: 3500
     boundsBehavior: Flickable.DragOverBounds
+
+    property real touchpadScrollFactor: 100
+    property real mouseScrollFactor: 50 
+
+    MouseArea {
+        anchors.fill: parent
+        acceptedButtons: Qt.NoButton
+        onWheel: function(wheelEvent) {
+            var delta = wheelEvent.angleDelta.y / 120;
+            // The angleDelta.y of a touchpad is usually small and continuous, 
+            // while that of a mouse wheel is typically in multiples of ±120.
+            var scrollFactor = Math.abs(wheelEvent.angleDelta.y) >= 120 ? root.mouseScrollFactor : root.touchpadScrollFactor;
+            var targetY = root.contentY - delta * scrollFactor;
+            targetY = Math.max(0, Math.min(targetY, root.contentHeight - root.height));
+            root.contentY = targetY;
+        }
+    }
 }
--- a/.config/quickshell/ii/modules/common/widgets/StyledListView.qml
+++ b/.config/quickshell/ii/modules/common/widgets/StyledListView.qml
@@ -15,6 +15,9 @@ ListView {
     property real dragDistance: 0
     property bool popin: true
 
+    property real touchpadScrollFactor: 100
+    property real mouseScrollFactor: 50
+
     function resetDrag() {
         root.dragIndex = -1
         root.dragDistance = 0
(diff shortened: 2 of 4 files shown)


end-4:
That mouse delta is most likely to be >= 120 is an interesting find. Since this is technically a hack we should add config options

Runzelee:
Totally agree that this is more of a workaround than a clean fix. Appreciate the follow-up commit.

end-4:
I got distracted and went to find neko nipples after testing the booru widget a bit
Anyway, thanks, and I'll merge this now

end-4:
Is it just me, or manual scrolling by changing `contentY` based on `contentHeight` and `height` really unreliable? 
Sometimes i can't scroll all the way to the bottom and have to rely on flicking... This applies to not just this PR but also the PageUp/PageDown scrolling on the left sidebar which is implemented in a similar way

Runzelee:
Could you be a bit more specific about which page won't scroll all the way down? I haven't been able to reproduce this yet.

end-4:
Notice how near the end of the video, the page bounces back up when I try to scroll down

https://github.com/user-attachments/assets/2f8a86ee-6ae7-40a2-b3e3-4bc9e5e0d0f1

Runzelee:
I watched your video closely. Still can't reproduce it on my side, but I have a hunch.
When I wrote the differentiate logic I didn't set `interactive: false` in `StyledListView` because I wanted to keep the old drag-to-scroll feel, and it worked fine for me. But I think there's a race condition here: my code is changing `contentY` at the same time the ListView's own scroll mechanism is kicking in when you scroll really fast, which could explain the bounce-back.

You could try setting `interactive: false` to avoid the conflict, though that would remove drag scrolling. Or, if you're not using a touchpad, just switch back to the old behavior.

That's just a guess though, could be something else entirely. Plus, you mentioned seeing similar behavior outside of this PR too.

PR #1784, opened 2025-08-08
Feature: Hide online models on local ai policy

Saint2999:
## Describe your changes

Added ai policy check during model init

I think it's ready


end-4:
uh thanks

PR #1786, opened 2025-08-08
Run hyprctl directly in HyprlandData service

Alvin-HZ:
## Describe your changes

Changes processes to run hyprctl directly instead of through a bash shell.
Able to be done after fe07298.

yes


end-4 (approved):
thank you

PR #1797, opened 2025-08-10
ServicesConfig now edits hyprlock.conf to change the clock format

lunstia:
## Describe your changes

hyprlock clock shows in 24 hour mode and ignores user's preference. When changing preference, I edit the hyprlock.conf file to only modify the $TIME to $TIME12 and vice versa based on the user choice.

Yes


end-4:
It's not that simple because this file gets overwritten with color generation when you change wallpaper
We'll have to make colors sourced from another file first, then color generation can overwrite that file without touching the file you're modifying for this

lunstia:
> It's not that simple because this file gets overwritten with color generation when you change wallpaper We'll have to make colors sourced from another file first, then color generation can overwrite that file without touching the file you're modifying for this

Ah, I think I found it, yeah makes sense why searching hyprlock.conf didn't bring up any relevant results when making sure there was no conflict. Will look more into it tomorrow

lunstia:
Seems to work just fine now, plus I think it's an extra benefit to make sure matugen doesn't overwrite the entire hyprlock file incase anyone wants to customize it in the future

.config/hypr/hyprlock-colors.conf
end-4, line 1: put this in `~/.config/hypr/hyprlock/colors.conf`

.config/matugen/templates/hyprland/hyprlock-colors.conf
end-4, line 1: additionally remove the previous hyprlock.conf template

end-4:
thanks, i'll merge this now

PR #1806, opened 2025-08-11
scriptsdata: Add `#!/bin/bash`

Ksawlii:
Add `#!/bin/bash` to scripts without .sh or .bash extension (bash syntax highlighting does not work without it).

Ready for merge


end-4:
> `# This is NOT a script for execution, but for loading functions, so NOT need execution permission or shebang.`

so no
you should be able to make your editor use bash highlighting... at least in vscode that's possible

PR #1813, opened 2025-08-13
add vpn toggle to sidebarRight and vpn status icon to bar

hoovad:
## Describe your changes

adds a VPN toggle to sidebarRight and a VPN status icon to the bar
only visible if Proton VPN is installed (by checking if the protonvpn-app command exists)
only disconnecting from the VPN is supported as disconnecting (from a VPN interface created by the Proton VPN app) makes the VPN interface disappear completely and since there can be multiple VPN interfaces

image
image

## Is it ready?

seems to be

### DIFF
--- a/.config/quickshell/ii/modules/bar/BarContent.qml
+++ b/.config/quickshell/ii/modules/bar/BarContent.qml
@@ -414,6 +414,13 @@ Item { // Bar content region
                             color: rightSidebarButton.colText
                         }
                         MaterialSymbol {
+                            Layout.rightMargin: indicatorsRowLayout.realSpacing
+                            text: VPN.materialSymbol
+                            iconSize: Appearance.font.pixelSize.larger
+                            color: rightSidebarButton.colText
+                            visible: VPN.available
+                        }
+                        MaterialSymbol {
                             text: Bluetooth.bluetoothConnected ? "bluetooth_connected" : Bluetooth.bluetoothEnabled ? "bluetooth" : "bluetooth_disabled"
                             iconSize: Appearance.font.pixelSize.larger
                             color: rightSidebarButton.colText
--- a/.config/quickshell/ii/modules/sidebarRight/SidebarRight.qml
+++ b/.config/quickshell/ii/modules/sidebarRight/SidebarRight.qml
@@ -166,6 +166,7 @@ Scope {
                             IdleInhibitor {}
                             EasyEffectsToggle {}
                             CloudflareWarp {}
+                            VPNToggle {}
                         }
 
                         // Center widget group
--- a/.config/quickshell/ii/modules/sidebarRight/quickToggles/VPNToggle.qml
+++ b/.config/quickshell/ii/modules/sidebarRight/quickToggles/VPNToggle.qml
@@ -1,0 +1,32 @@
+import qs.services
+import qs.modules.common
+import qs.modules.common.widgets
+import qs.modules.common.functions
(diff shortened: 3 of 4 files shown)


end-4:
ERROR: Failed to load configuration
 ERROR:   caused by @shell.qml[62:96]: Type Bar unavailable
 ERROR:   caused by @modules/bar/Bar.qml[92:21]: Type BarContent unavailable
 ERROR:   caused by @modules/bar/BarContent.qml[406:5]: Syntax error

end-4:
Actually, doesn't the Proton app already provide a tray icon with "Quick connect" and "Disconnect" options? Wouldn't another indicator on the bar be redundant? 
And the toggle only being able to disconnect doesn't feel right

hoovad:
> Actually, doesn't the Proton app already provide a tray icon with "Quick connect" and "Disconnect" options? Wouldn't another indicator on the bar be redundant?

yes, however with the tray being a dropdown now, you don't see if you're connected or not immediately, and its just nice to have it more integrated in the system

> And the toggle only being able to disconnect doesn't feel right

its a limitation, since there can be multiple VPN interfaces and in the case of the Proton VPN interface, once you disconnect, the interface completely disappears

end-4:
> yes, however with the tray being a dropdown now, you don't see if you're connected or not immediately

it's configurable, you can pin proton's item

0blivi0nis:
Would something like this be possible for tailscale/wireguard?

hoovad:
It already supports wireguard but it will only show up if proton VPN is installed

end-4:
I would not consider a button that only turns off a toggle button
Either we make this somehow be able to turn on or we not have this button

clsty:
I'm a bit curious that if the button could be dynamically hidden when the state is off.

My suggestion is that the toggle-on behavior should be configurable, i.e. running a user defined command. And if user has not configured it, hide the button when not clickable (i.e. when state is off).

hoovad:
hiding the button is dynamically is possible, and configuring a custom command to run is possible as well, not sure what @end-4 thinks about this though

end-4:
A button which you can only click to make it disappear is not a toggle button, that's why I don't like this
Maybe we can have some different place, or an extra island for such buttons

ghost:
> It already supports wireguard but it will only show up if proton VPN is installed

Only if proton VPN is installed? You might want it check for if there is a wireguard configuration OR proton VPN is installed.

bogdan-velicu:
Hi, I'm looking forward for this feature but I have created my VPN configurations manually via NetworkManager, can't we make this more generic, not only for Proton ? Like WireGuard and OpenVPN3 configs aswell ?
Is there any way I could contribute for this feature without creating own fork or something ?

hoovad:
The feature already has WireGuard and OpenVPN configurations in mind. The developer has decided to only enable the status icon if Proton VPN is installed.

ghost:
> The feature already has WireGuard and OpenVPN configurations in mind. The developer has decided to only enable the status icon if Proton VPN is installed.

Well that seems stupid. Why not change the dependency to wireguard or openvpn rather than proton vpn.

hoovad:
ask @end-4

end-4:
I thought you wanted to make a Proton-only toggle so I told you to have that constraint

hoovad:
no, i wanted to made it for wireguard and openvpn itself, its just that i was using proton VPN

PR #1816, opened 2025-08-15
Update workspace_action.sh

ashmartin1810:
This script is kinda an improvement for the workspace_action script of end4. This script allows for relative workspace keybind. The next and prev workspaces allow for relative Workspace support, using the second arg. while the first arg is "ch" or "mv"

i made this script as i found it annoying that when using the traditional keybind, bind = ... , workspace, r+1/r-1, hyprland changed 4 workspaces instead of 1. And upon investigation, it seems that using hyprctl is better. but the original workspace_action didnt work for relative workspace. and was frankly quite obfuscated. So this is a better script in my opinion.

## Describe your changes

Well, i just used simple if/else/elif statements for better maintainablity. Added support for relative workspaces, i guess?

I wouldnt be too sure. this is my first contribution to other's work on github. It might need some feedback/reviewing?

### DIFF
--- a/.config/hypr/hyprland/scripts/workspace_action.sh
+++ b/.config/hypr/hyprland/scripts/workspace_action.sh
@@ -1,2 +1,29 @@
-#!/usr/bin/env bash
-hyprctl dispatch "$1" $(((($(hyprctl activeworkspace -j | jq -r .id) - 1)  / 10) * 10 + $2))
+#!/bin/bash
+##This script is kinda an improvement for the workspace_action script of end4. This script allows for relative workspace keybind. The next and prev workspaces allow for relative-
+##Workspace support, using the second arg. while the first arg is "ch" or "mv"
+
+##Example Usage: bash workspace_action ch next
+
+curr_workspace="$(hyprctl monitors -j | jq -r '.[] | select(.focused) | .activeWorkspace.name')"
+next_workspace="$(echo $curr_workspace +1 | bc)" 
+prev_workspace="$(echo $curr_workspace -1 | bc)"
+
+if [ "$1" == "ch" ]; then
+  if [ "$2" == "next" ]; then
+    hyprctl dispatch workspace $next_workspace
+  elif [ "$2" == "prev" ]; then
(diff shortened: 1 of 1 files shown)


.config/hypr/hyprland/scripts/workspace_action.sh
```diff
@@ -1,2 +1,29 @@
-#!/usr/bin/env bash
-hyprctl dispatch "$1" $(((($(hyprctl activeworkspace -j | jq -r .id) - 1)  / 10) * 10 + $2))
+#!/bin/bash
+##This script is kinda an improvement for the workspace_action script of end4. This script allows for relative workspace keybind. The next and prev workspaces allow for relative-
+##Workspace support, using the second arg. while the first arg is "ch" or "mv"
```
end-4: > is kinda an improvement for the workspace_action script of end4

Not wrong but if you're giving me the script then no need to state it in the final version...

Also, does `#!/bin/bash` work on NixOS? If not or you're unsure just keep it `#!/usr/bin/env bash`

end-4 (commented):
I don't like how we need an extra flag and this locks us to two dispatchers guarded by an if-else (what about `movetoworkspacesilent`?)

Please make the script take `$1` and `$2` as arguments as the original, then check if `$2` is a number (without `+`/`-`/`~`, so `3` counts but `-5` doesn't). If `$2` is a number do the handling like in the original script and otherwise pass `$2` as is to hyprctl

end-4:
also for your use case of

```sh
next_workspace="$(echo $curr_workspace +1 | bc)" 
hyprctl dispatch workspace $next_workspace
```

why not just do `hyprctl dispatch workspace +1`?
`r+1` means the next workspace on the current monitor, while `+1` means the next numerical workspace, can be on any monitor

.config/hypr/hyprland/scripts/workspace_action.sh
ashmartin1810: I am sorry, this is my first pr. i mightve been a little unprofessional. it is fixed now... i believe

PR #1820, opened 2025-08-15
feat(wallpaper selector): add wallpaper selector menu

sinnayuh:
## Describe your changes

Implemented a Wallpaper Changer using the same style and theming as the Overview Module. It utilizes native Qs functions to generate thumbnails for display to the user, eliminating the need for additional scripts to generate thumbnails.

Incorporates lazy-loading to handle cases where a thumbnail hasn't been generated yet (which can occur with some 4K images during reading and resizing). Thumbnails fade in gradually. All my wallpapers are 1080p or 1440p, so they load instantly.

Users can select a new wallpaper using either arrow keys or the mouse. The feature integrates seamlessly with the current system by using switchwall.sh to apply the new wallpaper.

Added a keybind `(Super + Comma)` to toggle the Wallpaper Selection Menu.

Wallpapers are pulled from `~/Pictures/Wallpapers`.

Tested on a fresh install, works as intended.

image

https://github.com/user-attachments/assets/25257438-0fd4-49a0-a3cd-dde2fcdf00fe

### DIFF
--- a/.config/hypr/hyprland/keybinds.conf
+++ b/.config/hypr/hyprland/keybinds.conf
@@ -25,6 +25,7 @@ bindit = ,Super_R, global, quickshell:workspaceNumber # [hidden]
 bindd = Super, V, Clipboard history >> clipboard, global, quickshell:overviewClipboardToggle # Clipboard history >> clipboard
 bindd = Super, Period, Emoji >> clipboard, global, quickshell:overviewEmojiToggle # Emoji >> clipboard
 bindd = Super, Tab, Toggle overview, global, quickshell:overviewToggle # [hidden] Toggle overview/launcher (alt)
+bindd = Super, Comma, Toggle wallpaper overview, global, quickshell:wallpaperOverviewToggle # Wallpaper overview
 bindd = Super, A, Toggle left sidebar, global, quickshell:sidebarLeftToggle # Toggle left sidebar
 bind = Super+Alt, A, global, quickshell:sidebarLeftToggleDetach # [hidden]
 bind = Super, B, global, quickshell:sidebarLeftToggle # [hidden]
--- a/.config/quickshell/ii/modules/wallpaperOverview/WallpaperOverview.qml
+++ b/.config/quickshell/ii/modules/wallpaperOverview/WallpaperOverview.qml
@@ -1,0 +1,244 @@
+import qs
+import qs.services
+import qs.modules.common
+import qs.modules.common.widgets
+import qs.modules.common.functions
+import QtQuick
+import QtQuick.Controls
+import QtQuick.Layouts
+import Quickshell
+import Quickshell.Io
+import Quickshell.Wayland
+import Quickshell.Hyprland
+
+// Fullscreen panel similar to Overview, but shows a grid of wallpaper previews.
+Scope {
+    id: scope
+
+    Variants {
+        id: variants
+        model: Quickshell.screens
+
+                    PanelWindow {
+                id: root
+                required property var modelData
+                readonly property HyprlandMonitor monitor: Hyprland.monitorFor(root.screen)
+                property bool monitorIsFocused: (Hyprland.focusedMonitor?.id == monitor?.id)
+                screen: modelData
+                visible: GlobalStates.wallpaperOverviewOpen && monitorIsFocused
+
+            WlrLayershell.namespace: "quickshell:wallpaper-overview"
+            WlrLayershell.layer: WlrLayer.Overlay
+            WlrLayershell.keyboardFocus: WlrKeyboardFocus.OnDemand
+            color: "transparent"
+
+            anchors { top: true; bottom: true; left: true; right: true }
+
+            ColumnLayout {
+                id: layout
+                anchors.horizontalCenter: parent.horizontalCenter
+                anchors.top: parent.top
+                spacing: 8
+
+                Item { width: 1; height: 1 }
+
+                Rectangle {
+                    id: bg
+                    focus: true
+                    color: Appearance.colors.colLayer0
+                    border.width: 1
+                    border.color: Appearance.colors.colLayer0Border
+                    radius: Appearance.rounding.screenRounding
+
+                    // Compact size for 4 thumbnails per row, 3 rows high
+                    implicitWidth: Math.min(root.width * 0.7, 900)
+                    implicitHeight: Math.min(root.height * 0.6, 500)
+
+                    Keys.onPressed: (event) => {
+                        if (event.key === Qt.Key_Escape) {
+                            GlobalStates.wallpaperOverviewOpen = false
+                            event.accepted = true
+                        } else if (event.key === Qt.Key_Left) {
+                            grid.moveSelection(-1)
+                            event.accepted = true
+                        } else if (event.key === Qt.Key_Right) {
+                            grid.moveSelection(1)
+                            event.accepted = true
+                        } else if (event.key === Qt.Key_Up) {
+                            grid.moveSelection(-grid.columns)
+                            event.accepted = true
+                        } else if (event.key === Qt.Key_Down) {
+                            grid.moveSelection(grid.columns)
+                            event.accepted = true
+                        } else if (event.key === Qt.Key_Return || event.key === Qt.Key_Enter) {
+                            grid.activateCurrent()
+                            event.accepted = true
+                        }
+                    }
+
+                    ColumnLayout {
+                        anchors.fill: parent
+                        anchors.margins: 8
+                        spacing: 8
+
+                        GridView {
+                            id: grid
+                            readonly property int columns: 4  // Fixed to 4 columns
+                            property int currentIndex: 0
+                            readonly property int rows: Math.max(1, Math.ceil(count / columns))
+
+                            Layout.preferredWidth: columns * cellWidth
+                            Layout.alignment: Qt.AlignHCenter
+                            Layout.fillHeight: true
+                            cellWidth: 220
+                            cellHeight: 140
+                            clip: true
+                            interactive: true
+                            keyNavigationWraps: true
+                            boundsBehavior: Flickable.StopAtBounds
+                            
+                            // Performance optimization: cache more delegates for smoother scrolling
+                            cacheBuffer: cellHeight * 2  // Cache 2 extra rows above/below visible area
+                            ScrollBar.horizontal: ScrollBar { 
+                                policy: ScrollBar.AsNeeded
+                                visible: false
+                            }
+                            ScrollBar.vertical: ScrollBar { 
+                                policy: ScrollBar.AsNeeded
+                                visible: false
+                            }
+                            // Back to simple wallpapers array
+                            model: Wallpapers.wallpapers
+                            onModelChanged: currentIndex = 0
+
+                            function moveSelection(delta) {
+                                // Clear all hover states when using keyboard navigation
+                                for (let i = 0; i < count; i++) {
+                                    const item = itemAtIndex(i)
+                                    if (item) {
+                                        item.isHovered = false
+                                    }
+                                }
+                                currentIndex = Math.max(0, Math.min(count - 1, currentIndex + delta))
+                                positionViewAtIndex(currentIndex, GridView.Contain)
+                            }
+                            function activateCurrent() {
+                                const path = Wallpapers.wallpapers[currentIndex]
+                                if (!path) return
+                                GlobalStates.wallpaperOverviewOpen = false
+                                Wallpapers.apply(path)
+                            }
+
+                            delegate: Item {
+                                width: grid.cellWidth
+                                height: grid.cellHeight
+                                property bool isHovered: false
+
+                                Rectangle {
+                                    anchors.fill: parent
+                                    radius: Appearance.rounding.windowRounding
+                                    color: Appearance.colors.colLayer1
+                                    border.width: (index === grid.currentIndex || parent.isHovered) ? 3 : 0
+                                    border.color: Appearance.colors.colSecondary
+                                }
+
+                                Rectangle {
+                                    anchors.fill: parent
+                                    anchors.margins: 8
+                                    color: Appearance.colors.colLayer2
+                                    radius: Appearance.rounding.elementRounding
+                                    
+                                    // Loading placeholder
+                                    Rectangle {
+                                        anchors.centerIn: parent
+                                        width: Math.min(parent.width * 0.4, 32)
+                                        height: Math.min(parent.height * 0.4, 32)
+                                        radius: Appearance.rounding.elementRounding
+                                        color: Appearance.colors.colLayer3
+                                        visible: thumbnailImage.status !== Image.Ready
+                                        
+                                        // Simple loading animation
+                                        opacity: 0.3
+                                        SequentialAnimation on opacity {
+                                            running: parent.visible
+                                            loops: Animation.Infinite
+                                            NumberAnimation { to: 1.0; duration: 800; easing.type: Easing.InOutSine }
+                                            NumberAnimation { to: 0.3; duration: 800; easing.type: Easing.InOutSine }
+                                        }
+                                    }
+
+                                    Image {
+                                        id: thumbnailImage
+                                        anchors.fill: parent
+                                        source: `file://${modelData}`
+                                        fillMode: Image.PreserveAspectCrop
+                                        asynchronous: true
+                                        cache: false
+                                        smooth: true
+                                        
+                                        // Much smaller sourceSize for faster loading - this is key!
+                                        // Using smaller dimensions significantly reduces decode time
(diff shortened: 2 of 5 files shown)


yahya9090:
i used it but for some reason its blank

sinnafuls:
> i used it but for some reason its blank
> 
> 

Weird, and ~/Pictures/wallpapers is populated with images?

yahya9090:
> > i used it but for some reason its blank
> > 
> > 
> 
> Weird, and ~/Pictures/wallpapers is populated with images?

Yeah

sinnafuls:
> > > i used it but for some reason its blank
> > > 
> > > 
> > 
> > Weird, and ~/Pictures/wallpapers is populated with images?
> 
> Yeah

Weird, I just tried on a fresh install again and it works, and another user in the Discord tried it as well and it works?

What is "blank" anyway? The wallpaper when you set it? Or like the actual menu, like it's not populated?

Or are the images there, but just not loaded in correctly? There is a lazy loading function when Qml is still processing the thumbnails, if you open the selection menu and wait a couple seconds does it populate or no?

.config/hypr/hyprland/keybinds.conf
```diff
@@ -25,6 +25,7 @@ bindit = ,Super_R, global, quickshell:workspaceNumber # [hidden]
 bindd = Super, V, Clipboard history >> clipboard, global, quickshell:overviewClipboardToggle # Clipboard history >> clipboard
 bindd = Super, Period, Emoji >> clipboard, global, quickshell:overviewEmojiToggle # Emoji >> clipboard
 bindd = Super, Tab, Toggle overview, global, quickshell:overviewToggle # [hidden] Toggle overview/launcher (alt)
+bindd = Super, Comma, Toggle wallpaper overview, global, quickshell:wallpaperOverviewToggle # Wallpaper overview
```
end-4: We should call this panel `wallpaperSelector` instead of `wallpaperOverview` then have it use the wallpaper switching keybind, and the current direct execution of the script should become a fallback

So make the above `Ctrl+Super+T` and change this line
```ini
bindd = Ctrl+Super, T, Change wallpaper, exec, ~/.config/quickshell/$qsConfig/scripts/colors/switchwall.sh # Change wallpaper
```
into this
```ini
bindd = Ctrl+Super, T, Change wallpaper, exec, qs -c $qsConfig ipc call TEST_ALIVE || ~/.config/quickshell/$qsConfig/scripts/colors/switchwall.sh # Change wallpaper
```

.config/quickshell/ii/modules/wallpaperOverview/WallpaperOverview.qml
```diff
@@ -0,0 +1,430 @@
...
+import Quickshell.Io
+import Quickshell.Wayland
+import Quickshell.Hyprland
+
+Scope {
+    id: scope
+
+    Variants {
+        id: variants
+        model: Quickshell.screens
```
end-4: Wallpaper selector is not something you'd benefit from showing on all monitors so don't use `Variants` here (well, maybe useful later if we have per-monitor wallpaper configuration, but that's not for this PR)

And below you're making only the variant on the focused monitor show, it is better to just bind the `screen` to the focused screen instead

.config/quickshell/ii/modules/wallpaperOverview/WallpaperOverview.qml
```diff
@@ -0,0 +1,430 @@
...
+        PanelWindow {
+            id: root
+            required property var modelData
+            readonly property HyprlandMonitor monitor: Hyprland.monitorFor(root.screen)
+            property bool monitorIsFocused: (Hyprland.focusedMonitor?.id == monitor?.id)
+            screen: modelData
+            visible: GlobalStates.wallpaperOverviewOpen && monitorIsFocused
+            property var filteredWallpapers: Wallpapers.wallpapers
+
+            WlrLayershell.namespace: "quickshell:wallpaper-overview"
```
end-4: use camelCase for the namespace
so `quickshell:wallpaperSelector`

.config/quickshell/ii/modules/wallpaperOverview/WallpaperOverview.qml
```diff
@@ -0,0 +1,430 @@
...
+            WlrLayershell.layer: WlrLayer.Overlay
+            WlrLayershell.keyboardFocus: WlrKeyboardFocus.OnDemand
+            color: "transparent"
+
+            anchors {
+                top: true
+                bottom: true
+                left: true
+                right: true
+            }
```
end-4: But the widget is just at the top? Why make this fullscreen?
... Oh right, the wonky bouncy animation. It needs to be toned down btw

We should somehow set the size to the max the panel can be, not fullscreen... While fullscreen works it takes up more vram and is more likely to crash quickshell

.config/quickshell/ii/modules/wallpaperOverview/WallpaperOverview.qml
```diff
@@ -0,0 +1,430 @@
...
+            ColumnLayout {
+                id: layout
+                anchors.horizontalCenter: parent.horizontalCenter
+                anchors.top: parent.top
+                spacing: 8
+
+                Item {
+                    width: 1
+                    height: 1
+                }
```
end-4: This is pointless. `width` and `height` by convention don't cascade to ancestors. And you should set the `implicitWidth`/`implicitHeight` of the `PanelWindow` to be that of the `ColumnLayout`

If you copied this from my overview widget: okay I should remove it too lol, before I didn't know the above

.config/quickshell/ii/modules/wallpaperOverview/WallpaperOverview.qml
```diff
@@ -0,0 +1,430 @@
...
+                                GlobalStates.wallpaperOverviewOpen = false;
+                            }
+                            event.accepted = true;
+                        }
+                    }
+                }
+
+                Rectangle {
+                    id: bg
+                    focus: true
```
end-4: why `focus: true` here? don't we want focus to be at the text field? 
well, from my testing there is no problem, but i don't see the point of this line

.config/quickshell/ii/modules/wallpaperOverview/WallpaperOverview.qml
```diff
@@ -0,0 +1,430 @@
...
+                            damping: 0.2
+                        }
+                    }
+
+                    Behavior on implicitHeight {
+                        SpringAnimation {
+                            spring: 3
+                            damping: 0.2
+                        }
+                    }
```
end-4: Too wonky, for each Behavior use this instead of a SpringAnimation
```qml
animation: Appearance.animation.elementMove.numberAnimation.createObject(this)
```
_this is also the animation used for the expanding/shrinking of the left sidebar_

.config/quickshell/ii/modules/wallpaperOverview/WallpaperOverview.qml
```diff
@@ -0,0 +1,430 @@
...
+                                        duration: animationCurves.expressiveDefaultSpatialDuration
+                                        easing.bezierCurve: animationCurves.expressiveDefaultSpatial
+                                    }
+                                }
+
+                                Behavior on height {
+                                    NumberAnimation {
+                                        duration: animationCurves.expressiveDefaultSpatialDuration
+                                        easing.bezierCurve: animationCurves.expressiveDefaultSpatial
+                                    }
```
end-4: When do these animate?

.config/quickshell/ii/modules/wallpaperOverview/WallpaperOverview.qml
```diff
@@ -0,0 +1,430 @@
...
+                                        }
+                                    }
+
+                                    Image {
+                                        id: thumbnailImage
+                                        anchors.fill: parent
+                                        source: `file://${modelData}`
+                                        fillMode: Image.PreserveAspectCrop
+                                        asynchronous: true
+                                        cache: false
```
end-4: i wonder if `cache: true` is a good idea

.config/quickshell/ii/modules/wallpaperOverview/WallpaperOverview.qml
```diff
@@ -0,0 +1,430 @@
...
+import Quickshell.Hyprland
+
+Scope {
+    id: scope
+
+    Variants {
+        id: variants
+        model: Quickshell.screens
+
+        PanelWindow {
```
end-4: Since wallpaper selector is a short-lived thing, put it in a Loader

.config/quickshell/ii/modules/wallpaperOverview/WallpaperOverview.qml
```diff
@@ -0,0 +1,430 @@
...
+
+                                        mipmap: false
+
+                                        opacity: status === Image.Ready ? 1 : 0
+                                        Behavior on opacity {
+                                            NumberAnimation {
+                                                duration: 200
+                                                easing.type: Easing.OutCubic
+                                            }
+                                        }
```
end-4:
```
Behavior on opacity {
    animation: Appearance.animation.elementMoveFast.numberAnimation.createObject(this)
}
```

.config/quickshell/ii/modules/wallpaperOverview/WallpaperOverview.qml
```diff
@@ -0,0 +1,430 @@
...
+                function onWallpaperOverviewOpenChanged() {
+                    if (GlobalStates.wallpaperOverviewOpen && monitorIsFocused) {
+                        filterField.forceActiveFocus();
+                    }
+                }
+            }
+        }
+    }
+
+    GlobalShortcut {
```
end-4: Also add an IpcHandler for compatibility (makes it easier for those trying to run the shell under non-hyprland environments)

.config/quickshell/ii/services/Wallpapers.qml
```diff
@@ -0,0 +1,76 @@
...
+    onSearchDirsChanged: reload()
+
+    function apply(path) {
+        if (!path || path.length === 0) return
+        applyProc.command = [
+            "bash", "-c",
+            `${StringUtils.shellSingleQuoteEscape(Directories.wallpaperSwitchScriptPath)} ` +
+            `--image ${StringUtils.shellSingleQuoteEscape(path)}`
+        ]
+        applyProc.running = true
```
end-4: There's `Process::exec` for this https://quickshell.org/docs/master/types/Quickshell.Io/Process/#exec

.config/quickshell/ii/services/Wallpapers.qml
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    // Folder model
+    FolderListModel {
+        id: files
+        nameFilters: extensions.map(ext => `*.${ext}`)
+        showDirs: false
+        showDotAndDotDot: false
```
end-4, line 98: Why not show dirs, dot and dot dot? If we're making something to replace the wallpaper picker we must not sacrifice the (intended) functionality of the old one. Directory navigation must be available.

.config/quickshell/ii/modules/wallpaperOverview/WallpaperOverview.qml
end-4: fe23017a97faec7d5cf44b4f65d64bd67c71fb0e

.config/quickshell/ii/modules/wallpaperOverview/WallpaperOverview.qml
end-4: fe23017a97faec7d5cf44b4f65d64bd67c71fb0e

.config/quickshell/ii/modules/wallpaperOverview/WallpaperOverview.qml
end-4: 8feee4e61a0b06832ae894e499a4ec72c90a4993

.config/quickshell/ii/modules/wallpaperOverview/WallpaperOverview.qml
end-4: 8feee4e61a0b06832ae894e499a4ec72c90a4993

.config/quickshell/ii/modules/wallpaperOverview/WallpaperOverview.qml
end-4: The loading times of the images in subsequent runs are really bad if we do this. However to properly solve this we will need to use thumbnails. Looks like we need to mess with `~/.cache/thumbnails`

end-4: I'm marking this as resolved because I've put it in a loader
Thumbnail is a new todo item

end-4:
- [x] Thumbnail generation according to [freedesktop specs](https://specifications.freedesktop.org/thumbnail-spec/latest/)

.config/hypr/hyprland/keybinds.conf
end-4: 2d8eb163e784b44f9ac39c60da7b41f91eb0bf69

.config/quickshell/ii/modules/wallpaperOverview/WallpaperOverview.qml
end-4: 2d8eb163e784b44f9ac39c60da7b41f91eb0bf69

.config/quickshell/ii/modules/wallpaperOverview/WallpaperOverview.qml
end-4: well, removing this breaks it so imma ignore it for now

.config/quickshell/ii/modules/wallpaperOverview/WallpaperOverview.qml
end-4: nope, we'll use thumbnails to make sure they're fast anyway

.config/quickshell/ii/modules/wallpaperOverview/WallpaperOverview.qml
end-4: 28fe7817b470ff575bf79b8324d821b1107e5066

.config/quickshell/ii/services/Wallpapers.qml
end-4: 42695d9253df2568c181c2fdd6b7d1963591375b

.config/quickshell/ii/modules/wallpaperOverview/WallpaperOverview.qml
end-4: removed in 89e726b5a2d7d22501b79df98ebb9d411ad9b76a

rudraaaaa-05:
Hey, How can i make this work with video wallpapers? When i apply the video wallpaper it just gets removed instantly.
Also a lot of image (jpg) thumbnails are missing, the image resolution is 1920x1080, but i'm sure that doesn't matter

Thumbnail Issue:
image
(there's also like a weird overlap of the pictures in the bottom while scrolling)

end-4:
Thumbnail generation is to be added and video files are not yet supported
It's a PR for a reason

0blivi0nis:
damn this has turned into something so much better than sleex design, the custom file picker is 🔥

end-4:
The last problem to solve is thumbnail maintenance: If image changes, the thumbnail should be regenerated
The picker currently looks like this 
image

end-4:
I realized just generating the image is still not fully spec-compliant... anyone got an idea of what tool we could use?

0blivi0nis:
Would you have to do something similar to what i was doing with the other wallpaper selector? 

Also, currently, if you switch folders, and close the selector then open it again it opens to the previous folder instead of opening the Wallpapers folder. I think it should function this way

end-4:
> Would you have to do something similar to what i was doing with the other wallpaper selector?

The solution in your PR is even further away from being spec-compliant than what we have right now, but if you can make line 57 of `ThumbnailImage.qml` a script that properly sets metadata (and using it to properly regenerate the thumbnail when necessary) according to the spec (see the link above) that would be awesome

Ricing is normally "It's my machine I can do weird shit" but this is rice is not going to run on only one single person's machine

end-4:
This one is interesting: https://github.com/difference-engine/thumbnail-generator-ubuntu

It works, but only generates "normal" size thumbnails and there doesn't seem to be a flag 

Edit: I managed to add sizes in https://github.com/difference-engine/thumbnail-generator-ubuntu/pull/11 ... I hope they still accept PRs after 5 years

end-4:
Since the script is small and the maintainers seem inactive to accept my PR I decided to just copy it over.
When it gets merged and the python package gets updated we can use it, but for now this copying works and is not a big problem

So we're ready :tada:

It would be nice to have indeterminate progressbar while thumbnails are being generated, but for now this works

lash0000:
It got merged? yay

Issue #1825, opened 2025-08-16
Speed Meter


sayeed205:
### What would you like to be added?

I want to add a speed meter in the bar that shows current internet speed

### How will it help?

It will help to to check my internet speed at the moment at a glance

### Extra info

The preview will look like this and clicking it will cycle through diffrent modes
modes: total, download only, upload only, both side by side
<img width="1131" height="48" alt="Image" src="https://github.com/user-attachments/assets/c2dcf573-f9aa-4bfb-a24a-c4fc27789806" />

M-0u554m4:
is there any update on this feature I have the same request ?

PR #1826, opened 2025-08-16
feat(bar): add network speed display to bar

sayeed205:
* Introduced `NetworkSpeed.qml` to display current network speed.
* Configurable display modes: total, download, upload, or both.
* Added toggle in `InterfaceConfig.qml` to enable/disable network speed display.
* Updated `Config.qml` to support network speed configuration.
* Integrated network speed component into `BarContent.qml`.

### DIFF
--- a/.config/quickshell/ii/modules/bar/NetworkSpeed.qml
+++ b/.config/quickshell/ii/modules/bar/NetworkSpeed.qml
@@ -1,0 +1,174 @@
+import qs.modules.common
+import qs.modules.common.widgets
+import qs.services
+import qs
+import QtQuick
+import QtQuick.Layouts
+
+Item {
+    id: root
+    implicitWidth: networkLayout.implicitWidth + 16
+    implicitHeight: Appearance.sizes.barHeight
+
+    // Display modes: 0=total, 1=download, 2=upload, 3=both
+    property int displayMode: 0
+
+    // Helper function to format network speed
+    function formatSpeed(bytesPerSecond) {
+        if (bytesPerSecond < 1024) {
+            return bytesPerSecond.toFixed(0) + " B/s";
+        } else if (bytesPerSecond < 1024 * 1024) {
+            return (bytesPerSecond / 1024).toFixed(1) + " KB/s";
+        } else if (bytesPerSecond < 1024 * 1024 * 1024) {
+            return (bytesPerSecond / (1024 * 1024)).toFixed(1) + " MB/s";
+        } else {
+            return (bytesPerSecond / (1024 * 1024 * 1024)).toFixed(1) + " GB/s";
+        }
+    }
+
+    function getDisplayText() {
+        var downloadSpeed = ResourceUsage.networkDownloadSpeed;
+        var uploadSpeed = ResourceUsage.networkUploadSpeed;
+        var totalSpeed = downloadSpeed + uploadSpeed;
+
+        switch (displayMode) {
+        case 0 // Total speed
+        :
+            return formatSpeed(totalSpeed);
+        case 1 // Download only
+        :
+            return "↓ " + formatSpeed(downloadSpeed);
+        case 2 // Upload only
+        :
+            return "↑ " + formatSpeed(uploadSpeed);
+        case 3 // Both (dual row)
+        :
+            return ""; // Handled separately
+        default:
+            return formatSpeed(totalSpeed);
+        }
+    }
+
+    RowLayout {
+        id: networkLayout
+        anchors.centerIn: parent
+        spacing: 6
+
+        MaterialSymbol {
+            text: "network_check"
+            iconSize: Appearance.font.pixelSize.normal
+            color: Appearance.colors.colOnLayer1
+        }
+
+        // Single line display (modes 0, 1, 2)
+        StyledText {
+            id: singleLineText
+            visible: displayMode !== 3
+            font.pixelSize: Appearance.font.pixelSize.small
+            color: Appearance.colors.colOnLayer1
+            text: getDisplayText()
+        }
+
+        // Side by side display (mode 3)
+        RowLayout {
+            visible: displayMode === 3
+            spacing: 4
+
+            StyledText {
+                id: downloadText
+                font.pixelSize: Appearance.font.pixelSize.small
+                color: Appearance.colors.colOnLayer1
+                text: "↓ " + formatSpeed(ResourceUsage.networkDownloadSpeed)
+            }
+
+            StyledText {
+                id: uploadText
+                font.pixelSize: Appearance.font.pixelSize.small
+                color: Appearance.colors.colOnLayer1
+                text: "↑ " + formatSpeed(ResourceUsage.networkUploadSpeed)
+            }
+        }
+    }
+
+    MouseArea {
+        id: mouseArea
+        anchors.fill: parent
+        hoverEnabled: true
+        acceptedButtons: Qt.LeftButton
+        onClicked: {
+            displayMode = (displayMode + 1) % 4;
+        }
+    }
+
+    StyledPopup {
+        hoverTarget: mouseArea
+
+        ColumnLayout {
+            anchors.centerIn: parent
+            spacing: 4
+
+            // Header
+            RowLayout {
+                spacing: 5
+
+                MaterialSymbol {
+                    fill: 0
+                    font.weight: Font.Medium
+                    text: "network_check"
+                    iconSize: Appearance.font.pixelSize.large
+                    color: Appearance.colors.colOnSurfaceVariant
+                }
+
+                StyledText {
+                    text: Translation.tr("Network Speed")
+                    font {
+                        weight: Font.Medium
+                        pixelSize: Appearance.font.pixelSize.normal
+                    }
+                    color: Appearance.colors.colOnSurfaceVariant
+                }
+            }
+
+            // Speed info
+            RowLayout {
+                spacing: 5
+
+                MaterialSymbol {
+                    text: "download"
+                    color: Appearance.colors.colOnSurfaceVariant
+                    iconSize: Appearance.font.pixelSize.large
+                }
+                StyledText {
+                    text: Translation.tr("Download:")
+                    color: Appearance.colors.colOnSurfaceVariant
+                }
+                StyledText {
+                    Layout.fillWidth: true
+                    horizontalAlignment: Text.AlignRight
+                    color: Appearance.colors.colOnSurfaceVariant
+                    text: formatSpeed(ResourceUsage.networkDownloadSpeed)
+                }
+            }
+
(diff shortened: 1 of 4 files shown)


sayeed205:
Closes #1825

sayeed205:
Preview:
total:
image
download:
image
upload:
image
both:
image

CodeRiderSun:
How did u get NetworkDownloadSpeed and networkUploadSpeed?

.config/quickshell/ii/modules/bar/NetworkSpeed.qml
```diff
@@ -0,0 +1,174 @@
...
+    function formatSpeed(bytesPerSecond) {
+        if (bytesPerSecond < 1024) {
+            return bytesPerSecond.toFixed(0) + " B/s";
+        } else if (bytesPerSecond < 1024 * 1024) {
+            return (bytesPerSecond / 1024).toFixed(1) + " KB/s";
+        } else if (bytesPerSecond < 1024 * 1024 * 1024) {
+            return (bytesPerSecond / (1024 * 1024)).toFixed(1) + " MB/s";
+        } else {
+            return (bytesPerSecond / (1024 * 1024 * 1024)).toFixed(1) + " GB/s";
+        }
```
end-4, line 26: If you base this on 1024 it should be KiB, MiB, GiB

end-4 (commented):
We're running out of bar space. I'm not really into the idea of adding extra merely nerd-neuron-activating information that's not really actionable...

If we have network speed it should be a tooltip under the wifi icon

Also like above, I don't see the network speed retrieval implemented?

nyx-4:
The top bar has accumulated clutter over time, the new UI is _not so good_. I think we should first clean up some stuff. 


## Annoyances:
1. The battery is completely out-of-place.
2. The weather thing:
    - is overflowing into battery
    - has different background
    - do not respect overall margin/padding/spacing (whatever jargon...)
    - kind-of overflows out of bar (at-least gives the illusion of overflowing)
3. The percent is missing from system icons, kinda annoying.
4. Using pi-chart for RAM/CPU usage, YIKES!>

image

image

image


## More annoyances:
1. The pop-up of Weather thing and resources thing are very different, they should better be unified.
2. The CPU has only one info row, should have three for consistency.

image

image

sansmoraxz:
> The CPU has only one info row, should have three for consistency.

Maybe include current clock & power draw.

end-4:
> The top bar has accumulated clutter over time, the new UI is _not so good_. I think we should first clean up some stuff.
> ...

## Annoyances
1. The redesign from line circular progresses to filled pills and filled circular progresses (or pie charts) was to make the top bar less "noisy" (so less distracting)... I guess the pill isn't super consistent but it's lonely by itself so also using a filled circular progress would make the left side feel more "heavy" than the right side...
2. The weather thing barely works, I don't really care about it so I didn't give it much thought. Where do you think I should put it?... Oh, maybe we can have it replace the util buttons, since they're fillers anyway
3. This is part of "reducing noise" from the bar
4. They're just circular progresses but filled. Reason: as in (1)

## More annoyances
1. It doesn't look that bad, it still uses the same colors/fonts and have the same feel... I can make the resources tooltip fancier in a different way and you wouldn't complain right? 🤔
2. What other measurements do you suggest we put there? We display info for the sake of providing information, not just for looking nice. 

> Maybe include current clock & power draw.

Power draw is already on battery tooltip, but current clock maybe okay

sayeed205:
> We're running out of bar space

That why it is toggleable if anyone want it or not, just like weather and we can make it off by default

moechakucha:
> That why it is toggleable if anyone want it or not, just like weather and we can make it off by default

i think the biggest problem is that it simply looks ass for people who prefer to enable such options to get as much info as possible directly from their shell

the key is to make it just a possible choice to have all the thingy displayed on the shell and for now the setup simply doesnt want to work like that if it's the case

rudraaaaa-05:
image

how can i fix this

sayeed205:
> > That why it is toggleable if anyone want it or not, just like weather and we can make it off by default
> 
> i think the biggest problem is that it simply looks ass for people who prefer to enable such options to get as much info as possible directly from their shell
> 
> the key is to make it just a possible choice to have all the thingy displayed on the shell and for now the setup simply doesnt want to work like that if it's the case

Okay understood 
Any suggestions though?

yahya9090:
> image
> 
> how can i fix this

yeah same here

sansmoraxz:
> > how can i fix this
> 
> yeah same here

Some way to configure the interface name may be necessary. These ppl are probably having some virtual network interface selected (like veth from docker).

yahya9090:
> > > how can i fix this
> > 
> > yeah same here
> 
> Some way to configure the interface name may be necessary. These ppl are probably having some virtual network interface selected (like veth from docker).
> 
> 
> 

I see

.config/quickshell/ii/modules/bar/NetworkSpeed.qml
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    function getDisplayText() {
+        var downloadSpeed = ResourceUsage.networkDownloadSpeed;
+        var uploadSpeed = ResourceUsage.networkUploadSpeed;
```
xXFiEsTaDeAmOnXx, line 31: > how can i fix this
>
> yeah same here

The `NaN` values in the network speed appear because the properties `networkDownloadSpeed` and `networkUploadSpeed` from the singleton `ResourceUsage` are being used, but the business logic inside `ResourceUsage.qml` is not implemented yet.

xXFiEsTaDeAmOnXx: Apply this patch using `git am 0001-Fixed-NaN-network-speeds.patch` and it should be fixed. 

<details>
<summary>0001-Fixed-NaN-network-speeds.patch</summary>

```
From 21ef466117c71c2cc4d591a1198be62c02812a55 Mon Sep 17 00:00:00 2001
From: xXFiEsTaDeAmOnXx <niklas.kleiser@web.de>
Date: Sat, 23 Aug 2025 10:05:19 +0200
Subject: [PATCH] Fixed NaN network speeds

---
 .../quickshell/ii/modules/common/Config.qml   |  3 +-
 .../quickshell/ii/services/ResourceUsage.qml  | 44 ++++++++++++++++++-
 2 files changed, 44 insertions(+), 3 deletions(-)

diff --git a/.config/quickshell/ii/modules/common/Config.qml b/.config/quickshell/ii/modules/common/Config.qml
index 5d9b9f88..8aa4c19a 100644
--- a/.config/quickshell/ii/modules/common/Config.qml
+++ b/.config/quickshell/ii/modules/common/Config.qml
@@ -176,7 +176,8 @@ Singleton {
                     property int fetchInterval: 10 // minutes
                 }
                 property JsonObject networkSpeed: JsonObject {
-                    property bool enable: false
+                  property bool enable: true
+                  property string interf: ""
                 }
             }
 
diff --git a/.config/quickshell/ii/services/ResourceUsage.qml b/.config/quickshell/ii/services/ResourceUsage.qml
index 65052840..6825a12e 100644
--- a/.config/quickshell/ii/services/ResourceUsage.qml
+++ b/.config/quickshell/ii/services/ResourceUsage.qml
@@ -21,6 +21,10 @@ Singleton {
     property double cpuUsage: 0
     property var previousCpuStats
 
+    property double networkDownloadSpeed: 0
+    property double networkUploadSpeed: 0
+    property var previusNetworkStats
+
 	Timer {
 		interval: 1
         running: true 
@@ -29,6 +33,8 @@ Singleton {
             // Reload files
             fileMeminfo.reload()
             fileStat.reload()
+            fileTxStat.reload()
+            fileRxStat.reload()
 
             // Parse memory and swap usage
             const textMeminfo = fileMeminfo.text()
@@ -52,11 +58,45 @@ Singleton {
                 }
 
                 previousCpuStats = { total, idle }
-            }
+              }
+
+               
+              // Parse Network usage
+              const tx = Number(fileTxStat.text())
+              const rx = Number(fileRxStat.text()) 
+              if(previusNetworkStats){
+                const txDiff = tx - previusNetworkStats.tx 
+                const rxDiff = rx - previusNetworkStats.rx
+
+                networkDownloadSpeed = rxDiff / 1 //second
+                networkUploadSpeed = txDiff / 1 //second
+
+              }
+
+            previusNetworkStats = {tx, rx}
             interval = Config.options?.resources?.updateInterval ?? 3000
         }
 	}
 
 	FileView { id: fileMeminfo; path: "/proc/meminfo" }
-    FileView { id: fileStat; path: "/proc/stat" }
+  FileView { id: fileStat; path: "/proc/stat" }
+  FileView { id: fileTxStat; path: "/sys/class/net/" +  Config.options.bar.networkSpeed.interf.trim() + "/statistics/tx_bytes" }
+  FileView { id: fileRxStat; path: "/sys/class/net/"+ Config.options.bar.networkSpeed.interf.trim() + "/statistics/rx_bytes" }
+
+  Process { // use first ustable interface if interface is not set in config
+    id: interfaceProc
+    command: ["bash", "-c", "ls /sys/class/net/ | grep -vE '^(lo|br|docker|vir)' | head -n 1"] //exclude some bridge and virtual network interfaces
+    running: true
+
+    stdout: StdioCollector {
+      onStreamFinished: {
+        if(Config.options.bar.networkSpeed.interf== ""){
+            
+          Config.options.bar.networkSpeed.interf = this.text
+          print(this.text)
+        }
+      }
+    }
+  }
+
 }
-- 
2.50.1
```
</details>

It will also select a suitable network interface for you.

You may need to select the correct interface in `Config.options.bar.networkSpeed.interf` if the correct interface is not selected.

Preview:
image

Abhi0808raj:
I had created and added My own Network Speed Indicator and Adjusted my Top bar. Here's My ss
image


image

TowhidMolla-Dev:
> I had created and added My own Network Speed Indicator and Adjusted my Top bar. Here's My ss

How did you do that?can you share the code

0blivi0nis:
> I had created and added My own Network Speed Indicator and Adjusted my Top bar. Here's My ss image
> 
> image

This looks better imo. Maybe add a scrolling widget that lets you scroll and replace only that widget. 

For example, scroll on the media widget and it shows the network speed but keeps the resources where they were. Only replaces the media widget.  A fork of the ags version did this and honestly was one of the best features from that fork that I copied.

pragnyanramtha:
@sayeed205 does this only work for Wi-Fi or Ethernet too ?

M-0u554m4:
@sayeed205  is there any update on this feature. is it planed to be merged to main or should I just integrate it into my setup manually?

Mohd-Saaad:
@mahdjourOussama  were you able to implement it? I found that its read only folder I can't create new file.

PR #1832, opened 2025-08-17
Update config.fish

VLOD-ZDOV:
## Describe your changes

!! and !$ in fish


end-4:
if it's available as a plugin just install it as a plugin?

PR #1841, opened 2025-08-17
Make KDE file dialogue always spawn in reasonable size

nonetrix:
## Describe your changes

I always make this change to my config because the KDE file picker acts strangely under anything that isn't KDE and likes to spawn in super small window sometimes and not remember resizes etc.

This should be fine as is I'd think


end-4:
I don't get any problem but this is more comfortable iI guess

nonetrix:
I found out the issue is with QT apps not using xdg-portal, I confused the two because they look identical outside of the QT one being more buggy I guess. I don't see a reason to keep this, so going make a PR to revert it, sorry and my bad

PR #1846, opened 2025-08-18
Addresses improper env substitution for terminal launch

sansmoraxz:
Escape terminal env for keybinds

## Describe your changes

Addresses improper env substitution raised in: #1753

Yes


yasink21:
yes, I can confirm that now correct terminal is starting

PR #1850, opened 2025-08-19
Feat: Improved Workspace_action.sh and added relative workspace support.

ashmartin1810:
## Problem
The previous script didnt take into account for relative workspaces, and my previous pull request was... bad. 

## Fix
Added smart branching to detect if the target is a literal number(1,2,90), or a relative number(+1, -20) or a string target. ## 

## Bugs
None. ive been stress testing it since the last 2 
Yes, i believe so.


end-4:
I still don't understand why you need this, but it works

Issue #1860, opened 2025-08-21
Stop Local AI from running continously


reakjra:
### What would you like to be added?

I don't know if I'm missing out something about it or idk, but, if you run a local AI model through the sidebar, even after clearing the chat/changing model, it won't stop running, so, it will continously take up VRAM and RAM usage until you manually stop with let's say `ollama stop <model>`. Maybe a /stop command would fix that? 

### How will it help?

Prevent useless resources usage

xXFiEsTaDeAmOnXx:
The amount of time a model stays loaded in memory can be configured using the [`OLLAMA_KEEP_ALIVE`](https://pkg.go.dev/github.com/ollama/ollama/envconfig#KeepAlive) environment variable. By default, this value is set to **5 minutes**.  

You can check how long a model is loaded with the following command:

```bash
ollama ps

NAME           ID              SIZE     PROCESSOR          CONTEXT    UNTIL              
gpt-oss:20b    f2b8351c629c    14 GB    24%/76% CPU/GPU    4096       4 minutes from now
```
The `UNTIL` column indicates how long the model will remain in memory. In this case it is 4 minutes from now.

You can also set the keep-alive time for a specific request directly using a curl command:

```bash
curl http://localhost:11434/api/chat \
            -d '{
          "model": "gpt-oss:20b",
          "keep_alive": "1h",
          "messages": [{"role":"user","content":"hello"}]
        }'
``` 

This will keep the model in memory for **1 hour**.

This works only with the Ollama native endpoint (`/api/chat`) and not with the OpenAI-compatible endpoint (`/v1/chat/completions`).  Currently the  OpenAI-compatible API is used which does not allow this attribute.


You could also add a configurable value in `Config.qml` to adjust the keep_alive time per request and implement a specific `OllamaAIAPIStragegy.qml` for that.  This can be useful for systems where you don’t have direct access to set environment variables, such as when using an external Ollama server hosted by a university or other organization.

reakjra:
Oh thanks that is useful! Tho, I guess that kinda doesn't help with consistency? 'Cause, it would mean I have still to wait the used timer + a too low one would make it impossible to have a conversation with the LLM without it loading off and on from the memory. So I personally think a manual /stop command that can just instantly stop it would be less a burden, maybe.

xXFiEsTaDeAmOnXx:
Yeah, a stop button would be great 👍

PR #1880, opened 2025-08-26
Added ai stop button

xXFiEsTaDeAmOnXx:
## Describe your changes
As requested in [issue #1860](https://github.com/end-4/dots-hyprland/issues/1860), this adds a stop button to stop the AI while generating.

If you are using a local AI, this will use `ollama stop` to unload the model from memory.

The stop button is dynamically shown while a message is generating. After a stop you can still send messages but the local model needs to be moved to memory again.

| Preview when writing | Preview when generating | Preview when stopped |
|---------|---------|-------------------|
|image|Pasted image|Pasted image (2)|

 

We might need to discuss design behavior and button behavior.


reakjra:
This is actual perfect, tho, maybe the stop button should be always clickable when the model is on memory? Like, it get swapped back to the send button if you type something, but if its empty maybe the stop one should persist so that you can just load it off with that instead of having to write a message and stop the AI while it is sending the output?

xXFiEsTaDeAmOnXx:
Okay, now it should also be visible if it is still loaded after a run.

carabistouflette:
@xXFiEsTaDeAmOnXx seems to have some conflict with the repo

RandomlyNamed:
What if a user isn't running ollama? For example I use llama.cpp instead of ollama for local models. Will this work in my case?

PR #1886, opened 2025-08-28
bar: add workspace number mapping

kirisaki-vk:
## Describe your changes
Thanks for this awesome dotfile !!!
This PR adds a mapping to workspace numbers so users can assign custom characters to represent a workspace.
In my case, I use it to represent my worspace as kanji like shown
image

Yes, it's ready. It's just a small change and it's my first contribution to this project

### DIFF
--- a/.config/quickshell/ii/modules/bar/Workspaces.qml
+++ b/.config/quickshell/ii/modules/bar/Workspaces.qml
@@ -216,7 +216,10 @@ Item {
                         anchors.centerIn: parent
                         horizontalAlignment: Text.AlignHCenter
                         verticalAlignment: Text.AlignVCenter
-                        font.pixelSize: Appearance.font.pixelSize.small - ((text.length - 1) * (text !== "10") * 2)
+                        font {
+                            pixelSize: Appearance.font.pixelSize.small - ((text.length - 1) * (text !== "10") * 2)
+                            family: Config.options?.bar.workspaces.useNerdFont ? Appearance.font.family.iconNerd : Appearance.font.family.main
+                        }
                         text: Config.options?.bar.workspaces.numberMap[button.workspaceValue - 1] || button.workspaceValue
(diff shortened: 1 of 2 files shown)


end-4:
We should use an object. What if someone has a workspace 9999 kink?

kirisaki-vk:
At the start I thought of that but I don't know how to make a dynamic object like a Map in qml. I'm not really familiar with qml but I just know the basic

kirisaki-vk:
If I do 
```qml
                property JsonObject workspaces: JsonObject {
                    property bool monochromeIcons: true
                    property int shown: 10
                    property bool showAppIcons: true
                    property bool alwaysShowNumbers: false
                    property int showNumberDelay: 300 // milliseconds
                    property JsonObject numberMapping: JsonObject {}
                }
```
Will it be deserialized correctly ?

end-4:
Use a `var` (see [Quickshell docs page on JsonAdapter](https://quickshell.org/docs/v0.1.0/types/Quickshell.Io/JsonAdapter/))

So have this (we leave some values that effectively don't do anything to demonstrate how they should be declared)
```qml
property var numberMap: { 
  "1": "1" ,
  "2": "2"
}
```

kirisaki-vk:
I tried to make it like that but it does make Quickshell crash, I don't know why

end-4:
oh right, qml shenanigans

we gotta wrap brackets like this `({ ... })`

kirisaki-vk:
Nope
Seems like it works fine when I don't put `numberMapping` in the config.json file. 
But it crash when I try to put any values in it, even those default

end-4:
Indeed... I guess we can go with the array then
Though the name still should be changed to numberMap

end-4:
I added a nerd font option because there definitely will be people who like fancy icons
Thanks for the PR, I'll merge this now

PR #1891, opened 2025-08-29
lock: use default pam config and start without password

### LINKED ISSUE #1800: Fingerprint (fprint) and Face unlock (Howdy) support in lock screen

this is probably doable by creating or modifying the pam file used but i also want the thing in like windows where they have the icon to signify that it's currently reading the fingerprint or face etc

### How will it help?

_No response_

### Extra info

_No response_

moetayuko:
Fixes #1800

## Describe your changes

use default pam config and start without password

ready

### DIFF
--- a/.config/quickshell/ii/modules/lock/LockContext.qml
+++ b/.config/quickshell/ii/modules/lock/LockContext.qml
@@ -33,20 +33,12 @@ Scope {
     }
 
     function tryUnlock() {
-        if (currentText === "") return;
-
         root.unlockInProgress = true;
         pam.start();
     }
 
     PamContext {
         id: pam
-
-        // Its best to have a custom pam config for quickshell, as the system one
-        // might not be what your interface expects, and break in some way.
-        // This particular example only supports passwords.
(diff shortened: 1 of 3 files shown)


.config/quickshell/ii/modules/lock/LockContext.qml
```diff
@@ -33,21 +33,13 @@ Scope {
...
 
     PamContext {
         id: pam
 
-        // Its best to have a custom pam config for quickshell, as the system one
-        // might not be what your interface expects, and break in some way.
-        // This particular example only supports passwords.
-        configDirectory: "pam"
-        config: "password.conf"
-
```
end-4, line 50: No problem on my end without these lines, but I guess this can only be a problem if the user changes their (defaul somewhere else) config somehow?

moetayuko: it could be, but very unlikely. users of custom pam modules (e.g. howdy) tend to directly change the system pam config to make use of them in sudo and wider authentication scenarios

PR #1906, opened 2025-08-31
tray: use more descriptive text for tooltip if available

moetayuko:
## Describe your changes

id is not designed for frontend presentation

ready


end-4:
The id still has to be shown so we know what name to put in the pinned list

.config/quickshell/ii/modules/bar/SysTrayItem.qml
```diff
@@ -72,7 +72,8 @@ MouseArea {
 
     StyledToolTip {
         content: {
-            let c = root.item.id
+            let c = root.item.tooltipTitle.length > 0 ? root.item.tooltipTitle
+                    : (root.item.title.length > 0 ? root.item.title : root.item.id);
```
end-4:
```qml
let c = root.item.id;
if (root.item.tooltipTitle.length > 0) c += " • " + root.item.tooltipTitle
if (root.item.tooltipDescription.length > 0) c += "\n" + root.item.tooltipDescription
```

end-4 (changes requested):

moetayuko:
> The id still has to be shown so we know what name to put in the pinned list

are there other approaches to workaround this? the motivation of this change is that id is "ugly". for instance, `id` for NetworkManager applet is `nm-applet` and `title` is `Network`. Obviously, `id` is for internal use and not expected to be shown to end users.

end-4:
Maybe we can make middle-clicking an entry toggle its pinning state? Then we can not show the id completely

moetayuko:
> Maybe we can make middle-clicking an entry toggle its pinning state? Then we can not show the id completely

some systray impl use middle-clicking for [`SecondaryActivate`](https://www.freedesktop.org/wiki/Specifications/StatusNotifierItem/StatusNotifierItem/), idk how it behaves here.

moetayuko:
What about having another config item to toggle id? The tooltip is shown `Title [id]` or whatever when enabled.

The toggle should be false by default. User can temporarily enable it to figure out the id to fill the pinned list

Another option is to print the id to console when adding, so user can run qs from thermal to get the ids.

end-4:
> some systray impl use middle-clicking for [`SecondaryActivate`](https://www.freedesktop.org/wiki/Specifications/StatusNotifierItem/StatusNotifierItem/), idk how it behaves here.

It's actually not implemented here because I wasn't aware, but I should

> What about having another config item to toggle id? The tooltip is shown `Title [id]` or whatever when enabled.

Sounds good to me, let's go with that

moetayuko:
> > What about having another config item to toggle id? The tooltip is shown `Title [id]` or whatever when enabled.
> 
> Sounds good to me, let's go with that

done

.config/quickshell/ii/modules/bar/SysTrayItem.qml
```diff
@@ -29,6 +29,12 @@ MouseArea {
         }
         event.accepted = true;
     }
+    onEntered: {
+        tooltip.content = item.tooltipTitle.length > 0 ? item.tooltipTitle
+                : (item.title.length > 0 ? item.title : item.id);
+        if (item.tooltipDescription.length > 0) tooltip.content += " • " + item.tooltipDescription;
+        if (Config.options.bar.tray.showIconId) tooltip.content += "\n[" + item.id + "]";
+    }
```
end-4: why are we using this instead of `content` in the tooltip?

.config/quickshell/ii/modules/common/Config.qml
```diff
@@ -172,6 +172,7 @@ Singleton {
                 }
                 property JsonObject tray: JsonObject {
                     property bool monochromeIcons: true
+                    property bool showIconId: false
```
end-4: kind of a nitpick but i think `showItemId` would be more precise (id and icon are different props)

.config/quickshell/ii/modules/bar/SysTrayItem.qml
moetayuko: `content` seems to be assigned once when initialized and never updated later, but title and description can be changed by the app at any time

end-4: That shouldn't happen but is indeed the case... fine I guess

PR #1910, opened 2025-08-31
Customize harmony values for colors from wallpaper

sansmoraxz:
## Describe your changes

Adjust terminal harmony

Yes. We can adjust by editing the config json to override harmony values

### DIFF
--- a/.config/quickshell/ii/modules/common/Config.qml
+++ b/.config/quickshell/ii/modules/common/Config.qml
@@ -91,6 +91,11 @@ Singleton {
                     property bool enableAppsAndShell: true
                     property bool enableQtApps: true
                     property bool enableTerminal: true
+                    property JsonObject generationProps: JsonObject {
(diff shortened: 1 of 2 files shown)


.config/quickshell/ii/modules/common/Config.qml
```diff
@@ -91,6 +91,11 @@ Singleton {
                     property bool enableAppsAndShell: true
                     property bool enableQtApps: true
                     property bool enableTerminal: true
+                    property JsonObject generationProps: JsonObject {
+                        property real harmony: 0.5
+                        property real harmonizeThreshold: 100
+                        property real termFgBoost: 0.35
+                    }
```
end-4: - The default harmony is 0.8
- We should make it clear these properties are for the terminal only. so something like `terminalGenerationProps`

end-4:
uhhh also thanks

PR #1921, opened 2025-09-02
improve system tray visibility handling

amirarfan:
## Describe your changes

This change improves how the system tray popup behaves. It fixes a bug where the popup would remain visible even after the main bar was hidden. The tray now automatically closes when the bar disappears, preventing orphaned popups.

Ready for review.


end-4:
This makes it impossible to use the overflow menu when bar is set to automatically hide

end-4:
We now have click-outside-to-close which I think is good enough

Issue #1940, opened 2025-09-05
popupBackgroundMargin is non-existent property


BeshoyEhab:
### Step 1. Before you submit

- [x] I have read the [Troubleshooting](https://end-4.github.io/dots-hyprland-wiki/en/ii-qs/04troubleshooting/) and [Usage](https://end-4.github.io/dots-hyprland-wiki/en/ii-qs/02usage/) pages.
- [x] I've successfully updated to the latest version following the [guidance](https://end-4.github.io/dots-hyprland-wiki/en/ii-qs/01setup/#updating).
- [x] I've successfully updated the system packages to the latest.
- [ ] I've ticked the checkboxes without reading their contents

### Step 2. Quick diagnose info

<details><summary>Quick diagnose</summary>

```
<!-- Run `./diagnose` inside the repo, and paste the result here! -->
/========================\
| Checking git repo info |
\========================/

[===diagnose===] git remote get-url origin
https://github.com/end-4/dots-hyprland.git
[---SUCCESS---]

[===diagnose===] git rev-parse HEAD
8360d4d589d8641085786cd851d309b247509daf
[---SUCCESS---]

/=================\
| Checking distro |
\=================/

[===diagnose===] ii_check_distro
LSB Version:	n/a
Distributor ID:	Arch
Description:	Arch Linux
Release:	rolling
Codename:	n/a
[---SUCCESS---]

/====================\
| Checking variables |
\====================/

[===diagnose===] declare -p XDG_CACHE_HOME
declare -x XDG_CACHE_HOME="/home/Bisho/.cache"
[---SUCCESS---]

[===diagnose===] declare -p XDG_CONFIG_HOME
declare -x XDG_CONFIG_HOME="/home/Bisho/.config"
[---SUCCESS---]

[===diagnose===] declare -p XDG_DATA_HOME
declare -x XDG_DATA_HOME="/home/Bisho/.local/share"
[---SUCCESS---]

[===diagnose===] declare -p XDG_STATE_HOME
declare -x XDG_STATE_HOME="/home/Bisho/.local/state"
[---SUCCESS---]

[===diagnose===] declare -p ILLOGICAL_IMPULSE_VIRTUAL_ENV
declare -x ILLOGICAL_IMPULSE_VIRTUAL_ENV="~/.local/state/quickshell/.venv"
[---SUCCESS---]

/============================\
| Checking directories/files |
\============================/

[...]

PR #1941, opened 2025-09-05
Fix non-existent property error in SysTray.qml

Vercixx:
## Describe your changes
This PR fixes an error, mentioned in #1940, by adding an additional import to SysTray.qml

Should be enough to fix an error


end-4:
The right way to fix this is to update using install.sh instead of update.sh

PR #1950, opened 2025-09-07
Make window snap respect window gaps

kxrur:
## Describe your changes
Without this option the windows snap to the borders of the monitor which is ugly with round corners.

https://github.com/user-attachments/assets/661c801e-1bf5-45c5-bc27-5b366b555926

might want to increase the snapping trigger zone [docs](https://wiki.hypr.land/Configuring/Variables/#snap) (default is 10 px) but i'll leave that up to you


end-4:
very interesting
thank you

PR #1966, opened 2025-09-11
fix(brightness): use the monitor's max brightness instead of '100'

Emmmmllll:
## Summary
Use the actual max Monitor brightness instead of hard coded `100`.

## Problem
Some monitors report a max brightness other than `100` from ddcutil (in my case 50).
Thus the brightness only changes from e.g. 1-50 and stays the same from 50+.

## Solution
In the ddc monitor initialization remember the max brightness and use it instead of `100` when updating the brightness.
If not using ddc the max brightness defaults to 100.

## Changes
`.config/quickshell/ii/services/Brightness.qml`

## Test Scenarios
- [x] On local machine

### DIFF
--- a/.config/quickshell/ii/services/Brightness.qml
+++ b/.config/quickshell/ii/services/Brightness.qml
@@ -76,7 +76,7 @@ Singleton {
         required property ShellScreen screen
         readonly property bool isDdc: root.ddcMonitors.some(m => m.model === screen.model)
         readonly property string busNum: root.ddcMonitors.find(m => m.model === screen.model)?.busNum ?? ""
-        property int maxBrightness: 100
+        property int rawMaxBrightness: 100
         property real brightness
         property bool ready: false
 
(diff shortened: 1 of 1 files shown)


.config/quickshell/ii/services/Brightness.qml
```diff
@@ -95,16 +96,17 @@ Singleton {
             stdout: SplitParser {
                 onRead: data => {
                     const [, , , current, max] = data.split(" ");
-                    monitor.brightness = parseInt(current) / parseInt(max);
+                    monitor.maxBrightness = parseInt(max);
+                    monitor.brightness = parseInt(current) / monitor.maxBrightness;
```
end-4: This way brightness is still % instead of based on the whatever non-100 max

end-4:
Wait uhh I was kinda wrong there
The real problem is if you set maxBrightness like that, for brightnessctl the max would be gigantic, and the setting command still uses % and would set something like 69000% instead of 69000 or 71%. So removing the `%` in the brightnessctl setting command should do

.config/quickshell/ii/services/Brightness.qml
end-4: i checked the part below and turns out this is fine

PR #1970, opened 2025-09-11
fix(ai): Add JSON escaping and fallback model to primary-buffer-query.sh

vvdbx:
## Summary
- Fix JSON injection vulnerability by properly escaping clipboard content using `jq`
- Add fallback to llama3.2 model when primary model detection fails or returns empty responses
- Add content length limiting (2000 chars) to prevent overflow
- Add silent curl flag and error handling for better reliability

## Security Fix Details
The original script was vulnerable to JSON injection where malicious clipboard content could break out of the JSON string and potentially execute arbitrary commands. This is now fixed by using `jq` to properly escape all content.

## Test plan
- [x] Test with normal clipboard content
- [x] Test with content containing quotes, newlines, and special characters
- [ ] Test fallback behavior when primary model is unavailable
- [x] Verify content length limiting works correctly

🤖 Generated with [Claude Code](https://claude.ai/code)


vvdbx:
Closing this PR in favor of #1973 which contains only the critical security fix without the problematic fallback mechanism. The security fix is the urgent priority that needs to be merged.

PR #1975, opened 2025-09-12
Translation : update zh_CN.json

xiaomeng2004:
Update the Chinese translation file to match the latest interface display


end-4:
Thank you

PR #1977, opened 2025-09-12
Disable mouse acceleration

haxtaxed:
Force disabled mouse acceleration, seems like a more sane default


end-4:
no it's not. only osu lazer players and artists needs that

PR #2013, opened 2025-09-17
feat(sideright): add Flatpak support for EasyEffects service and toggle

0blivi0nis:
This commit introduces support for the Flatpak version of EasyEffects within QuickShell, alongside the existing native package support. This allows users to seamlessly manage EasyEffects whether it's installed via a traditional package manager or Flatpak. 

Key changes include:
- **`.config/quickshell/ii/services/EasyEffects.qml`**:
    - Modified `fetchAvailabilityProc` to check for both native `easyeffects` and Flatpak `com.github.wwmm.easyeffects`.
    - Updated `fetchActiveStateProc` to detect active processes for both native and Flatpak EasyEffects.
    - Enhanced `disable` function to gracefully terminate both native and Flatpak EasyEffects processes.
    - Improved `enable` function to launch the Flatpak version if available, falling back to the native version otherwise.
- **`.config/quickshell/ii/modules/sidebarRight/quickToggles/EasyEffectsToggle.qml`**:
    - Modified the `altAction` (right-click) to prioritize launching the Flatpak EasyEffects GUI, with a fallback to the native application.

These changes ensure a more robust and flexible integration of EasyEffects into QuickShell, accommodating different installation methods without requiring additional user configuration. The reason behind this pull request is so I don't have to download all the effects separately.

✅ Yes, I have tested it with both regular and flatpak version of easyeffects and it works fine

### DIFF
--- a/.config/quickshell/ii/modules/sidebarRight/quickToggles/EasyEffectsToggle.qml
+++ b/.config/quickshell/ii/modules/sidebarRight/quickToggles/EasyEffectsToggle.qml
@@ -21,7 +21,7 @@ QuickToggleButton {
     }
 
     altAction: () => {
-        Quickshell.execDetached(["easyeffects"])
+        Quickshell.execDetached(["bash", "-c", "flatpak run com.github.wwmm.easyeffects || easyeffects"])
         GlobalStates.sidebarRightOpen = false
     }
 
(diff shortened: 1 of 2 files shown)


end-4:
thank you

PR #2016, opened 2025-09-18
bar: fixed kb layout+variant display in top right

Smartyguy001:
## Describe your changes

In the top-right bar, the keyboard layout did not accurately change when also providing a variant.
Here is an example of a keyboard setup where nothing happened before:
```
kb_layout = us,ro
kb_variant = ,std
kb_options = grp:shifts_toggle
```
In this case, pressing both shift keys would cause the layout to change, allowing for the input of Romanian characters. However, in the top right, "us" will still be displayed.
The fix is supposed to handle the layout + variant case.

Just see if you like it :)


end-4:
Acts funny with vertical bar

end-4:
I fixed the above
Thanks for your contribution

PR #2019, opened 2025-09-18
Trying to improve transparency.

gautam-kumar-0:
## Describe your changes
A very simple fix to improve the visual hierarchy in transparent mode.
I know it's not perfect , yet I think its better 🙃

|Before | After[updated] |
|---|---|
|original_dark|dark_adjusted_transparency|
|original_dark1|dark_adjusted_transparency1|
|original_light|light_transparency1|
|original_light1|light_transparency|

You can see the screenshots and I was not able find any side effects. Merge it if you like it or just reject it.


end-4:
Just using some transparent white is bad. The notifications on the first comparison look weird.
But this indeed makes things look clearer. I think this is heading in the right direction and we can just find cleaner values, like using transparentized colOnLayer0 instead of transparentized white

gautam-kumar-0:
Adjusted notification transparency and switched to using `colLayer0` and `colOnLayer0` instead of "white"

end-4:
what happened?

end-4:
I tested the colors anyway and found button hovering to be kind of weird... Guess that's why

gautam-kumar-0:
> I tested the colors anyway and found button hovering to be kind of weird... Guess that's why

It also had issues with notification popup , so I thought this was too simplistic approach. Hovering and Active colors could be modified in  the same way but again that will make it complicated. I also tried to use Qt's blur features but It was crashing the quickshell upon closing the sidebar.

Edit: I also got better results by just adjusting  `backgroundTransparency` and `contentTransparency` in `config.json`

PR #2030, opened 2025-09-21
Changed the night light on/off/auto mode to continuous instead of right click to auto

tnakirhs:
When Night Light is enabled by clicking the Quicktoggle, and then switched to Auto Mode via a right-click, the Night Light remains enabled if the current time falls within the Scheduled Time range (from the start time to the end time).

However, if the Quicktoggle is clicked again, the Night Light gets disabled, even though Auto Mode is still active and the current time is still within the Scheduled Time range.

Upon reloading the QuickShell, the Night Light is enabled again, but the Quicktoggle remains in the Disabled state, while Auto Mode is still active.

I have made the states continuous—Off, On, and Auto—so that only one state can be active at a time.

### DIFF
--- a/.config/quickshell/ii/modules/sidebarRight/quickToggles/NightLight.qml
+++ b/.config/quickshell/ii/modules/sidebarRight/quickToggles/NightLight.qml
@@ -7,22 +7,38 @@ import Quickshell.Io
 
 QuickToggleButton {
     id: nightLightButton
-    property bool enabled: Hyprsunset.active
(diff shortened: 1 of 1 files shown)


end-4:
Well I made on/off and auto separate because I wanted this, but the way I implemented the service didn't allow for the red arrows

image

end-4:
e97f819a5cb249667aeb191366ec56a32b4d00ee should make the behavior consistent with the above state diagram

Issue #2026, opened 2025-09-20
Move icons position in the dock


reakjra:
### What would you like to be added?

Since currently when you hold and drag the mouse on the icons inside the dock, they will move altogether basically doing nothing, since the size of the dock grows with the number of the apps, maybe it could be more useful adding the feature to swap the icons position individually left/right, like on Windows' taskbar. 

Currently:

https://github.com/user-attachments/assets/b118fa11-0a5d-4276-9b9b-48a25deb89da

### How will it help?

makes the dock more dynamic

Sanaruca:
#3045

PR #2050, opened 2025-09-26
feature(dock): drag & drop feature in the dock

reakjra:
## Describe your changes

Added a drag & drop feature in the dock as I requested in https://github.com/end-4/dots-hyprland/issues/2026. Now you can move icons in the dock

https://github.com/user-attachments/assets/f732da06-97e2-463f-9c13-61f587bfef29

Mostly yes, it works fine, it feels just a little clunky and maybe it should be adjusted the fluidity of the icon moving

EDIT: Nvm, it has stilll a few flaws: 
Icon not disappearing after moving it and then closing the application
Seems there's a issue about icons not showing app


reakjra:
I need to revision the whole code since it seems the algorithm genuinely sucks when fixed the flaws I mentioned. Apps would not switch from left to right due the index method, etc.

PR #2056, opened 2025-09-27
Added gentoo eBuild

jwihardi:
## Describe your changes
Added a Gentoo eBuild based off the the PKGBUILDs plus some necessary dependencies.


clsty:
Thanks for your effort.

Currently, non-Arch(-based) distros are not supported and we do not accept PRs like this cuz we do not have ability to maintain them.

You're welcome to create your own repo which supports gentoo and then make PR in [dots-hyprland-wiki](https://github.com/end-4/dots-hyprland-wiki) to add your repo link there, specifically [01.setup.md](https://github.com/end-4/dots-hyprland-wiki/blob/main/src%2Fcontent%2Fdocs%2Fen%2Fii-qs%2F01.setup.md).

See also #1061, it's the planned future how we would achieve cross-platform.

clsty:
Update:

Since the install.sh has been newly designed also for other distros, you are welcomed to make PR following the [document](https://github.com/end-4/dots-hyprland-wiki/blob/c836a54b895afbdb9b01378ed94bd15b3d3071dd/src/content/docs/en/ii-qs/01.setup.md?plain=1#L32-L46).

jwihardi:
Okay thank you, fixing it rn

PR #2067, opened 2025-09-28
cheatsheet: added element weight to periodic table

Smartyguy001:
## Describe your changes

Added the element weight to the periodic table in the cheatsheet. (I find it useful for chemistry class lol)
It was fairly easy, as the data was already stored in periodic_table.js.

The design/layout is pretty rough, but it seems readable enough. I'm open to any sort of changes and proposals!


end-4:
i realized how ass this looks when you hover over the elements, but this has never looked good in the first place

PR #2071, opened 2025-09-28
feat(alertSounds): add battery & pomodoro alert sounds

0blivi0nis:
## Describe your changes

- Battery Alerts: Implemented audio notifications for battery thresholds
- Pomodoro Alerts: Added chime sounds for Pomodoro timer events
- Settings: Option to toggle alert sounds & change battery full warning

settings

Yes, has been tested

### DIFF
--- a/.config/quickshell/ii/modules/common/Config.qml
+++ b/.config/quickshell/ii/modules/common/Config.qml
@@ -375,7 +375,7 @@ Singleton {
             property JsonObject sounds: JsonObject {
                 property bool battery: false
                 property bool pomodoro: false
-                property string theme: "ocean"
+                property string theme: "freedesktop"
             }
 
             property JsonObject time: JsonObject {


end-4:
- No binary files. Use sounds from the system sound pack
- Make it disabled by default
- Make full battery alert have a configurable threshold, default to 101% (to disable)

0blivi0nis:
> * No binary files. Use sounds from the system sound pack
> 
>     * Make it disabled by default
> 
>     * Make full battery alert have a configurable threshold, default to 101% (to disable)

That should be everything. Anything over 100 for the threshold wont play anyways since the battery cant reach that level and trigger the alert

0blivi0nis:
canberra-gtk-play doesnt follow kde system sound themes so had to add another option to choose the theme manually. By default the options is ocean with a choice of freedesktop as well . If you install another sound theme you have to make sure the file extension is oga otherwise it wont work.

end-4:
Can you reapply your changes to a new branch based on upstream? I don't want the binary files sticking around because they're currently in the history

end-4:
> canberra-gtk-play doesnt follow kde system sound themes so had to add another option to choose the theme manually. By default the options is ocean with a choice of freedesktop as well . If you install another sound theme you have to make sure the file extension is oga otherwise it wont work.

~~canberra-gtk-play?? But we're using ffplay?~~ Okay I see...  Assuming sound packs are in `/usr/share/sounds` would be incorrect though

0blivi0nis:
> ~canberra-gtk-play?? But we're using ffplay?~ Okay I see... Assuming sound packs are in `/usr/share/sounds` would be incorrect though

I checked a bunch of sound packs installed from the aur and they all install there. Unless you install them manually to `~/.local/share/sounds` I dont see another location for system sounds. Also some sound packs ship with the file extension .ogg but I couldnt find a way to add both .oga and .ogg, so if you know how please do so.

If there is a way to make canberra-gtk-play follow kde system sound settings that would be the best option since you only need the sound name and dont have to worry about path, extensions or choosing a theme. But from what I found there isnt a way, even tried setting `gtk-sound-theme-name=ocean` in settings.ini for gtk3/4. It still uses freedesktop sounds

> Can you reapply your changes to a new branch based on upstream? I don't want the binary files sticking around because they're currently in the history

Sounds good

0blivi0nis:
Does that work for cleaning up the history? or do I have to make a new PR?

.config/quickshell/ii/services/Audio.qml
```diff
@@ -48,7 +50,17 @@ Singleton {
             }
             lastVolume = sink.audio.volume;
         }
-        
+    }
+
+    function playSystemSound(soundName) {
+        const fullPath = `/usr/share/sounds/${root.audioTheme}/stereo/${soundName}.oga`;
```
nyx-4: IMO, this path should be atleast in configs, because I personally use some custom sound in ~/Downloads. please don't break my workflow...

P.S. This is only an advice, and not binding in any terms.

.config/quickshell/ii/services/TimerService.qml
```diff
@@ -65,8 +65,9 @@ Singleton {
             }
 
             Quickshell.execDetached(["notify-send", "Pomodoro", notificationMessage, "-a", "Shell"]);
-            if (alertSound)
-                Quickshell.execDetached(["ffplay", "-nodisp", "-autoexit", alertSound]);
+            if (Config.options.audio.alertSound.pomodoro) {
+                Audio.playSystemSound("alarm-clock-elapsed")
```
nyx-4: Which audio pack would you recommend? Since, default one's (`freedesktop`) are best described as **cacophony**.

.config/quickshell/ii/services/Audio.qml
0blivi0nis: Thats understandable. We could add the path in the config instead of adding theme name and just substitute that. 

The only thing with that is we would have to define the file extension in the actual service then unless we can find a way to add multiple extensions and have it automatically choose based on whats available. Because this would still only lock you to .oga files. Or if we can get camberra-gtk-play to follow kde sound themes then extensions arent needed, but that would also limit the function to just systemsounds.

In my personal setup i just added another similar function playSound and kept the path for that to '~/.config/quickshell/ii/assets/sounds/' and then define the files needed for other services where the function is called which lets me use different file extensions for different services if I have to.

.config/quickshell/ii/services/TimerService.qml
0blivi0nis: I personally found harmony2 to be the best. You can install it from the aur but then you have to go its folders and change the file extensions from .ogg to .oga for it to work with this current setup. 

I created a metapackage for myself that installs it and replaces the file extensions during the base install.

0blivi0nis:
Found this to work for both file types.

```
function playSystemSound(soundName) {
    const ogaPath = `/usr/share/sounds/${root.audioTheme}/stereo/${soundName}.oga`;
    const oggPath = `/usr/share/sounds/${root.audioTheme}/stereo/${soundName}.ogg`;

    // Try playing .oga first
    let command = [
        "ffplay",
        "-nodisp",
        "-autoexit",
        ogaPath
    ];
    Quickshell.execDetached(command);

    // Also try playing .ogg (ffplay will just fail silently if file doesn't exist)
    command = [
        "ffplay",
        "-nodisp",
        "-autoexit",
        oggPath
    ];
    Quickshell.execDetached(command);
}
```

If this looks good then i can push that. The only thing now is the the hard coded location

.config/quickshell/ii/services/Audio.qml
0blivi0nis: > IMO, this path should be atleast in configs, because I personally use some custom sound in ~/Downloads. please don't break my workflow...
> 
> P.S. This is only an advice, and not binding in any terms.

How are you playing sounds currently 🤔 and if you use them often put them in `~/.local/share/sounds/`

0blivi0nis:
Okay that should be everything. I dont see another way to make the path modular, or see the need for it since system sounds are placed in that directory anyways. Placing the location in Directories.qml wont work either since it needs another folder name after the theme name. And defining file extensions in the actual service also wont work since some themes use ogg instead of the default oga file format.

nyx-4:
> const ogaPath = `/usr/share/sounds/${root.audioTheme}/stereo/${soundName}.oga`;

```js
const ogaPath =  Config...basePath + `${root.audioTheme}/stereo/${soundName}.oga`; 
```

should work, wouldn't it??

P.S. I know it's kinda weird to ask, but can you just add an extra **if** so that I still can play that 5 min music during pomodoroBreak  👉 👈 🥺?

0blivi0nis:
> > const ogaPath = `/usr/share/sounds/${root.audioTheme}/stereo/${soundName}.oga`;
> 
> ```js
> const ogaPath =  Config...basePath + `${root.audioTheme}/stereo/${soundName}.oga`; 
> ```
> 
> should work, wouldn't it??
> 
> P.S. I know it's kinda weird to ask, but can you just add an extra **if** so that I still can play that 5 min music during pomodoroBreak 👉 👈 🥺?

Yup that will work but still only limit you to oga and ogg files which are mostly used by system sounds. I'll let @end-4 decide if that needs to be modular or not since I dont see the need for it because we are using system sounds. Doing it what way, we can also remove the stereo from there add it to the service directly but defining the file type in the service will break the alerts since some sound packs use ogg instead of the oga like ocean and freedesktop

I added the old function back just replaced the name from alertSound to breakSound. Now if you have a file defined like before it will play it otherwise it will only play the system alert sound. If you only want your custom file to play then disable pomodoro in the settings app.

P.s. Not weird to ask at all. Collaboration is what improves stuff!

.config/quickshell/ii/modules/settings/GeneralConfig.qml
```diff
@@ -47,12 +47,48 @@ ContentPage {
...
+                text: Translation.tr("Full warning")
+                value: Config.options.battery.full
+                from: 0
+                to: 100
+                stepSize: 5
+                onValueChanged: {
+                    Config.options.battery.full = value;
+                }
+            }
+        }
```
end-4: put below low/suspend stuff, it's less important

.config/quickshell/ii/modules/settings/GeneralConfig.qml
```diff
@@ -47,12 +47,48 @@ ContentPage {
...
+                ConfigSwitch {
+                    buttonIcon: "av_timer"
+                    text: Translation.tr("Pomodoro")
+                    checked: Config.options.audio.alertSound.pomodoro
+                    onCheckedChanged: {
+                        Config.options.audio.alertSound.pomodoro = checked;
+                    }
+                }
+            }
+        }
```
end-4: we sort alphabetically so alert sounds go before audio

.config/quickshell/ii/services/Audio.qml
```diff
@@ -48,7 +50,27 @@ Singleton {
             }
             lastVolume = sink.audio.volume;
         }
-        
+    }
+
+    function playSystemSound(soundName) {
+        const ogaPath = `/usr/share/sounds/${root.audioTheme}/stereo/${soundName}.oga`;
+        const oggPath = `/usr/share/sounds/${root.audioTheme}/stereo/${soundName}.ogg`;
```
end-4, line 57: this feels wrong...

.config/quickshell/ii/services/Battery.qml
```diff
@@ -45,12 +54,49 @@ Singleton {
...
+                Translation.tr("Battery full"),
+                Translation.tr("Please unplug the charger"),
+                "-u", "critical",
+                "-a", "Shell"
+            ]);
+        }
+        else if (percentage < Config.options.battery.full / 100 && fullNotificationPlayed) {
+            fullNotificationPlayed = false;
+        }
+    }
```
end-4: Please implement the full warning like how low and critical are implemented

end-4:
> P.S. I know it's kinda weird to ask, but can you just add an extra if so that I still can play that 5 min music during pomodoroBreak 👉 👈 🥺?

I'm already tired of this so no

.config/quickshell/ii/modules/settings/GeneralConfig.qml
0blivi0nis: Figured they would go under audio thats why I put them there. Made a new section above audio now. Should I move it out of audio in the config too? 

```

    "audio": {
        "alertSound": {
            "battery": true,
            "pomodoro": true,
            "theme": "freedesktop"
        },
        "protection": {
            "enable": false,
            "maxAllowed": 99,
            "maxAllowedIncrease": 10
        }
    },
```
or is this fine?

.config/quickshell/ii/services/Audio.qml
0blivi0nis: I cant really find another way to support other sound packs. With ffplay we need to define the file types. 

We can get rid of the file extensions and the path if we use canberra-gtk-play, but I haven't been able to figure out a way for canberra-gtk-play to use the system sound chosen by kde even after defining it as an env variable or in gtk settings.

0blivi0nis:
> > P.S. I know it's kinda weird to ask, but can you just add an extra if so that I still can play that 5 min music during pomodoroBreak 👉 👈 🥺?
> 
> I'm already tired of this so no

So remove the breakSound I added?

.config/quickshell/ii/modules/settings/GeneralConfig.qml
```diff
@@ -10,11 +10,36 @@ import qs.modules.common.widgets
 ContentPage {
     forceWidth: true
 
+    ContentSection {
+        icon: "notification_sound"
+        title: Translation.tr("Alerts")
```
end-4: Sound is the main aspect here, so "Alert sounds" is better than "alerts". In fact, just make this "Sounds" (and adjust the order accordingly)

.config/quickshell/ii/services/Battery.qml
```diff
@@ -18,39 +19,78 @@ Singleton {
...
             Quickshell.execDetached(["bash", "-c", `systemctl suspend || loginctl suspend`]);
         }
     }
+
+    onIsFullAndChargingChanged: {
+        if (available && isFullAndCharging) Quickshell.execDetached([
+            "notify-send",
+            Translation.tr("Battery full"),
+            Translation.tr("Please unplug the charger"),
+            "-u", "critical",
```
end-4: Not exactly critical

end-4:
> So remove the breakSound I added?

I didn't even notice that's for the 5 minute thing
But since that's the case remove it

0blivi0nis:
That should be everything, force pushed again to clean up the commits.

Moved alertSounds from under audio to sounds by itself.

0blivi0nis:
We need to decide if we are gonna ship ocean or freedesktop by default. Ocean doesnt have suspend-error which is used for critical battery warning, so if we decide to ship ocean by default that needs to be changed otherwise we just have to change the default theme to freedesktop

end-4:
Default to freedesktop, because everyone has it
It sucks, but we're disabling sounds by default anyway

0blivi0nis:
> Default to freedesktop, because everyone has it It sucks, but we're disabling sounds by default anyway

Done

0blivi0nis:
Replaced by #2223, couldnt really figure out a way to push to this with all the new folder changes

PR #2072, opened 2025-09-28
fix: transparent wallpaper bg

claymorwan:
## Describe your changes

Made the wallpaper's background transparent 

ya it works

### DIFF
--- a/.config/quickshell/ii/modules/background/Background.qml
+++ b/.config/quickshell/ii/modules/background/Background.qml
@@ -93,7 +93,7 @@ Variants {
             left: true
             right: true
         }
-        color: CF.ColorUtils.transparentize(CF.ColorUtils.mix(Appearance.colors.colLayer0, Appearance.colors.colPrimary, 0.75), (bgRoot.wallpaperIsVideo ? 1 : 0))
+        color: "transparent"
         Behavior on color {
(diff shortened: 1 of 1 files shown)


end-4:
Don't you see the `bgRoot.wallpaperIsVideo` check? If it's not working that should be fixed, not by making the bg totally transparent

claymorwan:
> Don't you see the `bgRoot.wallpaperIsVideo` check? If it's not working that should be fixed, not by making the bg totally transparent

no i didn't, is it supposed to make the bg transparent when it's a video ? cuz my issue is that im using linux-wallpaperengine and i can't see the wallpaper cuz of that bg

end-4:
It's supposed to make the background transparent when the wallpaper is set to video. Wallpaper engine is not supported and I am not accepting a hard coded value that breaks the intended use case in favor of an unsupported one.

claymorwan:
o ok then i suppose i'll have to change it myself each time, just outta curiosity tho what is the opaque wallpaper background for ?

claymorwan:
o ok then i suppose i'll have to change it myself each time, just outta curiosity tho what is the opaque wallpaper background for ?

end-4:
smoother transition

PR #2083, opened 2025-09-30
Papirus Icon Theme(Catppuccin Folder Colors) from wallpaper

RawnakHasan:
## Describe your changes

Adds support for the Papirus Icon Theme and automatically sets folder colors based on the primary Catppuccin color extracted from the current wallpaper.

It is as I am using it for a week

image


end-4:
I don't like Papirus. It does not respect original icons (in a kiddy way instead of developing its own visual style) and looks material 2 ish

PR #2084, opened 2025-09-30
feat(bar): Display a numberr of system package updates

difrost:
## Describe your changes

With this PR one can enable a display of available updates to system packages (currently supported on Arch and Fedora Linux). The counter is refreshed within a customizable interval (specified in minutes via Settings) - manual refresh is triggered on-click. Hover over the counter will show last fetch time.

image

image

Note: While I'm an engineer/developer with long experience that's my first attempt on QML/QT/QS and this entire PR has been done in good old style by just looking at others code and - mostly - ignoring docs.


0blivi0nis:
Nice! The settings should go in bar config imo

difrost:
I've turned into a separate `System` section to use it further for all  outside-of-shell, system related controls. The placement is also not the best imho - tho I can say same for many bar items. 

All in all this is more of a beta version. I would like to add a script that will actually allow quick update via click on the counter (instead of fetching the updates). For that I need to resurrect my old laptop running Fedora. Plus we should add a better runtime awareness - exactly what distro we run on, what is the AUR helper if running on Arch... That will help with further OS integrations.

clsty:
> Plus we should add a better runtime awareness - exactly what distro we run on, what is the AUR helper if running on Arch... That will help with further OS integrations.

I guess you may use something like this:

https://github.com/end-4/dots-hyprland/blob/f87e6e2b6ae2f22ee3d9ade6525a213c80aeeb7c/diagnose#L35-L45

end-4:
This is yet another useless piece of information that adds no value to productivity thrown at a random place, so I'm really hesitant...

Maybe it's time for the window title on the left side to go? Since it was also a filler I added because I didn't know what to put there

PR #2086, opened 2025-09-30
lock: added reboot button

jojo2504:
## Describe your changes
Added a reboot button in between sleep and shutdown in the `LockSurface`

yes


jojo2504:
image

I find this QoL useful if someone has now the lockscreen autolaunched and want to reboot to windows for example, or any others reasons he may want to reboot without needing to unlock the session for nothing.

end-4:
We should put it to the right of the shutdown button so the order is consistent with the session menu
Small change so I'm going to do it and merge now, thanks

PR #2087, opened 2025-09-30
fix: ignore case (The latest Hyprland process name has changed)

cyclesw:
hyprland's main process name has been changed from Hyprland to hyprland (all lower-case).


clsty:
That's true, I also noticed that it used to be Hyprland but now it's hyprland.

end-4:
So that's why the button won't work... thank you

PR #1953, opened 2025-09-09
Fix various issues of overview with multiple monitors setup

cupcat121:
## Describe your changes
It happens with laptop linked with external monitors, and there are monitors disabled by Hyprland.
This fixed OvervideWidget cannot get correct 'monitors' in multi-monitors setup.

Tested case:
```
[{
    "id": 1,
    "name": "DP-2",
    "description": "HKC OVERSEAS LIMITED G24H1 0000000000001",
    "make": "HKC OVERSEAS LIMITED",
    "model": "G24H1",
    "serial": "0000000000001",
    "width": 1920,
    "height": 1080,
    "refreshRate": 200.00000,
    "x": 0,
    "y": 0,
    "activeWorkspace": {
        "id": 3,
        "name": "3"
    },
    "specialWorkspace": {
        "id": 0,
        "name": ""
    },
    "reserved": [0, 40, 0, 0],
    "scale": 1.00,
    "transform": 0,
    "focused": true,
    "dpmsStatus": true,
    "vrr": true,
    "solitary": "0",
    "activelyTearing": false,
    "directScanoutTo": "0",
    "disabled": false,
    "currentFormat": "XRGB8888",
    "mirrorOf": "none",
    "availableModes": ["1920x1080@60.00Hz","1920x1080@200.00Hz","1920x1080@180.00Hz","1920x1080@120.00Hz","1920x1080@119.88Hz","1920x1080@59.94Hz","1920x1080@50.00Hz","1680x1050@59.95Hz","1280x1024@75.03Hz","1280x1024@60.02Hz","1440x900@59.89Hz","1280x720@59.94Hz","1280x720@50.00Hz","1024x768@75.03Hz","1024x768@60.00Hz","800x600@75.00Hz","800x600@60.32Hz","720x576@50.00Hz","720x480@59.94Hz","640x480@75.00Hz","640x480@59.94Hz","640x480@59.93Hz"]
},{
    "id": 0,
    "name": "eDP-1",
    "description": "LG Display 0x0625",
    "make": "LG Display",
    "model": "0x0625",
    "serial": "",
    "width": 1920,
    "height": 1080,
    "refreshRate": 144.00200,
    "x": 1920,
    "y": 0,
    "activeWorkspace": {
        "id": 2,
        "name": "2"
    },
    "specialWorkspace": {
        "id": 0,
        "name": ""
    },
    "reserved": [0, 40, 0, 0],
    "scale": 1.00,
    "transform": 0,
    "focused": false,
[...]

Issue #1866, opened 2025-08-23
Overview looking weird on multiple monitors with different sizes


Stanko2:
### Step 1. Before you submit

- [x] I have read the [Troubleshooting](https://end-4.github.io/dots-hyprland-wiki/en/ii-qs/04troubleshooting/) and [Usage](https://end-4.github.io/dots-hyprland-wiki/en/ii-qs/02usage/) pages.
- [x] I've successfully updated to the latest version following the [guidance](https://end-4.github.io/dots-hyprland-wiki/en/ii-qs/01setup/#updating).
- [x] I've successfully updated the system packages to the latest.
- [ ] I've ticked the checkboxes without reading their contents

### Step 2. Quick diagnose info

<details><summary>Quick diagnose</summary>

```

/========================\
| Checking git repo info |
\========================/

[===diagnose===] git remote get-url origin
https://github.com/end-4/dots-hyprland
[---SUCCESS---]

[===diagnose===] git rev-parse HEAD
2b87d939bb20efe2b424e74bdaac26160139bfdb
[---SUCCESS---]

/=================\
| Checking distro |
\=================/

[===diagnose===] ii_check_distro
LSB Version:	n/a
Distributor ID:	Arch
Description:	Arch Linux
Release:	rolling
Codename:	n/a
[---SUCCESS---]

/====================\
| Checking variables |
\====================/

[===diagnose===] declare -p XDG_CACHE_HOME
./diagnose: line 30: declare: XDG_CACHE_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p XDG_CONFIG_HOME
./diagnose: line 30: declare: XDG_CONFIG_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p XDG_DATA_HOME
./diagnose: line 30: declare: XDG_DATA_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p XDG_STATE_HOME
./diagnose: line 30: declare: XDG_STATE_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p ILLOGICAL_IMPULSE_VIRTUAL_ENV
declare -x ILLOGICAL_IMPULSE_VIRTUAL_ENV="~/.local/state/quickshell/.venv"
[---SUCCESS---]

/============================\
| Checking directories/files |
\============================/

[===diagnose===] ls -l /home/stanko/.local/state/quickshell/.venv
total 16
[...]

PR #2091, opened 2025-09-30
fix overview scaling issue

unlimitedsola:
## Describe your changes

This PR fixes the scaling issue in the Overview widget. The issue occurs when a setup has monitors with different *scaled resolution*, as shown in https://github.com/end-4/dots-hyprland/issues/1866.

It also accounts for transformation/rotation, but only for the size of the preview window. The preview content is still in the wrong rotation, however.

Yes. The change is verified on my setup and I also got another confirmation here: https://github.com/end-4/dots-hyprland/pull/1953#issuecomment-3351677164.

I prioritized minimizing the changes to make it easier to grasp. Feel free to refactor or reorganize it as you see fit.

### DIFF
--- a/.config/quickshell/ii/modules/overview/OverviewWindow.qml
+++ b/.config/quickshell/ii/modules/overview/OverviewWindow.qml
@@ -17,14 +17,30 @@ Item { // Window
     property var availableWorkspaceWidth
     property var availableWorkspaceHeight
     property bool restrictToWorkspace: true
-    property real initX: Math.max((windowData?.at[0] - (monitorData?.x ?? 0) - monitorData?.reserved[0]) * root.scale, 0) + xOffset
-    property real initY: Math.max((windowData?.at[1] - (monitorData?.y ?? 0) - monitorData?.reserved[1]) * root.scale, 0) + yOffset
+    property real widthRatio: {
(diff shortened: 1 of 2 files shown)


.config/quickshell/ii/modules/overview/OverviewWindow.qml
```diff
@@ -90,7 +106,7 @@ Item { // Window
                 // console.log("Icon ratio:", root.compactMode ? root.iconToWindowRatioCompact : root.iconToWindowRatio)
                 // console.log("Scale:", root.monitorData.scale)
                 // console.log("Final:", Math.min(targetWindowWidth, targetWindowHeight) * (root.compactMode ? root.iconToWindowRatioCompact : root.iconToWindowRatio) / root.monitorData.scale)
-                return Math.min(targetWindowWidth, targetWindowHeight) * (root.compactMode ? root.iconToWindowRatioCompact : root.iconToWindowRatio) / root.monitorData.scale;
+                return Math.min(targetWindowWidth, targetWindowHeight) * (root.compactMode ? root.iconToWindowRatioCompact : root.iconToWindowRatio);
```
unlimitedsola, line 109: The division of `root.monitorData.scale` is removed because the `targetWindow{Width,Height}` already has the correct scaling accounted for.

end-4 (approved):
Thanks
To make the previews correct we'll need to rotate the ScreenCopy

PR #2105, opened 2025-10-02
bar: fix float bar tint loss

Sighthesia:
## Describe your changes

disable StyledRectangularShadow for the float bar in modules/bar/BarContent.qml, so as to make the float bar style have the same color tint with other bar styles(when transparency is enable).

It looks ok, all the three bar styles now have the same color. Let me know if the shadow is desired or further changes is needed, thanks a lot!

### DIFF
--- a/.config/quickshell/ii/modules/bar/BarContent.qml
+++ b/.config/quickshell/ii/modules/bar/BarContent.qml
@@ -27,7 +27,7 @@ Item { // Bar content region
 
     // Background shadow
     Loader {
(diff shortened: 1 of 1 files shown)


end-4:
No, this makes it less readable. Provide a config option instead of turning the shadow into an useless hardcoded piece of code that never runs like this

Sighthesia (commented):
ok thank you for your guidances, i have make the shadow as as config option `floatSytleShadow`, that can be configured by modifying the config.json.

PR #2108, opened 2025-10-03
fix: Fetch AI API key after lockscreen unlock

kryiexi:
## Describe your changes
The AI sidebar requests the keyring ASAP, which is problematic when using the new lock/LaunchOnStartup feature with the intent to unlock the keyring with the password provided to the lockscreen 

[lock: also unlock keyring on unlock] 3c00805

This change ensures that the AI waits until the lockscreen has been unlocked before requesting the keyring. Due to a race condition, a timer was added to allow the config to load and lock the lockscreen before checking if the lockscreen was locked.

Ready

### DIFF
--- a/.config/quickshell/ii/services/Ai.qml
+++ b/.config/quickshell/ii/services/Ai.qml
@@ -357,6 +357,30 @@ Singleton {
     }
     property ApiStrategy currentApiStrategy: apiStrategies[models[currentModelId]?.api_format || "openai"]
 
+    //Fetch API keys if needed
+    //Timer to beat race-condition for lock/launchOnStartup
+    Timer {
+        id: initTimer
+        interval: 2000 // Wait 2000ms for system to initialize
+        running: true
+        repeat: false
+        onTriggered: {
+            if (!GlobalStates.screenLocked) {
+                KeyringStorage.fetchKeyringData();
+            }
+        }
+    }
+
+    // Unlock keyring when lockscreen is unlocked
+    Connections {
+        target: GlobalStates
+        function onScreenLockedChanged() {
+            if (!GlobalStates.screenLocked) {
+                KeyringStorage.fetchKeyringData();
+            }
+        }
+    }
+
     Connections {
         target: Config
         function onReadyChanged() {
@@ -533,8 +557,6 @@ Singleton {
         modelId = modelId.toLowerCase()
         if (modelList.indexOf(modelId) !== -1) {
             const model = models[modelId]
-            // Fetch API keys if needed
-            if (model?.requires_key) KeyringStorage.fetchKeyringData();
             // See if policy prevents online models
             if (Config.options.policies.ai === 2 && !model.endpoint.includes("localhost")) {
                 root.addMessage(


harveywuk:
Can confirm this fixes the additional authentication  popup issues when using lock screen to unlock the keying with autologin SDDM

.config/quickshell/ii/services/Ai.qml
```diff
@@ -357,6 +357,30 @@ Singleton {
...
+
+    // Unlock keyring when lockscreen is unlocked
+    Connections {
+        target: GlobalStates
+        function onScreenLockedChanged() {
+            if (!GlobalStates.screenLocked) {
+                KeyringStorage.fetchKeyringData();
+            }
+        }
+    }
```
end-4: Race condition. Screen unlocked does not mean keyring is unlocked, since the keyring unlock is async. Add a signal to `Lock.qml`'s keyring unlock process and use that.

.config/quickshell/ii/services/Ai.qml
```diff
@@ -357,6 +357,30 @@ Singleton {
...
+        id: initTimer
+        interval: 2000 // Wait 2000ms for system to initialize
+        running: true
+        repeat: false
+        onTriggered: {
+            if (!GlobalStates.screenLocked) {
+                KeyringStorage.fetchKeyringData();
+            }
+        }
+    }
```
end-4: maybe we can use a `Connections` targeting `KeyringStorage.Component` hooked to `onCompleted`?

.config/quickshell/ii/services/Ai.qml
scotthanan: Rather than using a signal, we could created a bool `GlobalStates.keyringUnlocked` which gets set to true in `Lock.qml` under the `unlockKeyring()` function.

`Lock.qml`

```
function unlockKeyring() {
	Quickshell.execDetached({
		environment: ({
			UNLOCK_PASSWORD: root.currentText
		}),
		command: ["bash", "-c", Quickshell.shellPath("scripts/keyring/unlock.sh")]
	})
	GlobalStates.keyringUnlocked = true;
}
```

 `Ai.qml` - Use below connection instead of `onScreenLockedChange`
```
Connections {
    target: GlobalStates
    function onKeyringUnlockedChanged() {
        if (GlobalStates.keyringUnlocked) {
            KeyringStorage.fetchKeyringData();
        }
    }
}
```

.config/quickshell/ii/services/Ai.qml
scotthanan: With the changes mentioned in my above comment, the main reason for the timer now is when quickshell is refreshed rather than any race conditions.

On quickshell refresh, `GlobalStates.keyringUnlocked` will set back to its default `false`. Another monitor is required (such as this timer) to ensure that we attempt an unlock if quickshell is refreshed. 

Is there a better way to do this when a refresh is called?

.config/quickshell/ii/services/Ai.qml
end-4: Right, I forgot that Lock isn't a singleton
We can use GlobalStates, but maybe have a signal instead of a bool? unlocking once doesn't mean it's unlocked forever, and we just need a signal so the keyring storage thing can re-attempt retreival of the values

end-4:
Actually, Ai is currently the only use case for `KeyringStorage`, but this is not its job. Unlocking the keyring should be done directly by `KeyringStorage` and we should move the logic there

.config/quickshell/ii/services/Ai.qml
end-4: Why do we need to re-attempt an unlock when quickshell is restarted?

harveywuk: I cant see that you do need to re attempt as the keying should already be unlocked.

scotthanan: We don't need to re-attempt an unlock after quickshell is restarted, but we will need to fetch keyring data (AI grab the API Keys from the unlocked keyring)

.config/quickshell/ii/services/Ai.qml
scotthanan: I have pushed a commit updating to use a signal now.

scotthanan:
Logic has been migrated over to `KeyringStorage.qml` from `Ai.qml`.
Better way is still needed for handling a quickshell refresh rather than using a timer that triggers after the refresh to `fetchKeyringData` (keyring is still unlocked after a refresh)

.config/quickshell/ii/services/Ai.qml
end-4: Why the delay then?

.config/quickshell/ii/services/KeyringStorage.qml
```diff
@@ -71,6 +71,27 @@ Singleton {
...
+        id: initTimer
+        interval: 2000 // Wait 2000ms for system to initialize
+        running: true
+        repeat: false
+        onTriggered: {
+            if (!GlobalStates.screenLocked) {
+                KeyringStorage.fetchKeyringData();
+            }
+        }
+    }
```
end-4: I don't see why we need a Timer. Just trigger when `Component.onCompleted` ?

scotthanan: Using Component.onCompleted to fetch based on screenlock causes issues because it runs before the lock screen initializes on startup.

Currently the process is:
1. GlobalStates.keyringUnlocked() signal fires when the user unlocks the screen via Lock.qml's unlockKeyring() function
2. On Quickshell refresh/restart, the screen doesn't get unlocked, so the signal never fires as part of unlockKeyring()
3. The timer was used instead of Component.onCompleted so that it gives time for the lock screen to initialize and doesn't immediately try to open the keyring before credentials have been provided. On a restart/refresh, it will open the keyring as it recognizes the screen is unlocked.

Instead of a timer, I could call the unlock.sh script and check the exitCode as part of Component.onCompleted

`KeyringStorage.qml`
```
Process {
    id: checkKeyring
    command: ["bash", "-c", Quickshell.shellPath("scripts/keyring/unlock.sh")]

    onExited: (exitCode) => {
        if (exitCode === 1) {
            KeyringStorage.fetchKeyringData();
        }
    }
}
Component.onCompleted: {
    checkKeyring.running = true;
}

Connections {
    target: GlobalStates
    function onKeyringUnlocked() {
        KeyringStorage.fetchKeyringData();
    }
}
```

end-4: Does `KeyringStorage` depend on the lock? `fetchKeyringData()` only starts a process within `KeyringStorage` that runs a `secret-tool` lookup...

end-4: > The timer was used instead of Component.onCompleted so that it gives time for the lock screen to initialize and doesn't immediately try to open the keyring before credentials have been provided. On a restart/refresh, it will open the keyring as it recognizes the screen is unlocked.

The 2 seconds delay does not help. The user can take as much time as they want to provide a password. And `fetchKeyringData()` does not care about whether the screen is unlocked

.config/quickshell/ii/services/KeyringStorage.qml
```diff
@@ -71,6 +71,29 @@ Singleton {
...
+        id: lockState
+        command: ["bash", "-c", Quickshell.shellPath("scripts/keyring/unlock.sh")]
+
+        // Fetch data if keyring is already unlocked
+        onExited: (exitCode) => {
+            if (exitCode === 1) {
+                KeyringStorage.fetchKeyringData();
+            }
+        }
+    }
```
end-4, line 97: Okay, fetching on init only if the screen is unlocked makes sense. But currently the `unlock.sh` script will ask for password if not provided so this would hang

.config/quickshell/ii/services/KeyringStorage.qml
end-4: closing as outdated

end-4:
I finally understand what you're trying to do. We should make `LockContext` a singleton (and we should put it in the `services` folder) so we can properly catch the initialization of it and simplify this. There's not supposed to be two contexts running anyway

.config/quickshell/ii/services/KeyringStorage.qml
scotthanan: Ah, good point. 

Could bring out just the check portion of the script so that it fetches the keyring data if it is already unlocked (quickshell refresh).

i.e.
- The signal + connection handle the unlocking from lockscreen
- Skip section for quickshell refresh

```
# Skip if already unlocked
locked_state=$(busctl --user get-property org.freedesktop.secrets \
    /org/freedesktop/secrets/collection/login \
    org.freedesktop.Secret.Collection Locked)    
if [[ "${locked_state}" == "b false" ]]; then
    echo 'Keyring is already unlocked.' >&2
    exit 1
fi
```

Really the whole point of this was to properly catch the initialization and also survive a quickshell restart.
Although the above would work, probably better to do it the way you mentioned.

clsty:
Does this PR fix #2228?

end-4:
> Does this PR fix #2228?

No, that issue is because of their setup

end-4:
I fixed it in the above

PR #2112, opened 2025-10-03
fix(settings): fixed open config button file from settings

jojo2504:
## Describe your changes
Fixed the `Config file` button not working with the tooltip "*If the button doens't work [...]*"
I reused the hyprland script `~/.config/hypr/hyprland/scripts/launch_first_available.sh`.

- `Qt.openUrlExternally` doesn't make sense to me if we open a local file in an editor
  - Imported `Quickshell` and used `Quickshell.execDetached` instead
- `${Directories.config}` : the file prefix is still here -> `file://` (or something like that)
  - Used `CF.FileUtils.trimFileProtocol` to fix that
    > CF is coming from you and the start of the file

Implementing a proper interface to select the favorite editor and "remember it" might be practical in the future (for now it just loops through a list of app until it finds something)

I think so, you can test it on your system beforehand tho

### DIFF
--- a/.config/quickshell/ii/settings.qml
+++ b/.config/quickshell/ii/settings.qml
@@ -10,6 +10,7 @@ import QtQuick
 import QtQuick.Controls
 import QtQuick.Layouts
 import QtQuick.Window
+import Quickshell
 import qs
 import qs.services
 import qs.modules.common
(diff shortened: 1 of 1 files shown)


jojo2504:
image

Fixed that button

0blivi0nis:
Choose your editior in kde settings. The button opens it properly

jojo2504:
if you meant this
image

text editor is set to nvim and it does **NOT** work (it's fresh install in a VM)

end-4:
> text editor is set to nvim and it does NOT work (it's fresh install in a VM)

And does it work if you click on the file in your file explorer?

jojo2504:
double clicking in the file explorer like in windows ? yeah it works
image

end-4:
I can confirm. However using `launch_first_available` for a file is a bad choice, because there would be no end to adding specific applications. We can add a button (or have right-clicking that button) to copy the file, but this is not the right way.

end-4:
I added file path copying

PR #2114, opened 2025-10-03
Set background opacity when using live wallpaper

baoliejuan:
## Describe your changes
Fix what was causing thumbnail to display over mpvpaper

Live wallpaper still looks buggy when reload unless set transparent wallpaper bg as in #2072


end-4:
thank you

PR #2116, opened 2025-10-03
Gentoo Install

jwihardi:
I think it's ready. It worked for me ... I know I got rejected before but word on the street is that any distro is now allowed if you follow the same installation. 

Live builds on hyprland stuff since you guys are too fast for portage to keep up smh.

### Basic Information
- There are ebuilds for each illogical-impulse dependencies, matching the _dist-arch_ directories.
- The Hyprland illogical-impulse packages need to have some live builds to keep up with Arch, so it's using ** to unmask. I had to make ebuilds for hyprgraphics, hyprland-qtutils, hyprland-qt-support, hyprlang, hyprwayland-scanner.
- Quickshell also needs to use the live ebuild
- _keywords_ is for basic unmasked without the actual keywords, _install-deps.sh_ will append the unmasking keywords based on system architecture. The file includes all unmasks for the custom ebuilds, some dependencies might need to be unmasked manually.
- _useflags_ are for recommended use flags that will make it work. I didn't specify to init system, I'm assuming that's taken into account in make.conf.

### install-deps.sh (how it works)
- Enable localrepo and guru if not already enabled.
- _keywords_ is copied to _keywords-user_ which then has the correct unmask keywords appended.
- _keywords-user_ and _useflags_ are copied to the correct directories.
- Sync portage then update.
- Copy over and digest all custom live ebuilds. 
- Loop through all illogical-impulse ebuild to digest and emerge them.

### install-setup.sh (how it works)
- Basically the same as Arch
- Create the i2c group since Gentoo doesn't have that by default
- Enables bluetooth and ydotool, either for systemd or openrc
- chowns icons, konsole, hypr, and quickshell since those get emerge in as root

### Note
GCC is not dealt with in this. The user has to make sure themselves that they're using the correct GCC version (GCC-15 currently), and if not, change version and re-emerge world with an empty tree.


clsty:
Thank you. I'll review this when I'm available.

dist-gentoo/install-deps.sh
clsty, line 1: > GCC is not dealt with in this. The user has to make sure themselves that they're using the correct GCC version (GCC-15 currently), and if not, change version and re-emerge world with an empty tree.

I do not know about this clearly, but since user need to take care of this, please give a message in this script like this (just example, please modify as you wish):
```bash
  printf "${STY_YELLOW}"
  printf "===NOTE===\n"
  printf "Your GCC version info: $(gcc --version | grep gcc)\n"
  printf "Recommended GCC version: GCC-15\n"
  printf "You must make sure by yourself that correct GCC version is being used.\n"
  printf "If not, change version and re-emerge world with an empty tree.\n"
  printf "\n"
  printf "${STY_RESET}"
  pause
```

P.S. I've noticed [example](https://stackoverflow.com/questions/57458169/get-gcc-version-number-and-compare-with-bash) using `gcc --version | grep gcc | awk '{print $3}'` to get version number, but it does not always work, for example on Debian 13:
```plain
$ gcc --version | grep gcc
gcc (Debian 14.2.0-19) 14.2.0

$ gcc --version | grep gcc | awk '{print $3}'
14.2.0-19)
```
There is probably no reliable way in getting gcc version number, so just use `gcc --version | grep gcc` I guess.

clsty:
And two tips:
- You can use `./install.sh --skip-allgreeting --skip-allsetups --skip-allfiles` to test the `install-deps.sh` quickly. Note that `--skip-allgreeting` requires 5ecae9662f822b445c552ba279712f21d0aa2200 .
- (Optional) If you'd like to, you can add a `dist-gentoo/README.md`. Example content:
  - "About" section, such as contributor (yourself) with the lastest contribute time.
  - Information for future maintainers. I guess this part better goes in comments inside the script files though.
  - Information for users. To leave information for user, we should better use the script so users will know what they need when running the script immediately. However, if the information is too much, e.g. the instructions are too complex (we should avoid this situation from happening though), then the script should tell user to read dist-gentoo/README.md for more information.

jwihardi:
on it :)

jwihardi:
Should be Gucci for now.

clsty:
Thanks for your contribution!

PR #2117, opened 2025-10-03
Improve cookie clock style

vaguesyntax:
## Possible Combinations: 

image
image
image
image
image
image

## Settings Menu:
image

## Animations:
https://github.com/user-attachments/assets/b3b20e80-2bca-4f5d-a762-d57b51f5a6e7

## Changes:

- Tweakable settings for cookie clock style
- Supports 12 hour clock format (AM/PM)
- Quote for cookie clock style:
image

- Quote can now be edited , shown and hidden in settings
image 
image

Some settings cannot be enabled while their linked options are active: 
image

Yes it's ready for review.

### DIFF
--- a/.config/quickshell/ii/modules/settings/InterfaceConfig.qml
+++ b/.config/quickshell/ii/modules/settings/InterfaceConfig.qml
@@ -181,8 +181,8 @@ ContentPage {
                         value: "classic"
                     },
                     {
-                        displayName: Translation.tr("circle"),
-                        icon: "stroke_partial",
+                        displayName: Translation.tr("Hollow"),
+                        icon: "circle",
                         value: "hollow"
                     },
                     {


end-4:
Instead of having a new style name, just have options to tweak the params of the clock

vaguesyntax:
https://github.com/user-attachments/assets/1d974cf0-d3ff-421d-9983-add8f3e00dfa



I have added:
- Center glow 
- Transition animations
- Wave animation ( i have used a timer with 16 interval. if it effects performance too much , i can remove it)
- Dot to show seconds
- AM/PM inside clock

hexqubit:
> recording_2025-10-04_18.15.11.mp4 
> I have added:
> 
> * Center glow
> * Transition animations
> * Wave animation ( i have used a timer with 16 interval. if it effects performance too much , i can remove it)
> * Dot to show seconds

AM/PM missing?

vaguesyntax:
> > recording_2025-10-04_18.15.11.mp4
> > I have added:
> > 
> > * Center glow
> > * Transition animations
> > * Wave animation ( i have used a timer with 16 interval. if it effects performance too much , i can remove it)
> > * Dot to show seconds
> 
> AM/PM missing?


https://github.com/user-attachments/assets/172be736-5121-4e16-bc41-3e2cb9a9d8d2

It's supported.
Clock font size shrinks to fit in the frame if AM/PM is enabled.

.config/quickshell/ii/modules/common/Config.qml
```diff
@@ -129,10 +129,24 @@ Singleton {
...
+                        property bool centerGlow: true
+                        property bool dateInClock: true
+                        property bool waveAnimation: true
+                    }
+                    
                 }
                 property string wallpaperPath: ""
                 property string thumbnailPath: ""
                 property string quote: ""
+                property bool showQuote: true
```
end-4: Hmm since the quote is empty by default anyway let's have `showQuote` default to `false`

.config/quickshell/ii/modules/common/widgets/MaterialCookie.qml
```diff
@@ -1,20 +1,45 @@
...
 import QtQuick.Shapes
 import Quickshell
+import qs.modules.common
+
 
 Item {
     id: root
     
     property int sides: 12
+    property real animatedSides: 12
```
end-4: What's the point of adding animatedSides? Just make `sides` a `real` then apply a Behavior on it if necessary.

.config/quickshell/ii/modules/common/widgets/MaterialCookie.qml
```diff
@@ -1,20 +1,45 @@
...
+    Loader{
+        active: waveAnimation
+        sourceComponent: Timer{
+            interval: 16  // Does it effect performance, probably, is it noticeable, not really
+            running: true; repeat: true
+            onTriggered: {
+                root.waveTime += 0.05
+            }
+        }
+    }
```
end-4: 1. Use FrameAnimation: https://doc.qt.io/qt-6/qml-qtquick-frameanimation.html#details
2. This "wave time" is not a wave, it just spins the cookie, so call it `shapeRotation` (`rotation` would conflict with the default prop of an Item and would rotate also children)

.config/quickshell/ii/modules/settings/InterfaceConfig.qml
```diff
@@ -50,11 +64,286 @@ ContentPage {
...
                         value: "cookie"
-                    },
+                    }
                 ]
             }
         }
 
+
+        ConfigRow{
+            visible: Config.options.background.clock.style === "cookie"
```
end-4, line 126: Don't simply use a `visible` like this, it'll cause an abrupt jump when clock style is set to cookie. For now just remove it, and we can add an animation for dynamically showing items in the settings later

.config/quickshell/ii/modules/settings/InterfaceConfig.qml
```diff
@@ -50,11 +64,286 @@ ContentPage {
...
+                    ]
+                }
+            }
+            
+        }
+
+        
+            ContentSubsection {
+                enabled: Config.options.background.clock.style === "cookie"
+                visible: Config.options.background.clock.style === "cookie"
```
end-4: In cases like this, bind one to the other to reduce copy-pasting
But as mentioned in the previous comment, please remove `visible`. Same for the toggles below.

.config/quickshell/ii/modules/settings/InterfaceConfig.qml
```diff
@@ -50,11 +64,286 @@ ContentPage {
...
+                            value: "hide"
+                        }
+
+                    ]
+                }
+            }
+        
+
+        
+
```
end-4: Why the huge gap

.config/quickshell/ii/modules/settings/InterfaceConfig.qml
```diff
@@ -50,11 +64,286 @@ ContentPage {
...
+                            icon: "bubble_chart",
+                            value: "bubble"
+                        }
+                    ]
+                }
+            }
+        
+
+        ConfigSpinBox {
+            enabled: Config.options.background.clock.style === "cookie"
```
end-4: And this weird indentation

.config/quickshell/ii/modules/settings/InterfaceConfig.qml
```diff
@@ -50,11 +64,286 @@ ContentPage {
...
+            MaterialTextArea {
+                Layout.fillWidth: true
+                placeholderText: Translation.tr("Quote")
+                text: Config.options.background.quote
+                wrapMode: TextEdit.Wrap
+                onTextChanged: {
+                    Config.options.background.quote = text;
+                }
+            }
+        }
```
end-4, line 323: Since you also added a showQuote option, a toggle above this text field for it would be appropriate

.config/quickshell/ii/modules/background/CookieClock.qml
```diff
@@ -13,29 +13,71 @@ import Qt5Compat.GraphicalEffects
...
+
+
     property real implicitSize: 230
     property real hourHandLength: 72
-    property real hourHandWidth: 16
+    property real hourHandWidth: 20
     property real minuteHandLength: 95
-    property real minuteHandWidth: 8
+    property real minuteHandWidth: Config.options.background.clock.cookie.minuteHandStyle === "bold" ? hourHandWidth :
+                                    Config.options.background.clock.cookie.minuteHandStyle === "medium" ? 12 : 5
```
end-4: Note to self: see if this changed default is appropriate

.config/quickshell/ii/modules/common/Config.qml
```diff
@@ -129,10 +129,24 @@ Singleton {
...
+                    property JsonObject cookie: JsonObject {
+                        property string dialNumberStyle: "dots" // Options:   "dots" , "numbers", "full" , "none"
+                        property string hourHandStyle: "fill" // Options:     "classic", "fill", "stroke", "hide"
+                        property string minuteHandStyle: "medium" // Options: "classic", "thin", "medium", "bold", "hide"
+                        property string secondHandStyle: "dot" // Options:    "dot", "line" , "none" 
+                        property string dateStyle: "rotating" // Options:     "rotating", "square", "bubble" , "none"
+                        property bool timeIndicators: true
+                        property bool centerGlow: true
+                        property bool dateInClock: true
+                        property bool waveAnimation: true
```
end-4: Animation has to be false by default. Constantly animating in the background is expensive and most of the time meaningless in the context of a tiling window manager

.config/quickshell/ii/modules/background/CookieClock.qml
```diff
@@ -13,29 +13,71 @@ import Qt5Compat.GraphicalEffects
...
+        active: Config.option.background.clock.cookie.secondDot
+        sourceComponent: Timer {
+            interval: 1000 
+            running: true;repeat: true
+            onTriggered: {
+                var now = new Date()
+                clockSecond = now.getSeconds()
+            }
+        }
+    }
```
end-4: No, no, no.
Use `DateTime.clock`. Add a config option to make the clock precision seconds instead of minute, and only show second indicators when that is enabled

.config/quickshell/ii/modules/background/CookieClock.qml
```diff
@@ -50,93 +92,531 @@ Item {
...
+                horizontalAlignment: Text.AlignHCenter
+                text: Config.options.background.quote
+                font {
+                    family: Appearance.font.family.main
+                    pixelSize: Appearance.font.pixelSize.large
+                    weight: Font.Normal
+                    italic: true
+                }
+            }
+        }
```
end-4: Do not add the quote here, it is not the clock.

.config/quickshell/ii/modules/background/CookieClock.qml
```diff
@@ -50,93 +92,531 @@ Item {
...
+        height: cookie.height
+        rotation: secondHand.rotation + 45  // +45 degrees to align with minute hand
+        opacity: Config.options.background.clock.cookie.dateStyle === "rotating" ? 1.0 : 0
+        Behavior on opacity {
+            animation: Appearance.animation.elementMoveFast.numberAnimation.createObject(this)
+        }
+        onPaint: {
+            var ctx = getContext("2d");
+            ctx.clearRect(0,0,width,height);
+            ctx.font = "700 30px gabarito";
```
end-4: Do not hard code, use the font set in Appearance

.config/quickshell/ii/modules/common/Config.qml
end-4: I mean, look at this
image

.config/quickshell/ii/modules/background/CookieClock.qml
end-4: Okay, I realized the problem is the quote was loaded only with the digital text clock, but we should make quote a separate thing instead of doing this. For now let's just _not_ have this here, as it is not the right place.

.config/quickshell/ii/modules/settings/InterfaceConfig.qml
end-4: Okay, I realized how improper it would be if only `enable` is used. We should let `visible` stay but the config value accessing should be deduplicated

end-4:
Questionable readability:
image

.config/quickshell/ii/modules/settings/InterfaceConfig.qml
vaguesyntax, line 1: I intentionally made it so that some dial options and date options can’t be enabled at the same time, because having both active in certain cases can cause conflicts. But I guess the method I used wasn’t very logical.
image

end-4: Right... but "Full" and "Dots" should be fine, you were only allowing it for "None"

end-4:
> Questionable readability:

Okay, it might just be that some color schemes are bad. Your updated values make sense but it's still not great

vaguesyntax:
> > Questionable readability:
> 
> Okay, it might just be that some color schemes are bad. Your updated values make sense but it's still not great

But they look worse in my opinion
new/old

image
image

end-4:
> But they look worse in my opinion
> new/old 
> [image]

This is another case of varying schemes

end-4:
To do:
- Make the stuff moved to new files not rely on auto cascade (like declaring something in CookieClock and reference it in a HourHand without passing into a prop of HourHand)
- Fix the rotating date spacing: Maybe we could use Text for each char instead of a Canvas?

vaguesyntax:
To do (so we won’t forget):
- The rotating date indicator isn’t actually “rotating” (fixed with 0e89e8d) (find a better name?).

vaguesyntax:
> * Fix the rotating date spacing: Maybe we could use Text for each char instead of a Canvas?

I tried using Text with a Repeater but couldn’t get it to work. Still, I guess it’s possible.

end-4:
> Add animation to the seconds hand

Actually currently I have it enabled only when constantlyRotate is enabled... Maybe we should probably rename that config option to something like "agressiveAnimation"? 

> The rotating date indicator isn’t actually “rotating” 

Maybe "border"? It's rotated only when seconds precision is enabled and the second hand is shown (find that precision option in the General tab of the Settings app)

> Show quote on Cookie Clock style

Can leave for later

> Find a way to prevent some date indicators and dial styles from conflicting (They’re being deselected for now, but you can select them again afterward.)

I think it should be fine if the bg rect of the date is of a different enough color
It's not a big deal and after all the user can select another option when they find the combination weird anyway

vaguesyntax:
> Actually currently I have it enabled only when constantlyRotate is enabled... Maybe we should probably rename that config option to something like "agressiveAnimation"?

Oh, I didn’t notice it was on when constantlyRotate was enabled. It’s up to you whether to change the option name.

> Maybe "border"? It's rotated only when seconds precision is enabled and the second hand is shown (find that precision option in the General tab of the Settings app)

Somehow it's not rotating for me (fixed with 0e89e8d )

> I think it should be fine if the bg rect of the date is of a different enough color It's not a big deal and after all the user can select another option when they find the combination weird anyway

Fair enough. Will change the bg rect color.

vaguesyntax:
Uhh, so what’s currently holding up the merge?

.config/quickshell/ii/modules/background/cookieClock/DateIndicator.qml
vaguesyntax, line 1: Should we add an option to set the cookie sides for the date bubbles?

hexqubit:
Can’t wait to try this out. Thanks for adding it!

.config/quickshell/ii/modules/background/cookieClock/MinuteHand.qml
```diff
@@ -0,0 +1,52 @@
...
+            let position = parent.width / 2 - root.handWidth / 2;
+            if (root.style === "classic") position -= 15;
+            return position;
+        }
+        width: minuteHandLength
+        height: root.handWidth
+        
+        radius: root.style === "classic" ? 2 : root.handWidth / 2
+        color: Appearance.colors.colSecondary
+        opacity: 0.9
```
end-4: what the fuck is this

.config/quickshell/ii/modules/background/cookieClock/DateIndicator.qml
end-4: First things first, use loaders to only load the necessary bits depending on the config (sorry i forgot to tell you this)

end-4:
it appears two attempts at fixing the hour hand color are still not enough
image

.config/quickshell/ii/modules/background/cookieClock/DateIndicator.qml
vaguesyntax: I can add loaders to the date bubbles, but then the animation would be gone. Should I do that?

vaguesyntax: Nvm, you already figured it out.

end-4:
To do:
- Refractor date indicator styles to new files (like we've done for each part of the clock) into a new `dateIndicator` folder in the `cookieClock` folder
- Keep fixing colors (lol)
- Fix weird am/pm when the center hour marks ("glow") thing is enabled (here i disabled the hands so it looks clear)
image

I think this might the PR with the most number of commits so far on this repo :D

vaguesyntax:
> * Fix weird am/pm when the center hour marks ("glow") thing is enabled (here i disabled the hands so it looks clear)

There’s no problem on my side. I’m not sure if it’s something related to the localization of AM/PM (in my case ÖÖ/ÖS).


image



> * Keep fixing colors (lol)

I didn’t face any color issues like you did. Again, it could somehow be a problem with my setup.

end-4:
oh it's because am/pm get a different font size
that's my bad for failing to account for different languages

vaguesyntax:
I ‘tried’ to refactor the date indicators, but I can’t say they’re perfect (they’re not). I’d recommend you take a look at them.

shadowe1ite:
Not sure if I should ask this here, but does this support moving widgets using the mouse?

If not, could you add that feature? Something like an option to lock or unlock widgets so they can be moved around.

vaguesyntax:
> Not sure if I should ask this here, but does this support moving widgets using the mouse?
> 
> If not, could you add that feature? Something like an option to lock or unlock widgets so they can be moved around.

This PR isn’t going to add that feature, but someone has already coded it and said they’re going to open a PR for it in the future.

shadowe1ite:
> > Not sure if I should ask this here, but does this support moving widgets using the mouse?
> > If not, could you add that feature? Something like an option to lock or unlock widgets so they can be moved around.
> 
> This PR isn’t going to add that feature, but someone has already coded it and said they’re going to open a PR for it in the future.

Oh ok. thanks

0blivi0nis:
Since you guys are working on the clock, would it be possible to add a center clock toggle for the background as well like it does for the lock screen?

end-4:
@0blivi0nis stop asking for more in each pull request. Each PR should focus on its goal only, then other improvements can come later.

end-4:
I've gone through the marks and time column. Still gotta
- Review the hands, center dot, and date indicator
- Reorganize properties: Maybe putting everything every var in CookieClock was a bad idea? It would be a pain to offer configuration options for all numbers anyway. The colors should be there, but sizes can be hardcoded in the components
- Figure out the colors

end-4:
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

PR #2129, opened 2025-10-05
Use thumbnail or wallpaper for Appearance

baoliejuan:
## Describe your changes

fix the error message that it can't resolve video


nonetrix:
Edit: ignore what I said, works fine. Only issue now is window trails seemingly

PR #2131, opened 2025-10-06
Submap indicator

0blivi0nis:
## Describe your changes
- Add `Super+R` to resize windows with `arrow keys`
  - After pressing `Super+R`, use the `arrow keys` to resize the window
  - `Shift+arrow keys` will resize in larger steps
  - Press `ESC` or `Super+R` to go back to regular keybinds
- Add Hyprland submap indicator in the bar

| Submap Status         | Image                                                                                                       |
|-----------------------|-------------------------------------------------------------------------------------------------------------|
| No active submap      | No active submap        |
| Defined submap        | Defined submap          |
| None defined submap   | None defined submap     |

https://github.com/user-attachments/assets/4cfdc589-b5ae-4a05-b26d-6553271ca33a

The indicator currently only shows the resize and virtual-machine submap as an icon. If the icon is not defined then it shows the actual name of the submap. If you end up adding more submaps, define them in the config file, the order of icon and name should match.

 Yes, it is ready

### DIFF
--- a/.config/quickshell/ii/modules/bar/HyprlandSubmapIndicator.qml
+++ b/.config/quickshell/ii/modules/bar/HyprlandSubmapIndicator.qml
@@ -1,8 +1,8 @@
-import QtQuick
 import qs.services
 import qs.modules.common
 import qs.modules.common.widgets
-import QtQuick.Layouts // For RowLayout
+import QtQuick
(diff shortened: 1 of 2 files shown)


0blivi0nis:
Replaced by #2225

PR #2153, opened 2025-10-10
Show notification indicator on "tray"

An-dz:
## Describe your changes

This adds a notification indicator on the "tray" when there are notifications or when they are silent. Otherwise the icon is hidden.

Here's how it shows up:

image
image

Reasons behind it:
- If you silenced notifications, you might've forgotten that.
- If a notification was received when you were away, or you just didn't notice, at least there's something bringing attention to it now. That's why I added a counter, the counter is "out-of-place" so you notice there was a missing notification.

It's ready, been using for few weeks and no problems.

### DIFF
--- a/.config/quickshell/ii/modules/bar/BarContent.qml
+++ b/.config/quickshell/ii/modules/bar/BarContent.qml
@@ -305,17 +305,25 @@ Item { // Bar content region
                             animation: Appearance.animation.elementMoveFast.numberAnimation.createObject(this)
                         }
                         Rectangle {
+                            id: notifPing
                             visible: !Notifications.silent && Notifications.unread > 0
-                            anchors.right: parent.right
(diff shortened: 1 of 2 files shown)


end-4:
It should only show unread count, not total count

end-4:
I'm not a fan of how tiny the number is 
Maybe just a dot without numbers will do? Like the indicator for finished AI or booru response for the left sidebar

An-dz:
> Like the indicator for finished AI or booru response for the left sidebar

There's an indicator for that? Didn't even know it, that's pretty neat.

> I'm not a fan of how tiny the number is Maybe just a dot without numbers will do?

That's a good option, will probably use it myself. I tried with smaller instead of smallest, but then the icon gets mostly covered with the counter.

Issue #1705, opened 2025-07-23
Some keybinds stop working when i change keyboard layout


codessine:
### Step 1. Before you submit

- [x] I have read the [Troubleshooting](https://end-4.github.io/dots-hyprland-wiki/en/ii-qs/04troubleshooting/) and [Usage](https://end-4.github.io/dots-hyprland-wiki/en/ii-qs/02usage/) pages.
- [x] I've successfully updated to the latest version following the [guidance](https://end-4.github.io/dots-hyprland-wiki/en/ii-qs/01setup/#updating).
- [x] I've successfully updated the system packages to the latest.
- [ ] I've ticked the checkboxes without reading their contents

### Step 2. Quick diagnose info

<details><summary>Quick diagnose</summary>

```
<!-- Run `./diagnose` inside the repo, and paste the result here! -->
```

</details>

### Step 3. Describe the issue

I changed the keyboard layout in the hypr/hyprland/general.conf
i did this in a fresh installation of arch
i only tried french idk about other layouts

<details><summary>Logs</summary>

```
<!-- Put your log content here!-->
```

</details>

### Reminder

- [x] I agree that it's usually impossible for others to help me without my logs.

Mubieox:
Hi,
I have the same problem.

I had to manually change the keyboard shortcuts to their equivalents on a FR_AZERTY keyboard.

To change them, the method not recommended by the wiki is to directly edit the `.config/hypr/hyprland/keybinds.conf` file.

Here is the link to the wiki page that explains how to do this: https://end-4.github.io/dots-hyprland-wiki/en/ii-qs/03config/#configuring-hyprland, it's in the Note box.

PR #2157, opened 2025-10-10
Update keybinds with keycodes for workspace navigation

Kemipso:
Fixes https://github.com/end-4/dots-hyprland/issues/1705

## Describe your changes
Some keyboard layouts such as Azerty and Bépo default to symbols for the digits row, contrary to Qwerty-based layouts. Pressing Shift allows typing the actual digits instead of symbols.

Unfortunately, this breaks Workspace navigation in Hyprland configurations unless using keycodes for binding the digits row (1, 2, 3...0).

This commit does exactly that. It was tested with three keyboards, each time with 2 layouts: Azerty (French) / Qwerty (US). The physical keyboards were:
- Asus Strix Scope II 96 Wireless
- Keychron Q3 HE QMK (ISO layout)
- The keyboard integrated in the Lenovo Thinkpad P16s Gen3 laptop.

I tested with only 3 physical keyboards, on Arch with Hyprland installed (these dots, and in the past with Prasanthrangan's HyDE).
It is hard for me to tell if there would be compatibility issues with other hardware / OS.

### DIFF
--- a/.config/hypr/hyprland/keybinds.conf
+++ b/.config/hypr/hyprland/keybinds.conf
@@ -108,6 +108,7 @@ bind = Super+Alt, F, fullscreenstate, 0 3 # Fullscreen spoof
 bind = Super, P, pin # Pin
 
 #/# bind = Super+Alt, Hash,, # Send to workspace # (1, 2, 3,...)
+# We use raw keycodes because some keyboard layouts register number keys as different chars. The codes can be verified with `wev`
 bind = Super+Alt, code:10, exec, ~/.config/hypr/hyprland/scripts/workspace_action.sh movetoworkspacesilent 1 # [hidden]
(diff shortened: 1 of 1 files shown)


Kemipso:
I forgot to add reproduction steps for the issue described in #1705:

1. Update your ~/.config/hypr/custom/general.conf, and set fr (standing for "French", Azerty) as the default layout:
input {
	kb_layout = fr, us
	kb_options = grp:alt_shift_toggle
}

2. Save the changes
3. Try to toggle workspace with SUPER + 1 (Azerty will send & instead of 1)
4. Notice the workspace does not change. You may use "Alt + Shift" to toggle to US layout and still end up with broken keybindings, as hyprland interpreted wrongly the digits row during initialization.
5. To fix, change the config to load a Qwerty-like layout as the default (for example kb_layout = us, fr).

end-4:
Cool, thanks
I think we should have a small note there

PR #2187, opened 2025-10-13
Fix dominant color from least_busy_region as BGR

An-dz:
## Describe your changes

The least_busy_region.py script extracts a dominant colour, but it returns as BGR instead of RGB as Quickshell expects, so this reverses the color components so it returns it as RGB.

One can easily test this by simply creating a fully red image an running the script, you'll notice it returns the dominant colour as `#0000FE`, which is blue instead of red.

Yes


end-4:
So that's why the color detection sucked when I tried to make the clock adapt to it earlier... Thanks

PR #2190, opened 2025-10-13
Fix: Preserve chronological order of clipboard history

izumisagirii:
The fuzzy search logic was incorrectly sorting entries alphabetically when the search query was empty.
This change modifies the `fuzzyQuery` function to bypass the search and return the original, unsorted list when the input is empty.


end-4:
I don't think I've seen a case where this is a problem, but the fix makes sense. Thanks!

PR #2198, opened 2025-10-14
Don't scroll when deleting clipboard history items

An-dz:
## Describe your changes

When you delete an item in the clipboard history it scrolls to the first item. This should prevent that behaviour and just keep you where you are.

Also, I think this makes things more efficient because ScriptModel only updates the stuff with different keys now. I guess the default behaviour would compare the object itself that would always be false.

It's good.

Maybe more testing, but my tests worked fine. Writing more stuff goes back to first item, backspace also works like that.  All the possible search options worked fine. And deleting stuff on the clipboard history don't go jumping around anymore.

### DIFF
--- a/.config/quickshell/ii/modules/overview/SearchWidget.qml
+++ b/.config/quickshell/ii/modules/overview/SearchWidget.qml
@@ -403,7 +403,7 @@ Item { // Wrapper
                             return entry;
                         })
                         const commandResultObject = {
-                            key: `cmd /${root.searchingText}`,
+                            key: `cmd ${root.searchingText}`,
                             name: StringUtils.cleanPrefix(root.searchingText, Config.options.search.prefix.shellCommand).replace("file://", ""),
                             clickActionName: Translation.tr("Run"),
                             type: Translation.tr("Run command"),


.config/quickshell/ii/modules/overview/SearchWidget.qml
```diff
@@ -395,9 +399,11 @@ Item { // Wrapper
                         const appResultObjects = AppSearch.fuzzyQuery(StringUtils.cleanPrefix(root.searchingText, Config.options.search.prefix.app)).map(entry => {
                             entry.clickActionName = Translation.tr("Launch");
                             entry.type = Translation.tr("App");
+                            entry.key = entry.execute
                             return entry;
                         })
                         const commandResultObject = {
+                            key: `cmd /${root.searchingText}`,
```
end-4: why the slash here

end-4:
This seems to work, thanks

.config/quickshell/ii/modules/overview/SearchWidget.qml
An-dz: Didn't notice

PR #2214, opened 2025-10-16
Fix syntax error in if statement

TheBITLINK:
I was trying to install for the first time and just by coincidence there was a typo in the changes made just 1 hour ago. Seems to be just this line though.


clsty:
Thanks for the fix.

DenBuzz:
Insane response time, bravo!

clsty:
Just happen to check the notification.

PR #2217, opened 2025-10-16
Material Quick Panel Layout

vaguesyntax:
## Possible Combinations
|  |  |  | |
| -------- | ------- | --------  | -------- |
| image | image | image | image |

## Preview Video
https://github.com/user-attachments/assets/6ba2abbd-0f47-4d16-a6b8-af1809550439

## Settings
image

## Backend
Config file looks like this:
```
"quickToggles": {
    "material": {
      "align": "center",
      "border": true,
      "mode": "compact",
      "showBrightness": true,
      "showVolume": true,
      "sizes": [
        "1",
        "2",
        "2",
        "1",
        "2",
        "1",
        "1",
        "1",
        "2",
        "1",
        "1"
      ],
      "toggles": [
        "easyeffects",
        "network",
        "idleinhibitor",
        "bluetooth",
        "cloudflarewarp",
        "darkmode",
        "togglemic",
        "silent",
        "performanceprofile",
        "showkeyboard",
        "nightlight"
      ]
    },
    "style": "material"
  },
```

Quick toggles can also be manually edited from config.json. Edit mode makes it easier.

I think its ready for review.
*Note: I couldn't test bluetooth toggle properly because my desktop does not have an adapter for it*

### DIFF
--- a/dots/.config/quickshell/ii/modules/sidebarRight/quickToggles/androidStyle/AndroidQuickToggleButton.qml
+++ b/dots/.config/quickshell/ii/modules/sidebarRight/quickToggles/androidStyle/AndroidQuickToggleButton.qml
@@ -85,8 +85,8 @@ GroupButton {
         anchors {
             centerIn: root.expandedSize ? undefined : parent
             fill: root.expandedSize ? parent : undefined
-            leftMargin: 12
-            rightMargin: 12
+            leftMargin: 8
+            rightMargin: 8
         }
 
         Rectangle {


dots/.config/quickshell/todo
end-4, line 1: whats this

dots/.config/quickshell/ii/GlobalStates.qml
```diff
@@ -26,6 +26,7 @@ Singleton {
     property bool superDown: false
     property bool superReleaseMightTrigger: true
     property bool workspaceShowNumbers: false
+    property bool quickTogglesEditMode: false
```
end-4: I don't think this deserves to be in GlobalStates

dots/.config/quickshell/ii/modules/common/widgets/GroupButton.qml
```diff
@@ -17,13 +17,17 @@ Button {
     property real buttonRadiusPressed: Appearance?.rounding?.small ?? 6
     property var downAction // When left clicking (down)
     property var releaseAction // When left clicking (release)
+    property var rightReleaseAction // When right clicking (release)
+    property var middleReleaseAction // When middle clicking (release)
     property var altAction // When right clicking
     property var middleClickAction // When middle clicking
+    property var clickAndHold // When clicked and hold
```
end-4: Why do we need this separate clickAndHold? It was intentional that holding and right clicking do the same so it's easy to use on touch devices

dots/.config/quickshell/ii/modules/common/widgets/GroupButton.qml
```diff
@@ -39,9 +43,9 @@ Button {
     property color colBackgroundToggledHover: Appearance?.colors.colPrimaryHover ?? "#77699C"
     property color colBackgroundToggledActive: Appearance?.colors.colPrimaryActive ?? "#D6CEE2"
 
-    property real radius: root.down ? root.buttonRadiusPressed : root.buttonRadius
-    property real leftRadius: root.down ? root.buttonRadiusPressed : root.buttonRadius
-    property real rightRadius: root.down ? root.buttonRadiusPressed : root.buttonRadius
+    property real radius: root.clickedRadius > 0 && root.toggled ? root.clickedRadius :  root.down ? root.buttonRadiusPressed : root.buttonRadius  
+    property real leftRadius: root.clickedRadius > 0 && root.toggled ? root.clickedRadius :  root.down ? root.buttonRadiusPressed : root.buttonRadius
+    property real rightRadius: root.clickedRadius > 0 && root.toggled ? root.clickedRadius :  root.down ? root.buttonRadiusPressed : root.buttonRadius
```
end-4: Why is "toggled" radius called `clickedRadius`??

end-4:
The copilot review is a small sanity check only, I'll go through it myself don't worry

dots/.config/quickshell/ii/modules/sidebarRight/SidebarRightContent.qml
```diff
@@ -80,6 +95,17 @@ Item {
...
+                        visible: quickPanelStyle === "material"
+                        toggled: GlobalStates.quickTogglesEditMode
+                        buttonIcon: "edit"
+                        onClicked: {
+                            GlobalStates.quickTogglesEditMode = !GlobalStates.quickTogglesEditMode;
+                        }
+                        StyledToolTip {
+                            text: GlobalStates.quickTogglesEditMode ? Translation.tr("Middle Click: Toggle\nLeft/Right Click: Move\nHold: Toggle Size") : Translation.tr("Edit Material Panel Layout")
+                        }
+                    }
```
end-4: put a `Loader` please

dots/.config/quickshell/ii/modules/common/Config.qml
```diff
@@ -438,6 +438,20 @@ Singleton {
...
                 }
             }
+
+            property JsonObject quickToggles: JsonObject {
+                property string style: "material"
+                property JsonObject material: JsonObject {
+                    property bool border: false
+                    property bool showVolume: true
+                    property bool showBrightness: true
+                    property string mode: "compact" // Options: compact, medium, large
```
end-4: I haven't dived deep into how this works, but if it's just sizes maybe have this "mode" called "size" instead? You can still provide the [compact, medium, large] presets in the settings app but this I think can be an int

vaguesyntax:
> Why do we now have toggles in classicStyle for the old ones but the files from before are still there? Did you forget to remove the old dupes or I'm missing something?
- ~~I was having issue with git , i forget to remove them.~~ Removing them now

> For naming, use it as prefix instead of suffix (MaterialEasyEffects, MaterialIdleInhibitor instead of EasyEffectsMaterial, IdleInhibitorMaterial)
- It was called that way like 4-5 commits ago but i changed it to current one. Also "Android" is a better name i guess

> Right-clicking the new bluetooth toggle doesn't open bluetooth dialog
- My desktop still don't have a bluetooth adapter for it so i couldn't test it :(

> In edit mode, don't use the middle button for the primary action of enabling/disabling buttons, because it's the shittiest button on a mouse (it's easy to break, often require strong pressure, and for certain mice hurts to click). I think a nicer scheme would be LMB to enable/disable, Mouse4 to disable (bring down), Mouse5 to enable (bring up), RMB or hold to toggle size, scroll to change position. (If you think something's even better, feel free to suggest)
- I agree with you. The current mouse button scheme isn’t ideal. We should definitely change it. Maybe: left click for enabling/disabling, scroll for moving and right click for toggling size would be better. (*I didn’t use Mouse4-5 because someone with a mouse that doesn’t have side buttons wouldn’t be able to use that functionality*)

> Cloudflare Warp's tooltip is broken
- ~~Game mode toggle is also broken on my side but it may be related to Translation but i am not sure. It gets fixed when i write `"Game Mode"` instead of `Translation.tr("Game Mode")` (*in GameModeMaterial.qml file*)~~  fixed

dots/.config/quickshell/ii/modules/common/widgets/GroupButton.qml
vaguesyntax: I couldn’t use the current onPressAndHold, so I created this new function (my bad). We can revert it to the old way if we change the click scheme to something like `right click -> toggleSize`

end-4:
I just noticed the wiggle doesn't work correctly. Clicking one button should expand it and squish the adjacent ones, not expand the whole row. This can be fixed by setting a width for the ButtonGroup (haven't looked into the code but that's what I guess)

end-4:
Currently the sizes are kind of arbitrary. We should have a config option called "columns" and calculate the toggle size based on that

end-4:
> It was called that way like 4-5 commits ago but i changed it to current one. Also "Android" is a better name i guess

If you don't have any better idea we can go with "Android"

dots/.config/quickshell/ii/modules/common/Config.qml
```diff
@@ -438,6 +438,21 @@ Singleton {
...
+            property JsonObject quickToggles: JsonObject {
+                property string style: "material"
+                property JsonObject material: JsonObject {
+                    property bool inEditMode: false
+                    property bool border: false
+                    property bool showVolume: true
+                    property bool showBrightness: true
+                    property int mode: 0 // Options: 0: compact, 1: medium, 2: large
+                    property string align: "center" // Options: left, center, right
+                    property list<string> sizes: ["2","2"]
```
end-4: why are sizes strings if they're just 1 or 2?

vaguesyntax:
> Currently the sizes are kind of arbitrary. We should have a config option called "columns" and calculate the toggle size based on that

I’m not sure what kind of change you meant here 😅 Did you mean only the column size or the button sizes too (1,2)

end-4:
> I’m not sure what kind of change you meant here 😅 Did you mean only the column size or the button sizes too (1,2)

Whoops, I meant Compact/Medium/Large modes. I think it makes more sense to offer 4 columns and 5 column options. You might've noticed that the buttons currently don't fill the width properly

vaguesyntax:
So, uhh, I calculated the button widths like this (I’ve completely disabled compact-medium-large):
```
property int unusedButtonSize: 48
property int calculatedWidth: maxColumn == 4 ? 95 : 75
property int calculatedHeight: 55
baseWidth: unusedName === "" ? calculatedWidth * buttonSize - 5 : unusedButtonSize * 1.6
baseHeight: unusedName === "" ? calculatedHeight : unusedButtonSize
```
Is this what you meant?

dots/.config/quickshell/ii/modules/common/widgets/GroupButton.qml
end-4: Let's revert these weird new actions. press+hold and right click both trigger altAction, so use that

end-4:
no, the button width should be determined by container width minus gaps then divided by number of cols

end-4:
yo wtf 
image

vaguesyntax:
> yo wtf image

There is no problem on my side but we should change the icons they are not the greatest
image

dots/.config/quickshell/ii/modules/sidebarRight/SidebarRightContent.qml
```diff
@@ -2,30 +2,45 @@ import qs
...
+    
+    property bool wifiDialogMaterial: QuickTogglesUtils.showWifiDialog
+    property bool bluetoothDialogMaterial: QuickTogglesUtils.showBluetoothDialog
+
+    onWifiDialogMaterialChanged: {
+        root.showWifiDialog = wifiDialogMaterial
+    }
+    onBluetoothDialogMaterialChanged: {
+        root.showBluetoothDialog = bluetoothDialogMaterial
+    }
```
end-4, line 35: Pass signals out, don't do this

dots/.config/quickshell/ii/modules/common/Config.qml
```diff
@@ -438,6 +438,22 @@ Singleton {
                     property list<string> linkKeywords: ["hentai", "porn", "sukebei", "hitomi.la", "rule34", "gelbooru", "fanbox", "dlsite"]
                 }
             }
+
+            property JsonObject quickToggles: JsonObject {
+                property string style: "android"
+                property JsonObject android: JsonObject {
+                    property bool inEditMode: false
```
end-4, line 445: This is an even worse place than GlobalStates. How is edit state a config option?? Put it somewhere in the sidebar or quick toggle panel

dots/.config/quickshell/ii/modules/sidebarRight/quickToggles/androidStyle/AndroidQuickToggleButton.qml
```diff
@@ -0,0 +1,149 @@
...
+            width: calculatedWidth * 3 - calculatedWidth
+            StyledText {
+                anchors.bottom: parent.verticalCenter
+                anchors.left: parent.left
+                text: titleText
+                color: button.colText
+                font {
+                    family: Appearance.font.family.title
+                    pixelSize: titleTextSize
+                    weight: 500
```
end-4: font weight enums exist

dots/.config/quickshell/ii/modules/sidebarRight/quickToggles/androidStyle/AndroidQuickToggleButton.qml
```diff
@@ -0,0 +1,149 @@
...
+        }
+    }
+
+
+    Loader {
+        active: buttonSize === 2
+        sourceComponent: Item {
+            anchors.left: parent.left
+            anchors.leftMargin: 50
+            anchors.verticalCenter: parent.verticalCenter
```
end-4:
```qml
anchors {
    bottom: ...
    left: ...
}
```

dots/.config/quickshell/ii/modules/sidebarRight/quickToggles/androidStyle/AndroidScreenSnipToggle.qml
```diff
@@ -0,0 +1,22 @@
...
+import qs.services
+import Quickshell
+import "../"
+
+AndroidQuickToggleButton {
+    id: root
+    toggled: false
+    buttonIcon: "screenshot_region"
+    titleText: "Screenshot"
+    descText: "Click me"
```
end-4: 1. Use `Translation.tr`
2. Either have a meaningful description or not have it

dots/.config/quickshell/ii/modules/sidebarRight/quickToggles/androidStyle/AndroidToggles.qml
```diff
@@ -0,0 +1,51 @@
...
+                onAlignmentChanged: {
+                    if (alignment === "left") Layout.alignment = Qt.AlignLeft
+                    if (alignment === "right") Layout.alignment = Qt.AlignRight
+                    if (alignment === "center") Layout.alignment = Qt.AlignCenter
+                }
+                ButtonGroup {
+                    id: grid
+                    
+                    
+
```
end-4: weird blank space

dots/.config/quickshell/ii/modules/sidebarRight/SidebarRightContent.qml
```diff
@@ -115,33 +144,24 @@ Item {
...
-                    altAction: () => {
-                        Bluetooth.defaultAdapter.enabled = true;
-                        Bluetooth.defaultAdapter.discovering = true;
-                        root.showBluetoothDialog = true;
-                    }
+                Loader { 
+                    id: classicPanelLoader
+                    anchors.horizontalCenter: parent.horizontalCenter
+                    active: quickPanelStyle === "classic"
+                    sourceComponent: ClassicQuickPanel {}
```
end-4, line 161: Wifi/Bluetooth dialog open is broken for classic toggle panel

end-4:
Okay, so the wiggly buttons wiggle the wrong way is because they're put in a loader. A GroupButton when attempting to change`clickIndex` prop of ButtonGroup would fail because it tries to do that on the loader instead of the ButtonGroup

dots/.config/quickshell/ii/modules/sidebarRight/SidebarRightContent.qml
vaguesyntax: i think i fixed the wifi popup but i don't know bluetooth popup is fixed or not, i can't test it

end-4:
Please, never, ever, manually set `width` (unless you're doing masking/effects, which is not the case here). Anchors exist for sizing an element relative to its parent, and `implicitWidth` is the right property to set if the size should be reported to parents

vaguesyntax:
image

this one pixel is pissing me off..😑

end-4:
The QuickToggleUtils makes no sense. Why is it a singleton?? And somehow I can't find how the wifi/bluetooth toggling are handled

vaguesyntax:
> The QuickToggleUtils makes no sense. Why is it a singleton?? And somehow I can't find how the wifi/bluetooth toggling are handled

I made it singleton for easier access across different files while testing, we can revert it to a normal component/file now if you want. Network and bluetooth toggles are handled in ther individiual files like this: 
```
toggled: Network.networkName.length > 0 && Network.networkName != "lo"
halfToggled: Network.wifiEnabled
onClicked: {
    if (Config.options.quickToggles.android.inEditMode) return;
    Network.toggleWifi()
}
```

And also currently, buttons with the extra altAction function (like network, bluetooth, nightlight and easyeffects) can’t be resized.

end-4:
I decided to rewrite the way the two quick toggle implementations are loaded. I've included your new sliders, ported over the newly added toggle buttons, and made sure to give you credit (the commit is co-authored)
Thank you for your work!

PR #2219, opened 2025-10-16
feat: Add screen recording button and config toggle

Cleboost:
## Describe your changes

This pull request adds a new screen recording utility button to the bar, along with configuration options to control its visibility. The main changes include updating the configuration, UI components, and settings page to support this new feature.

**Screen recording utility button feature:**

* Added a new `showScreenRecord` property to `Config.options.bar.utilButtons` to allow toggling the visibility of the screen recording button.
* Introduced a new `Loader` in `UtilButtons.qml` that displays a screen recording button when enabled, which executes the `record.sh` script with fullscreen sound on click.

**Settings integration:**

* Added a new configuration row in `BarConfig.qml` to allow users to enable or disable the screen recording button from the settings UI.

Ready.. I think?

### DIFF
--- a/dots/.config/quickshell/ii/modules/settings/BarConfig.qml
+++ b/dots/.config/quickshell/ii/modules/settings/BarConfig.qml
@@ -229,7 +229,7 @@ ContentPage {
             uniform: true
             ConfigSwitch {
                 buttonIcon: "videocam"
-                text: Translation.tr("Screen recording")
+                text: Translation.tr("Record")
                 checked: Config.options.bar.utilButtons.showScreenRecord
                 onCheckedChanged: {
                     Config.options.bar.utilButtons.showScreenRecord = checked;


dots/.config/quickshell/ii/modules/bar/UtilButtons.qml
```diff
@@ -36,6 +36,22 @@ Item {
             }
         }
 
+        Loader {
+            active: Config.options.bar.utilButtons.showScreenRecord
+            visible: Config.options.bar.utilButtons.showScreenRecord
+            sourceComponent: CircleUtilButton {
+                Layout.alignment: Qt.AlignVCenter
+                onClicked: Quickshell.execDetached(["bash", "-c", "~/.config/hypr/hyprland/scripts/record.sh --fullscreen-sound"])
```
end-4: - `bash -c` not needed
- we should not use --fullscreen-sound so it's consistent with screen snip

dots/.config/quickshell/ii/modules/common/Config.qml
```diff
@@ -211,6 +211,7 @@ Singleton {
                     property bool showKeyboardToggle: true
                     property bool showDarkModeToggle: true
                     property bool showPerformanceProfileToggle: false
+                    property bool showScreenRecord: true
```
end-4: off by default

dots/.config/quickshell/ii/modules/bar/UtilButtons.qml
end-4: > bash -c not needed

nvm it's needed to expand the `~`

PR #2222, opened 2025-10-17
Fix escape keybind notification for virtual machines

Alvin-HZ:
## Describe your changes

Fixes keybind mentioned in notification


end-4:
thanks

PR #2223, opened 2025-10-17
✨ feat: sound alerts for battery and pomodoro

0blivi0nis:
## Describe your changes

- Battery Alerts: Implemented audio notifications for battery thresholds
- Pomodoro Alerts: Added chime sounds for Pomodoro timer events
- Settings: Option to toggle alert sounds & change battery full warning

settings

Yes, has been tested

### DIFF
--- a/dots/.config/quickshell/ii/services/Battery.qml
+++ b/dots/.config/quickshell/ii/services/Battery.qml
@@ -8,12 +8,14 @@ import QtQuick
 import Quickshell.Io
 
 Singleton {
+    id: root
     property bool available: UPower.displayDevice.isLaptopBattery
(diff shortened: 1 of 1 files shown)


dots/.config/quickshell/ii/services/Battery.qml
```diff
@@ -35,22 +37,58 @@ Singleton {
...
             "-u", "critical",
             "-a", "Shell"
         ]);
-            
+
+        if (available && Config.options.sounds.battery) {
+            if (isCriticalAndNotCharging) {
+                Audio.playSystemSound("suspend-error")
+            }
+        }
```
end-4: This will trigger every change, not only when it's not charging
Same with the full one below

PR #2250, opened 2025-10-19
Draggable background widgets

### LINKED ISSUE #1867: A way to manually move the clock?

hey, i just recently started trying linux again and the first thing i did was to install your rice because the last time i used it i instantly fell in love because i'm a huge fan of material you rices, please keep up your outstanding work! anyways i just think it would be nice if there is a way to move the clock manually as sometimes i find that the location is a bit unfavorable depending on each wallpaper i use

### How will it help?

Image

like here, i would prefer the clock to go to the top left more corner as there is not much detail there in the wallpaper, maybe make it so you can drag it while you're holding super and you can use super + right click on it to reset it back to default position? idk if this would be tricky to implement since the color would also have to change depending on the location aswell but it would be nice if this is possible, thank you

### Extra info

_No response_

vaguesyntax:
## Describe your changes
Closes #1867
- Added the ability to drag background widgets
- Added weather widget (to make it easier to understand and test the feature this PR adds)
- Added a ~~Widgets~~ **Background** tab to settings (to fix the cluttered “interface” tab)
- Also fixed the clock colors 
(they were wrong because of the matugen update ig)

image

## Preview (old)
https://github.com/user-attachments/assets/92fff195-f4cd-429b-bcdd-7c4970a83c1c

Yes, I think it’s ready — ~~except for one thing: I removed the part that finds the most suitable position using the Python script, since I thought we don’t need it anymore as users can now move widgets to most suitable position themselves.~~

~~BUG:Clock widget is not getting centered on the first boot~~ Fixed with b1df572

### DIFF
--- a/dots/.config/quickshell/ii/modules/background/widgets/ClockWidget.qml
+++ b/dots/.config/quickshell/ii/modules/background/widgets/ClockWidget.qml
@@ -22,7 +22,7 @@ BackgroundWidget {
     widgetPrefix: "clock"
 
     property color dominantColor: widget.collectorData.dominant_color
-    property bool dominantColorIsDark: dominantColor.hslLightness < 0.5
+    property bool dominantColorIsDark: dominantColor.hslLightness < 0.85
     property color colText: {
         return (GlobalStates.screenLocked) ? Appearance.colors.colOnLayer0 : CF.ColorUtils.colorWithLightness(Appearance.colors.colPrimary, (dominantColorIsDark ? 0.8 : 0.12));
     }
--- a/dots/.config/quickshell/ii/modules/background/widgets/cookieClock/CookieClock.qml
+++ b/dots/.config/quickshell/ii/modules/background/widgets/cookieClock/CookieClock.qml
@@ -20,9 +20,9 @@ Item {
     property real implicitSize: 230
 
     property color colShadow: Appearance.colors.colShadow
-    property color colBackground: Appearance.colors.colPrimaryContainer
-    property color colOnBackground: ColorUtils.mix(Appearance.colors.colSecondary, Appearance.colors.colPrimaryContainer, 0.15)
-    property color colBackgroundInfo: ColorUtils.mix(Appearance.colors.colPrimary, Appearance.colors.colPrimaryContainer, 0.55)
+    property color colBackground: Appearance.colors.colSecondaryContainer
+    property color colOnBackground: ColorUtils.mix(Appearance.colors.colPrimary, Appearance.colors.colSecondaryContainer, 0.15)
+    property color colBackgroundInfo: ColorUtils.mix(Appearance.colors.colSecondary, Appearance.colors.colSecondaryContainer, 0.55)
     property color colHourHand: Appearance.colors.colPrimary
     property color colMinuteHand: Appearance.colors.colTertiary
     property color colSecondHand: Appearance.colors.colPrimary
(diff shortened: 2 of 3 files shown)


Davidekiller1999:
I like it very much. I have a question: Since the watch is positioned automatically on the screen, what happens with this setting? Widgets can be "locked" to prevent custody movement?

vaguesyntax:
> I like it very much. I have a question: Since the watch is positioned automatically on the screen, what happens with this setting? Widgets can be "locked" to prevent custody movement?

~~The clock doesn’t automatically move to the _most suitable position_ with this PR (I removed that part).~~
added with  9b0ef15
The locking feature lets you lock a widget’s position so you don’t accidentally move it by misclicking.

dots/.config/quickshell/ii/modules/background/widgets/weather/WeatherRect.qml
```diff
@@ -0,0 +1,72 @@
...
+        color: Appearance.colors.colShadow
+        rotation: -widget.widgetRotation
+    }
+
+    Rectangle {
+        id: weatherRect
+        anchors.fill:parent
+        radius: Appearance.rounding.full
+        color: widget.colBackground
+        rotation: -widget.widgetRotation
```
end-4: Since we now have the material shapes library, use a `MaterialShape` with `shape: MaterialShape.Shape.Pill` so we don't have to deal with rotations

dots/.config/quickshell/ii/modules/background/widgets/weather/WeatherSymbols.qml
end-4, line 1: This is the exact same as `modules/bar/weather/WeatherIcons.qml`. Don't do this, put the icons into `modules/common/Icons.qml` instead (as a function, similar to the bluetooth device one)

dots/.config/quickshell/ii/modules/background/Background.qml
```diff
@@ -134,47 +134,12 @@ Variants {
...
-                    // console.log("[Background] Least busy region output:", output)
-                    if (output.length === 0)
-                        return;
-                    const parsedContent = JSON.parse(output);
-                    bgRoot.clockX = parsedContent.center_x * bgRoot.effectiveWallpaperScale;
-                    bgRoot.clockY = parsedContent.center_y * bgRoot.effectiveWallpaperScale;
-                    bgRoot.dominantColor = parsedContent.dominant_color || Appearance.colors.colPrimary;
-                }
-            }
-        }
```
end-4, line 177: Least busy region positioning must not be removed. Ideally we should be able to somehow in the config specify how each widget is placed. Something like a `placementStrategy` prop that can be either "free" or "leastBusy" for each of them

dots/.config/quickshell/ii/modules/settings/BackgroundConfig.qml
end-4, line 1: I think this is more appropriately named `BackgroundConfig`

dots/.config/quickshell/ii/modules/background/Background.qml
vaguesyntax: I will try implementing it in the new background widgets system , but I have one idea in mind:
I think it’s better to move `placementStrategy` into the a Config property as something like `leastBusyPlacedWidget: "clock" or "weather"`, because we can’t have two widgets both trying to go to the least busy position — they’d end up competing for the same spot.

vaguesyntax:
Uhh, end-4, I _tried_ to add the least busy region placement function back. It works - _kinda_. You should **definitely** take a look at it and fix the missings.

end-4:
I implemented the draggables in a more scalable way (least busy region can work with weather widget too)
I included your weather widget and the relocation of weather icons and made sure to credit you for that

Thanks for your contribution

end-4:
the commit range is 0e7422c335f54ab440ca02a294e32c930d5e3099 - 64f6081b149de14a75b84073678a306f44e84a98

Issue #2137, opened 2025-10-06
`sdata/subcmd-install/3.files.sh` enhancement


clsty:
### What would you like to be added?

Enhance [sdata/step/3.install-files.sh](https://github.com/end-4/dots-hyprland/blob/main/sdata/step/3.install-files.sh) in handling
- backup
- copy&overwrite
- output log

Quoting what I said in #2120:

> It's ugly because there're several complex snippets (`if else` thingy) which structure is similar, which means it should better be done via functions.
> 
> And most importantly, from the first place, it's not good to generate a pile of `*.new` or `*.old` garbages to deal with file conflicts.
> 
> I'll find out a better way, for example a customisable `syncpattern.conf` file containing information about how to deal with each path, each line of which contains two elements, one is file/directory path, another is syncing mode.

### How will it help?

**A lot.**

- #2133
- #2120
- #1863
- #1438
- #1522 
  - Talking about fcitx5 though, I think this repo should not touch fcitx5 thingy from the first place, nor should we add fcitx5 as dependency. See also #2141
- https://github.com/end-4/dots-hyprland/issues/1544#issuecomment-3020368549

### Extra info

- Currently the script will ask each step before using rsync but the level of automation is not high enough.
- We are not using any symlinks tool, like dotbot. It's super annoying for people only customed to the old-fashioned modification method on dotfiles, according to my previous experience.

clsty:
## Current ideas

### yaml config
About how to implement this:
> I'll find out a better way, for example a customisable syncpattern.conf file containing information about how to deal with each path, each line of which contains two elements, one is file/directory path, another is syncing mode.

Using a `syncpattern.conf` containing 2 elements each line is not extendable at all. I think a `syncpattern` file in `*.yaml` or `*.toml` can be used.

If we stick to Bash then `yq` can be used to parse `yaml`, but if things get more complicated we maybe need python... Better avoid that though.

### version number

Should have a version number inside the yaml configuration file. Everytime it has breaking change the major version number should increase, so that the install script can abort when the major version number is not the same.

### Flavor?
As we all know that not all users of this dotfiles use fish or kitty, or windows-like keybindings.
- As for myself, I'm deeply customed to Zsh and Vim-like keybindings. It's just like the water for fish and I can't live without it.
- Other examples are #2175 .
- In fact I has been even asked about why end-4 is not using Vim-like keybindings when I discuss with others in an Arch Linux chat group in Chinese.

[...]

PR #2278, opened 2025-10-24
feat(nix): add nix home-manager flake support

Dexter2038:
## Describe your changes
Add nix package manager support based on flake-approach.

Yes. It is ready. I built it with `nix build .` for flake.lock to appear so dependencies are locked.
I checked it with `home-manager switch --flake .#my-username` and I guess it works, status successful,
however I decided to remove `microtex` package because I don't know how to configure it,
I tried it, but couldn't handle it, file is there if anyone is willing to fix it further.
Everything is done declaratively, the "nix way". No scripts. So, no rsync, but symlinks.


clsty:
Nice. I'll review this when I'm available.

sdata/dist-nix/quickshell-packages.nix
```diff
@@ -0,0 +1,46 @@
...
+    wayland-protocols = pkgs.wayland-protocols;
+    wayland-scanner = pkgs.wayland-scanner;
+    xorg = pkgs.xorg;
+    libdrm = pkgs.libdrm;
+    libgbm = pkgs.libgbm;
+    pipewire = pkgs.pipewire;
+    pam = pkgs.pam;
+
+    # Set gitRev to a fixed value since we're using a specific commit
+    #gitRev = "3e2ce40b18af943f9ba370ed73565e9f487663ef";
```
clsty, line 29: We do need to pin the version.

Why is this commented out? Does it work?

EryyynIT:
image
There are some problems, but generally it works.
Some of them, I noticed:
- Themes are not being applied
- Screenshot shortcut doesn't work
- Config files are read only and it sometimes causes problems

sdata/dist-nix/flake.nix
```diff
@@ -0,0 +1,389 @@
+{
+  description = "end_4's Hyprland dotfiles";
```
clsty, line 2:
```suggestion
  description = "illogical-impulse at end-4/dots-hyprland";
```

sdata/dist-nix/flake.nix
```diff
@@ -0,0 +1,389 @@
...
+      widgetsPackages = import ./widgets-packages.nix { inherit pkgs; };
+      hyprlandPackages = import ./hyprland-packages.nix { inherit pkgs; };
+      #microtexPackages = import ./microtex-packages.nix { inherit pkgs microtex-src; };
+      quickshellPackages = import ./quickshell-packages.nix { inherit pkgs lib quickshell-src; };
+
+      # Combine all packages
+      allPackages = basicPackages // audioPackages // backlightPackages // fontsThemesPackages // kdePackages // portalPackages // pythonPackages // screencapturePackages // toolkitPackages // widgetsPackages // hyprlandPackages // quickshellPackages;
+
+      # Dotfiles configuration
+      dotfilesConfig = {
```
clsty, line 64: The config files... They surely will cause problem cuz they are read-only.

sdata/dist-nix/install-deps.sh
```diff
@@ -2,3 +2,12 @@
 # It's not for directly running.
 
 # This file is currently WIP.
+printf "[$0]: The repository has flake for Nix with Home-manager.
+
+In path 'sdist/nix/flake.nix'
+
+Search for 'FIXME' in flake.nix
```
clsty, line 9: This is far from enough automation. `install-deps.sh` should handle nearly everything including installing Nix if needed for non-NixOS distros.

sdata/dist-nix/hyprland-packages.nix
```diff
@@ -0,0 +1,26 @@
+# Hyprland compositor and related packages
+{ pkgs }:
+
+{
+  # Hyprland core
+  hyprland = pkgs.hyprland;
```
clsty, line 6: For hyprland NixGL is a must.

clsty:
I appreciate your effort but there are two major problems:
- `install-deps.sh` is far from enough automation.
- As declared in [./sdist/nix/README.md](https://github.com/end-4/dots-hyprland/blob/main/sdist/nix/README.md), `./sdist/nix` is for non-NixOS distros. You should at least test this on a non-NixOS distro like Fedora.
  - If you're contributing for installation on NixOS, as declared in https://ii.clsty.link/en/dev/inst-script/ , since `nixos` is the distro ID of NixOS (see https://github.com/which-distro/os-release/blob/main/nixos/25.05 ), you should contribute to `./sdist/nixos` instead. And again please make sure it has enough automation.

EryyynIT:
Fair, ok, I'll add automation to install script, make necessary changes to configuration and will test it on Debian. 

About pinned version, I thought at start to make it pinned, but flake.lock handles it, so I let it be just latest git version

clsty:
> About pinned version, I thought at start to make it pinned, but flake.lock handles it, so I let it be just latest git version

`flake.lock` is not something that should be edited manually, isn't it? Thus it does not fulfill the requirement cuz illogical-impulse is developed only under the **specific** commit of quickshell.

clsty:
> will test it on Debian

If you're testing it on Debian, please be aware that software installed by Nix will not work as expected if it relies on PAM, for example screenlock. See the links below:
- https://github.com/end-4/dots-hyprland/issues/1061#issuecomment-3354619429
  - (This comment also contains example config which uses nixGL)
- https://github.com/end-4/dots-hyprland/issues/1061#issuecomment-3403195230

EryyynIT:
> > About pinned version, I thought at start to make it pinned, but flake.lock handles it, so I let it be just latest git version
> 
> `flake.lock` is not something that should be edited manually, isn't it? Thus it does not fulfill the requirement cuz illogical-impulse is developed only under the **specific** commit of quickshell.

Yes, definitely lock files shouldn't be edited manually, cuz it'll break dependencies and so on.
Specific commit, ok. So it should be pinned. So, that's the variable that will pin needed quickshell commit. Just as needed.


> > will test it on Debian
> 
> If you're testing it on Debian, please be aware that software installed by Nix will not work as expected if it relies on PAM, for example screenlock. See the links below:
> 
> * [Use guix (or nix) to repack almost everything #1061 (comment)](https://github.com/end-4/dots-hyprland/issues/1061#issuecomment-3354619429)
>   
>   * (This comment also contains example config which uses nixGL)
> * [Use guix (or nix) to repack almost everything #1061 (comment)](https://github.com/end-4/dots-hyprland/issues/1061#issuecomment-3403195230)

Nice, then I have a change to take a look how it works (or not). I'll definitely do it on Debian.

About nixGL, should I nixGL'ize everything in dotfiles that uses GPU or just Hyprland? Stupid question, obviously everything.

Cu3PO42:
Hi! I'm a relatively experienced Nix user and happy to provide some feedback as well, if desired.

clsty:
> Hi! I'm a relatively experienced Nix user and happy to provide some feedback as well, if desired.

Actually I don't think we should count on this PR. It's obvious that it was for NixOS from the beginning, and Nix on non-NixOS distros just differs from Nix on NixOS.

@Dexter2038 TBH I suggest you to contribute to installation on NixOS, i.e. `sdata/dist-nixos/` instead. You didn't plan to contribute to Nix on non-NixOS from the beginning anyway.

@Cu3PO42 If you're willing to help for Nix on non-NixOS distros, comments on #1061 and PR are both welcomed.

EryyynIT:
Sorry, but I actually counted on Nix. From the start I, actually, wanted to make NixOS installation. But I've found that there is Nix installation script. I've read comments and got that it should be available not from only NixOS, but on any distro that has support for Nix package manager. So, I converted it to Home-manager. I will contribute to nixos installation, it's actually even easier for m, and much faster as it doesn't requires as much testing.

Cu3PO42:
So I agree there is a number of different settings in which Nix can be potentially used to install illogical-impulse on a system.

1. Directly as a NixOS module.
2. As a Home-Manager module on a NixOS-host.
3. As a Home-Manager module on a non-NixOS host.
4. Install applications directly into the user's profile, do not manage config settings.

In my opinion, 1. is a sub-optimal approach for the most part. There are some settings that need to be performed at NixOS-level, such as group configurations for ydotool or setting the user's shell. These settings can (and probably should) at some point be provided as a NixOS module. This is unrelated to the current PR and need not be considered at this point.

Installation of settings and most user-facing applications fit best in HM in my opinion, even on NixOS. This is option 2 above and the current focus of this PR.

Now, HM on non-NixOS can and should share almost all code with HM on NixOS. Notable differences include having nixGL wrappers as well as workarounds for PAM and Polkit. I have solutions for both in the works that do not require falling back to distro packages, they just need to be enabled.

Finally, option 4: installing just packages directly into the profile. This is the closest approximation of the installation experience on most other distributions. However, installing various dependencies globally is in some ways against the spirit of Nix. It would be ideal to create derivations bundling them with the various parts that need them. This is also good practice for HM, so again, work can be shared.

In general, we should strive to share as much code as possible between any of the variants. I see elements of that here already, but in particular the bundling of dependencies could be improved.

sdata/dist-nix/flake.nix
```diff
@@ -0,0 +1,389 @@
...
+    };
+
+    #microtex-src = {
+    #  url = "github:NanoMichael/MicroTeX";
+    #  flake = false;
+    #};
+
+    quickshell-src = {
+      url = "github:quickshell-mirror/quickshell";
+      flake = false;
```
Cu3PO42, line 32: Quickshell provides a perfectly reasonable Flake. Why not use that?

sdata/dist-nix/fonts-themes-packages.nix
```diff
@@ -0,0 +1,54 @@
...
+
+    nativeBuildInputs = [ pkgs.makeWrapper ];
+
+    installPhase = ''
+      mkdir -p $out/share/icons
+      cp -r src/breeze-plus* $out/share/icons/
+    '';
+  };
+
+  gabarito-pkg = pkgs.stdenv.mkDerivation {
```
Cu3PO42, line 20: While a reasonable approach, you can get this more easily from nixpkgs' `google-fonts`.

sdata/dist-nix/fonts-themes-packages.nix
```diff
@@ -0,0 +1,54 @@
...
+
+let
+  # Custom package derivations
+  # based on AUR analogs
+  breeze-plus-pkg = pkgs.stdenv.mkDerivation {
+    pname = "breeze-plus";
+    version = "6.2.5";
+    src = breeze-plus;
+
+    nativeBuildInputs = [ pkgs.makeWrapper ];
```
Cu3PO42, line 12: You're not using that.

sdata/dist-nix/fonts-themes-packages.nix
```diff
@@ -0,0 +1,54 @@
+# Fonts and themes packages
+{ pkgs, breeze-plus, matugen, gabarito }:
+
+let
+  # Custom package derivations
+  # based on AUR analogs
+  breeze-plus-pkg = pkgs.stdenv.mkDerivation {
+    pname = "breeze-plus";
+    version = "6.2.5";
+    src = breeze-plus;
```
Cu3PO42, line 10: Since Flake inputs are currently still eagerly fetched, it would be preferable to use `fetchurl` here. This would also make it possible to more easily profit from caching.

sdata/dist-nix/fonts-themes-packages.nix
```diff
@@ -0,0 +1,54 @@
...
+
+    nativeBuildInputs = [ pkgs.makeWrapper ];
+
+    installPhase = ''
+      mkdir -p $out/share/icons
+      cp -r src/breeze-plus* $out/share/icons/
+    '';
```
Cu3PO42, line 17: Meta license information is missing here and other places.

sdata/dist-nix/flake.nix
```diff
@@ -0,0 +1,389 @@
...
+      # Development shell with all packages
+      devShells.${system}.default = pkgs.mkShell {
+        buildInputs = builtins.attrValues allPackages;
+      };
+
+
+      # Home Manager configurations
+      homeConfigurations = {
+        # FIXME: change username
+        "your-username" = home-manager.lib.homeManagerConfiguration {
```
Cu3PO42, line 220: Instead of hardcoding one Home configuration, you should expose a module that others can easily consume and integrate into their configuration.

sdata/dist-nix/flake.nix
```diff
@@ -0,0 +1,389 @@
...
+          inherit pkgs;
+          modules = [
+            ({ config, lib, ... }:
+              {
+                # Basic configuration
+                # FIXME: change username
+                home.username = "your-username";
+                home.homeDirectory = "/home/your-username";
+                programs.home-manager.enable = true;
+                home.stateVersion = "25.05";
```
Cu3PO42, line 230: This must never be upstream as it should remain constant at the value the user created their configuration at.

sdata/dist-nix/flake.nix
```diff
@@ -0,0 +1,389 @@
...
+                  SDL_VIDEODRIVER = "wayland";
+                  XDG_CURRENT_DESKTOP = "Hyprland";
+                  XDG_SESSION_TYPE = "wayland";
+                  XDG_SESSION_DESKTOP = "Hyprland";
+
+                  # XDG base directories
+                  XDG_CONFIG_HOME = "$HOME/.config";
+                  XDG_DATA_HOME = "$HOME/.local/share";
+                  XDG_BIN_HOME = "$HOME/.local/bin";
+                  XDG_CACHE_HOME = "$HOME/.cache";
```
Cu3PO42, line 249: This is unnecessary.

sdata/dist-nix/flake.nix
```diff
@@ -0,0 +1,389 @@
...
+              {
+                # Basic configuration
+                # FIXME: change username
+                home.username = "your-username";
+                home.homeDirectory = "/home/your-username";
+                programs.home-manager.enable = true;
+                home.stateVersion = "25.05";
+
+                # Install all packages + fonts
+                home.packages = builtins.attrValues allPackages;
```
Cu3PO42, line 233: I don't think all packages that are needed internally should be installed globally.

sdata/dist-nix/flake.nix
```diff
@@ -0,0 +1,389 @@
...
+                      ];
+                    };
+
+                    Install = {
+                      WantedBy = [ "graphical-session.target" ];
+                    };
+                  };
+
+                  # Polkit agent (for authentication dialogs)
+                  polkit-gnome-authentication-agent-1 = lib.mkIf (pkgs.stdenv.isLinux) {
```
Cu3PO42, line 340: We can't very well be not on Linux, can we? That said, `systemd.user.services` is simply ignored on non-Linux, so you don't need this `mkIf` even if you wanted to support, say, Darwin.

sdata/dist-nix/flake.nix
```diff
@@ -0,0 +1,389 @@
...
+          # Kitty terminal
+          "kitty/kitty.conf".source = ../../dots/.config/kitty/kitty.conf;
+          "kitty/scroll_mark.py".source = ../../dots/.config/kitty/scroll_mark.py;
+          "kitty/search.py".source = ../../dots/.config/kitty/search.py;
+
+          # KDE Material You Colors
+          "kde-material-you-colors/config.conf".source = ../../dots/.config/kde-material-you-colors/config.conf;
+
+          # Kvantum themes
+          "Kvantum/kvantum.kvconfig".source = ../../dots/.config/Kvantum/kvantum.kvconfig;
```
Cu3PO42, line 101: You can link entire folders. There is no need to do this seperately for every file.

sdata/dist-nix/flake.nix
Cu3PO42: As they are here. They will. It is possible to use `mkOutOfStoreSymlink` to get writable files. This should probably be provided behind an option.

clsty:
About handling the dotfiles, i.e. `dots/`, if we are doing this using Nix then the following requirements must be fulfilled.

**1. Allow modifications over the existing dotfiles.**

Current state of `./setup install`:
After finishing running `./setup install` users can modify any dotfiles in a traditional way, and if they run `./setup install` again to update then they need to skip the steps which overwrite the targets that they have modified and later sync the upgrade manually for such targets by themselves.
(For Hyprland, specially we have a `custom` folder along with `~/.config/hypr/hyprland.conf` which will only get overwritten the first time but not the later times running `./setup install`.)

This works but is not elegant. A WIP solution is using yaml config to store the selected behavior for each target, see #2137.

If we use Nix then it must be at least better than the current state described above, mainly in terms of convenience and automation.

**2. Allow choosing targets.**

This is similar to the above. For example user may want to use their own `~/.config/foot` instead of the files under `dots/.config/foot` entirely.

**3. Easy developing dotfiles or at least not worse than current state.**

About the current state, idk because it's @end-4 who develops the dotfiles.
- If I were the one who develops the dotfiles, I will make changes to the local Git repo `dots-hyprland` and rerun `./setup install-files -f` to apply the changes to observe the outcome.
- @end-4 Could you please share about how you develop the dotfiles currently?

end-4:
I modify my local copy of stuff, copy the relevant parts over, optionally selectively pick changes then commit. It's.... the most obvious way but I guess not necessarily the cleanest

clsty:
I'm closing this PR cuz it does not meet the requirements, and also `dist-nix` has been updated so this one is somehow outdated.

PR #2231, opened 2025-10-18
Scratchpad functionality

maxmdv:
## Describe your changes

Added a scratchpad users can trigger with a shortcut (currently bound to Super+Ctrl+P as in pad).

## Is it ready? Questions/feedback needed?

https://github.com/user-attachments/assets/9fae0745-6f31-485e-bdc7-f9c1107ff02c

Works pretty well on my end!

PR #1858, opened 2025-08-21
More detailed resources usage

xXFiEsTaDeAmOnXx:
## Describe your changes
As requested in [issue #1847](https://github.com/end-4/dots-hyprland/issues/1847), this pull request adds additional CPU information to the `ResourcesPopup.qml` module. This includes the CPU frequency and the CPU temperature. 

A prieview:

image

## Is it ready? Questions/feedback needed?

The CPU info has been fully tested on both my desktop and laptop and works as expected.

We might consider adding GPU-related metrics for a more complete system overview.

PR #2303, opened 2025-10-26
Feature: MangoHud FPS limiter pop up

reakjra:
## Describe your changes

This just adds a pop up that makes possible to set a custom FPS limit through MangoHud option `fps_limit=x;` with the `Ctrl+Super+G` keybind. 

For me it comes handy to set a precise FPS limit without tinkering with the game files or taking time to change MangoHud's settings manually. It can be useful for games that run better with a certain lock of FPS, or while using mods that unlock the engine FPS limit in a game and maybe you get soft-locked in a boss fight, so you just lock back the FPS without needing to quit the game or things like that. Or more, idk, on Deady By Daylight some techs work better with 90FPS rather than 120, so with this you can fastly change the lock without losing much time etc. etc.

Anyway MangoHud it's practically a must for Linux gaming so yeah.
 
 
image

https://github.com/user-attachments/assets/c532dc63-ec5e-46ec-9ddc-15b7b7330e74
> You can see the FPS in the left upper corner; video is low quality due compression

 
 
It's ready.

### DIFF
--- a/dots/.config/hypr/hyprland/keybinds.conf
+++ b/dots/.config/hypr/hyprland/keybinds.conf
@@ -34,6 +34,7 @@ bindd = Super, Slash, Toggle cheatsheet, global, quickshell:cheatsheetToggle # T
 bindd = Super, K, Toggle on-screen keyboard, global, quickshell:oskToggle # Toggle on-screen keyboard
 bindd = Super, M, Toggle media controls, global, quickshell:mediaControlsToggle # Toggle media controls
 bind = Super, G, global, quickshell:crosshairToggle # Toggle crosshair
+bindd = Ctrl+Super, G, Toggle FPS limiter popup, global, quickshell:fpsLimiterToggle # Set FPS limit through MangoHud
 bindd = Ctrl+Alt, Delete, Toggle session menu, global, quickshell:sessionToggle # Toggle session menu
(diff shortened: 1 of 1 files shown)


end-4:
I think this and #2231 should be merged together into some sort of overlay like Game Bar on Windows

reakjra:
yes a Game Bar makes sense, tho, why with the scratchpad? I personally think that's should remain like a shortcut, its fast and handy. Maybe the Game Bar could consider a graph for https://github.com/end-4/dots-hyprland/pull/1858, a toggle for the crosshair, maybe even for MangoHud itself, for the mic, etc. etc. or you're talking about a more "generalist" overlay?

end-4:
The point is just to be an overlay. There's nothing wrong with having both a notepad and gaming features

end-4:
I've created a Game Bar-like overlay for misc widgets like this in 4f68e9e61afb2a75452d26a6c0ec86830e005fee. For instructions on how to make an overlay widget see [StyledOverlayWidget.qml](https://github.com/end-4/dots-hyprland/blob/main/dots/.config/quickshell/ii/modules/overlay/StyledOverlayWidget.qml)

reakjra:
@end-4 closing this for https://github.com/end-4/dots-hyprland/pull/2388.

PR #2307, opened 2025-10-26
Prepend Noto Sans Arabic in fontconfig

aervxa:
## Describe your changes
Arabic font rendering by default falls into the first supporting arabic glyph supporting font, which happens to be an urdu font, which is **not** arabic

since the dotfiles alr include a lot of font configs, why not include a good default for arabic too?

edit:
arabic is a special case that needs this setup, cjk fonts will auto load with the installing of noto-fonts-cjk
so no, i don't think this is a personal change, it's more of like a good default for readability ;)

### DIFF
--- a/dots-extra/fontsets/ar/conf.d
+++ b/dots-extra/fontsets/ar/conf.d
@@ -1,0 +1,19 @@
+<?xml version="1.0"?>
+<!DOCTYPE fontconfig SYSTEM "urn:fontconfig:fonts.dtd">
+<fontconfig>
+  <match target="font">
+    <edit name="rgba" mode="assign">
+      <const>none</const>
+    </edit>
+  </match>
+
+  <!-- Fix for: arabic fonts rendering in Noto Nastaliq Urdu | affects Chromium, Discord (maybe all chromium based apps, but not Spotify somehow) -->
+  <match target="pattern">
+    <test compare="eq" name="family">
+      <string>sans-serif</string>
+    </test>
+    <edit name="family" mode="prepend" binding="strong">
+      <string>Noto Sans Arabic</string>
+    </edit>
+  </match>
+</fontconfig>
--- a/dots/.config/fontconfig/conf.d
+++ b/dots/.config/fontconfig/conf.d
@@ -1,19 +1,9 @@
 <?xml version="1.0"?>
 <!DOCTYPE fontconfig SYSTEM "urn:fontconfig:fonts.dtd">
 <fontconfig>
-  <match target="font">
-    <edit name="rgba" mode="assign">
-      <const>none</const>
-    </edit>
-  </match>
-
-  <!-- Fix for: arabic fonts rendering in Noto Nastaliq Urdu | affects Chromium, Discord (maybe all chromium based apps, but not Spotify somehow) -->
-  <match target="pattern">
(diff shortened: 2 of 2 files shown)


aervxa:
binding=strong affects default font (oops)

aervxa:
Can I just apologize for the mistakes I made 😭
It's all good now, and yes, tested it twice, with every nitpick (i hope)

aervxa:
This **entire pr** is useless to all apps except chromium and discord (would say all chromium apps, but spotify is not affected)
all other apps render arabic in Noto Sans Arabic

### Other solutions I tried:
rejecting the nastaliq font as a whole as recommended in the [wiki](https://wiki.archlinux.org/title/Font_configuration/Examples#Excluding_Arabic_script_from_other_languages)
- caveat 1: someone wouldn't be able to display urdu in that case
- caveat 2: chromium and discord would then render arabic in Rubik, (in chromium, looks terrible. in discord, doesn't connect)

### what did work:
prepending the font, "noto sans arabic", the test checks only if family is of sans-serif
- pro 1: noto-sans-arabic is sans-serif
- pro 2: the prepend won't affect other familes (without the test, the terminal font breaks, so does default font line heights)

apologies for the many commits (maybe this is why they invented squash 🥲) (very unprofessional) (edit: def getting bashed)
this final one is what i was using on my previous arch install before messing around in this new install, it has no caveats, so far, i did use the font config for around 2 months i think and didn't face a single problem

end-4:
On arch wiki something like this is used
```
<test compare="contains" name="lang">
  <string>ar</string>
</test>
```

Here you don't have it, so wouldn't this affect every language?

aervxa:
The reason I did this instead ...
```xml
<test compare="eq" name="family">
  <string>sans-serif</string>
</test>
```
... is because discord doesn't mark the font as arabic or smt nor does chromium, so it doesn't meet that test requirement.
things like firefox works with the lang test

and arabic doesn't have all that many styles, so it just goes as "sans-serif" which is also the default font style
so prepending noto sans arabic on the main font style would prepend it to literally every sans-serif text
why only sans-serif? why not bind to everything?
terminals for whatever reason just immediately go to noto sans, which gets really ugly
terminals only use mono if iirc, so no biggie, whoever uses sans-serif will already have noto sans on the line, so that's fine

end-4:
I'm not a fan of Noto. I'm not accepting this. 
But maybe we can prepend Rubik before Noto fonts? 

Also, does this change hurt Urdu?

clsty:
I've moved it into `dots-extra/fontsets/ar`. To use this fontconfig, user has to run `./setup install --fontset ar`.

The changes on `dots/.config/fontconfig` has been reverted.

Merging.

aervxa:
@end-4 

> Also, does this change hurt Urdu?

This is why i prepended the font and did not reject it like how the arch wiki recommended
so, no

> But maybe we can prepend Rubik before Noto fonts?

Yes we can, It would still need the sans-serif test, cuz it doesn't work without that (doesn't work means falls back to urdu)
- can be achieved by just changing "Noto Sans Arabic" to "Rubik" in `<edit><string>`
edit: in `/dots-extra/fontsets/ar/conf.d` ofc

PR #2309, opened 2025-10-27
Smart-live-rebuild added to Gentoo

jwihardi:
# Added smart-live--rebuild for simplicity.


clsty:
Thanks for your contribution!

PR #2313, opened 2025-10-27
Install scripts readability improvement

majaahh:
## Describe your changes
Corrected indention, improved how cases look and some if's improvements

Needs testing

### DIFF
--- a/sdata/dist-arch/install-deps.sh
+++ b/sdata/dist-arch/install-deps.sh
@@ -39,8 +39,12 @@ fi
 
 # Issue #363
 case $SKIP_SYSUPDATE in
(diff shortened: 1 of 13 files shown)


clsty:
I apprieciate your consideration but I prefer the current state.

Talking about readability, everyone has their own understanding and taste, and so do I. For example, for some `case` syntax, I squash the lines together, so they look like a "paragraph" which differ from other lines, thus in a global view the structure is clearer.

Closing.

clsty:
I've noticed your fix for `outdate_detect` though. Done in 43960b3a605e8ba75191b3c972093a6cd42e171e. Thanks.

PR #2322, opened 2025-10-28
Changes with new submodule

end-4:
## Describe your changes

I made the characters in the password indicator very fancy. It requires a submodule so I did it in a separate branch to not break stuff

@clsty can you ensure that the installation works fine when there's a submodule? Currently `dots/.config/quickshell/ii/modules/common/widgets/shapes` points to [end-4/rounded-polygon-qmljs](https://github.com/end-4/rounded-polygon-qmljs), which is sort of a library for these cool shapes

## Preview :D

image

### DIFF
--- a/sdata/lib/functions.sh
+++ b/sdata/lib/functions.sh
@@ -287,3 +287,25 @@ function check_disk_space() {
   
   return 0
 }
+
+function auto_get_git_submodule(){
+  local git_submodules_list=()
+
+  while IFS= read -r path; do
+    [ -z "$path" ] && continue
+    git_submodules_list+=("$path")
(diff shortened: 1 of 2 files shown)


clsty:
> @clsty can you ensure that the installation works fine when there's a submodule? Currently `dots/.config/quickshell/ii/modules/common/widgets/shapes` points to [end-4/rounded-polygon-qmljs](https://github.com/end-4/rounded-polygon-qmljs), which is sort of a library for these cool shapes

Done in https://github.com/end-4/dots-hyprland/pull/2322/commits/4672138b009f25c1c2d546ca700e0c8f5730f809

end-4:
Thank you!

Issue #2320, opened 2025-10-28
suspend/resume causes high CPU/RAM usage and audio issues with QuickShell


lei-jia-xing:
### Step 1. Before you submit

- [x] I have read the [Troubleshooting](https://ii.clsty.link/en/ii-qs/04troubleshooting/) and [Usage](https://ii.clsty.link/en/ii-qs/02usage/) pages.
- [x] I've successfully updated to the latest version following the [guidance](https://ii.clsty.link/en/ii-qs/01setup/#updating).
- [x] I've successfully updated the system packages to the latest.
- [ ] I've ticked the checkboxes without reading their contents

### Step 2. Quick diagnose info

<details><summary>Quick diagnose</summary>

```
/========================\
| Checking git repo info |
\========================/

[===diagnose===] git remote get-url origin
https://github.com/end-4/dots-hyprland.git
[---SUCCESS---]

[===diagnose===] git rev-parse HEAD
44da20ed4aaca2b63a3dc6c6378f912f91b04bb3
[---SUCCESS---]

/=================\
| Checking distro |
\=================/

[===diagnose===] ii_check_distro_id
distro ID: arch
distro ID_LIKE: 
[---SUCCESS---]

[===diagnose===] ii_check_distro
LSB Version:	n/a
Distributor ID:	Arch
Description:	Arch Linux
Release:	rolling
Codename:	n/a
[---SUCCESS---]

/====================\
| Checking variables |
\====================/

[===diagnose===] declare -p XDG_CACHE_HOME
./diagnose: line 36: declare: XDG_CACHE_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p XDG_CONFIG_HOME
./diagnose: line 36: declare: XDG_CONFIG_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p XDG_DATA_HOME
./diagnose: line 36: declare: XDG_DATA_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p XDG_STATE_HOME
./diagnose: line 36: declare: XDG_STATE_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p ILLOGICAL_IMPULSE_VIRTUAL_ENV
declare -x ILLOGICAL_IMPULSE_VIRTUAL_ENV="~/.local/state/quickshell/.venv"
[---SUCCESS---]

/============================\
| Checking directories/files |
\============================/

[...]

PR #2323, opened 2025-10-29
fix: prevent volumeChanged recursion (#2320)

lei-jia-xing:
## Describe your changes

Fixed memory leak in Audio.qml by skipping volume writes during PipeWire resume, avoiding the ping-pong effect with PipeWire.

Yes, the change is small and ready for review.

### DIFF
--- a/dots/.config/quickshell/ii/services/Audio.qml
+++ b/dots/.config/quickshell/ii/services/Audio.qml
@@ -29,12 +29,18 @@ Singleton {
         property real lastVolume: 0
         function onVolumeChanged() {
             if (!Config.options.audio.protection.enable) return;
+            const newVolume = sink.audio.volume;
+            // when resuming from suspend, we should not write volume to avoid pipewire volume reset issues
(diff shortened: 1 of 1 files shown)


end-4:
> I see the problem

Wait no, why is the recursion a problem
It only changes the volume once and recurses at most once

end-4:
I think this is the real problem 🤔

```js
if (sink.ready && (isNaN(sink.audio.volume) || sink.audio.volume === undefined || sink.audio.volume === null)) {
    sink.audio.volume = 0;
}
```
Can you see if removing this fixes infinite recursion + leak for you? I've had the leak once or twice but it's not consistently reproducible

lei-jia-xing:
You're right — we shouldn’t modify volume during resume, as doing so triggers PipeWire’s protection logic.I will modify my pr

lei-jia-xing:
Everything seems fine now; it works perfectly on my system.

end-4:
thanks

PR #2324, opened 2025-10-29
fix(qml): avoid ColumnLayout polish() loop in NotificationItem.qml

JxJxxJxJ:


dots-hyprland
clsty, line 1: What's this thing?

end-4:
When does the polish() loop thing happen again?

end-4:
This also breaks group expanding animations

PR #2329, opened 2025-10-29
Add YAML-based file installation system (#2137)

mattvanharn:
- Replace hardcoded Bash logic with declarative YAML configuration
- Implement user preference wizard for shell/terminal/keybindings
- Add conditional file copying based on user preferences
- Support multiple sync modes: sync, soft, hard, hard-backup, soft-backup, skip, skip-if-exists
- Implement MD5 hash comparison for idempotent backups
- Add fontconfig fontset support via II_FONTSET_NAME
- Complete coverage of all config directories and files from original script

This is an experimental feature enabled via --exp-files flag.


clsty:
LGTM. I deeply appreciate your high-quality work.

Merging. (Related: #2137)

PR #2339, opened 2025-10-30
Add visibility/collapse settings for resources to settings app

### LINKED ISSUE #2282: Toggle to enable/disable resource usage display

A toggle to enable/disable resource usage display on the bar. 

### How will it help?

I think this will help perfomance a little bit and make the bar cleaner. 

### Extra info

Some other improvements:
- Option to select resources to show as per the user's preference. eg: disabling the swap usage info. 
- Show network usage stats

vaguesyntax:
## Describe your changes

Closes #2282

- Added **resource options** to the **“Bar”** settings, which include the following:
image

- Added `ConfigMultiSelectionArray` for the `collapsedResources`, which does exactly what the name says : Allows multiple selections.

_This setting is especially useful for users with smaller screens_

Yes.

### DIFF
--- a/dots/.config/quickshell/ii/modules/bar/Resource.qml
+++ b/dots/.config/quickshell/ii/modules/bar/Resource.qml
@@ -10,7 +10,7 @@ Item {
     property int warningThreshold: 100
     property bool shown: true
     clip: true
-    visible: width > 0 && height > 0
+    visible: implicitWidth > 0 && implicitHeight > 0
     implicitWidth: resourceRowLayout.x < 0 ? 0 : resourceRowLayout.implicitWidth
     implicitHeight: Appearance.sizes.barHeight
     property bool warning: percentage * 100 >= warningThreshold
--- a/dots/.config/quickshell/ii/modules/bar/Resources.qml
+++ b/dots/.config/quickshell/ii/modules/bar/Resources.qml
@@ -7,7 +7,7 @@ MouseArea {
     id: root
     property bool borderless: Config.options.bar.borderless
     property bool alwaysShowAllResources: false
-    implicitWidth: rowLayout.visible ? rowLayout.implicitWidth + rowLayout.anchors.leftMargin + rowLayout.anchors.rightMargin : 0
+    implicitWidth: rowLayout.visible ? rowLayout.implicitWidth  : 0
     implicitHeight: Appearance.sizes.barHeight
     hoverEnabled: true
 
@@ -22,8 +22,8 @@ MouseArea {
     
         spacing: 0
         anchors.fill: parent
-        anchors.leftMargin: 4
-        anchors.rightMargin: 4
+        //anchors.leftMargin: 4
+        //anchors.rightMargin: 4
 
         Resource {
             id: memoryResource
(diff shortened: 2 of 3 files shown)


dots/.config/quickshell/ii/modules/common/Config.qml
```diff
@@ -205,8 +205,11 @@ Singleton {
                 property bool verbose: true
                 property bool vertical: false
                 property JsonObject resources: JsonObject {
-                    property bool alwaysShowSwap: true
-                    property bool alwaysShowCpu: true
+                    property bool showCpu: true 
+                    property bool showSwap: true
+                    property bool showMemory: true
+                    property string style: "icon" // Options: "icon", "detailed"
+                    property list<string> collapseResources: ["swap"]
```
end-4: No reason to have `show-` options single bools but collapsible stuff a list
Make it something like the quick toggles: a list of objects like this
```
{"type": "cpu", "collapse": true}
```

dots/.config/quickshell/ii/modules/common/widgets/ConfigMultiSelectionArray.qml
end-4, line 1: I'm somewhat fine with this as a programmer, but disappointed as a designer.
An array of closely grouped buttons is not for individually toggleable options. Consult the M3 specs.

end-4 (changes requested):

dots/.config/quickshell/ii/modules/common/widgets/ConfigMultiSelectionArray.qml
end-4: Now I'm disappointed at myself
image

dots/.config/quickshell/ii/modules/common/Config.qml
vaguesyntax: I've used a list of objects and changed the settings: 

image

Blade04208:
hope thoust dont mind me butting in here, but would this not be better UX for this ui? Mixing multisel's and single-sel's is kind of a bad idea if the multisel's are only used twice

Frame 34588072(1)

id implement this myself but i dont know quickshell or qt and can only design so

vaguesyntax:
> hope thoust dont mind me butting in here, but would this not be better UX for this ui? Mixing multisel's and single-sel's is kind of a bad idea if the multisel's are only used twice

Actually, that's not a bad idea, but I thought about creating a multi-selection button group so it could also be used in different sections of the settings. If end thinks the same way as you (removing/not using the multi-selection button group), I can refactor the settings layout like yours.

Blade04208:
> I thought about creating a multi-selection button group so it could also be used in different sections of the settings.

But where could it be used that a toggle or single-sel couldn't? the example in the [m3 docs](https://m3.material.io/components/button-groups/guidelines#2697ae16-8e80-4409-ba3f-ec3144564ab3) makes sense as its a filter in a searchsheet and also on the small screen of a phone, but i don't really see where it could be used on a desktop settings menu like:
- ~~filters for search contents~~ chips or a dropdown using radio/checkboxes/togglebuttons is better for ux
-  image -- nope, this is better organized imo
- overlay maybe? sounds like a good enough use for it

If end or yourself wants to keep it in, by all means, but I dont see any practical use for it lol

PR #2344, opened 2025-10-31
Improve music recognition script source management

Yunexiz:
Refactor source type handling and validation checks.

## Describe your changes

The song recognition kept getting the wrong source. This change makes it get the default sink and also changes some of the validation, checking if all dependencies are installed and if pulseaudio recognizes the source.

Yes. Give feedback as you will.


end-4:
uhh thanks 
i forgot to press merge

PR #2346, opened 2025-10-31
feat(polkit): fingerprint support

0blivi0nis:
## Describe your changes
1. No fingerprint
image

2. If user setup requires both password and fingerprint
    - First the password prompt is shown, after entering the password the message to use fingerprint shows up

| Password | Fingerprint |
|--------------|-----------|
| Password Image | Fingerprint Image |

3. If user has setup to use fingerprint or password
    - First the fingerprint prompt shows, when that fails the password prompt shows
 
| Fingerprint | Fingerprint Fails |
|--------------|-----------|
| Fingerprint Image | Password Image |

4. Waffles

| Fingerprint | Fingerprint Fails |
|--------------|-----------|
| Fingerprint Image | Password Image |

1. It would be great to have a Ctrl + C keybind to cancel the fingerprint prompt and switch to the password prompt.
    - I tried doing this by adding the keybind and setting `PolkitService.isFingerprintCurrentlyOffered = false;`, but that didn’t work.

Other than the keybind to cancel the fingerprint for scenario 3, it is ready ✅

### DIFF
--- a/dots/.config/quickshell/ii/modules/waffle/polkit/WPolkitContent.qml
+++ b/dots/.config/quickshell/ii/modules/waffle/polkit/WPolkitContent.qml
@@ -104,6 +104,12 @@ Rectangle {
                         horizontalAlignment: Text.AlignLeft
                         text: PolkitService.cleanMessage
                     }
+                    WText {
+                        Layout.fillWidth: true
+                        wrapMode: Text.Wrap
+                        horizontalAlignment: Text.AlignHCenter
+                        text: PolkitService.flow.supplementaryMessage
+                    }
                     WAppIcon {
                         iconName: "fingerprint"
                         isMask: PolkitService.flow?.iconName === ""


Cu3PO42:
Hi! Unfortunately I don't have the time to contribute any code right now, but I can contribute some ideas:

- It is, in general, not possible to parse PAM configurations reliably, since some distributions extend the base language in incompatible ways. (This is also why getting PAM to work in Nix-installed applications on non-NixOS is kind of tricky.)
- We therefore have to rely on just the messages sent during authentication.
- The standard fingerprint module sets the supplementaryMessage to indicate availability of fingerprint auth.
- Instead of the custom `interactionAvailable`, you probably should rely on `flow.isResponseRequired`.

0blivi0nis:
I tried hiding the password prompt for scenario 2 after entering it but it doesnt work, and I think this looks better anyways.

The custom `interactionAvailable` is done by @end-4 so I didnt touch any of that

PR #2349, opened 2025-10-31
simplify install-yay

risjsnx:
simplify the install-yay to just run [sudo pacman -S --noconfirm yay] the no confim option is because yay package in pacman in really small like ~10MiB

### DIFF
--- a/sdata/lib/functions.sh
+++ b/sdata/lib/functions.sh
@@ -292,8 +292,9 @@ function check_disk_space() {
   return 0
 }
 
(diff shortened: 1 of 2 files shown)


clsty:
I guess you're using chaotic-AUR or Manjaro... Anyway it's not applicable for vanilla Arch Linux and some other Arch(-based) distros. And also if you use chaotic-AUR/Manjaro you're more likely to meet problems when using illogical-impulse.

risjsnx:
@clsty 好—好

PR #2354, opened 2025-11-01
Save paths for screen recording and screenshots

### LINKED ISSUE #2305: Add Option to Change Screen Record Save Location

I like to keep my screen recordings in a different directory instead of `~/Videos`

Can you add an option to define the path where recordings are saved?

### How will it help?

It gives users more control instead of hardcoding the save path

### Extra info
_No response_

vaguesyntax:
## Describe your changes

Closes #2305
- Added config options to the **Services** tab in the settings app for screen recording and screenshot save paths.

image

Yep, settings layout _and section_ can be changed though.

### DIFF
--- a/dots/.config/quickshell/ii/modules/regionSelector/RegionSelection.qml
+++ b/dots/.config/quickshell/ii/modules/regionSelector/RegionSelection.qml
@@ -33,9 +33,9 @@ PanelWindow {
     property var selectionMode: RegionSelection.SelectionMode.RectCorners
     signal dismiss()
     
-    property string permanentScreenshotDir: Config.options.screenSnip.savePath && Config.options.screenSnip.savePath !== ""
-        ? Config.options.screenSnip.savePath
-        : ""
+    property string saveScreenshotDir: Config.options.screenSnip.savePath !== ""
+                                       ? Config.options.screenSnip.savePath
+                                       : ""
 
     property string screenshotDir: Directories.screenshotTemp
     property string imageSearchEngineBaseUrl: Config.options.search.imageSearch.imageSearchEngineBaseUrl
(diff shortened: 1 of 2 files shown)


dots/.config/quickshell/ii/modules/settings/ServicesConfig.qml
```diff
@@ -85,6 +85,31 @@ ContentPage {
...
+            text: Config.options.screenRecord.savePath
+            wrapMode: TextEdit.Wrap
+            onTextChanged: {
+                Config.options.screenRecord.savePath = text;
+            }
+        }
+        
+        MaterialTextArea {
+            Layout.fillWidth: true
+            placeholderText: Translation.tr("Screenshot Path (leave empyt to just copy)")
```
madjidtaha:
```suggestion
            placeholderText: Translation.tr("Screenshot Path (leave empty to just copy)")
```

hexqubit:
Thanks

dots/.config/quickshell/ii/modules/regionSelector/RegionSelection.qml
```diff
@@ -33,6 +33,10 @@ PanelWindow {
     property var selectionMode: RegionSelection.SelectionMode.RectCorners
     signal dismiss()
     
+    property string permanentScreenshotDir: Config.options.screenSnip.savePath && Config.options.screenSnip.savePath !== ""
```
end-4: i don't think it should be called "permanent", sounds dangerous, just call it `screenshotDir`

dots/.config/quickshell/ii/services/DateTime.qml
```diff
@@ -22,6 +22,7 @@ Singleton {
     property string shortDate: Qt.locale().toString(clock.date, Config.options?.time.shortDateFormat ?? "dd/MM")
     property string date: Qt.locale().toString(clock.date, Config.options?.time.dateFormat ?? "dddd, dd/MM")
     property string collapsedCalendarFormat: Qt.locale().toString(clock.date, "dd MMMM yyyy")
+    property string fileDateTime: Qt.locale().toString(clock.date, "dd-MM-yyyy_HH.mm")
```
end-4: Since we're using the shell (bash) anyway, just use `date` command like in record.sh instead of introducing this ugly format here

end-4 (changes requested):

dots/.config/quickshell/ii/modules/regionSelector/RegionSelection.qml
vaguesyntax: Uhh, there was already a variable named screenshotDir so I called it `saveScreenshotDir`

end-4 (approved):
Image

PR #2355, opened 2025-11-01
Add battery health to battery popup

EinBowser:
## Describe your changes

Adding an optional battery health display to the battery popup. 
It is running a helper script to get the designed/current capacity of the first battery (BAT0 should be default for most devices) and calculates the health.

### Why a separate helper function?
The currently used UPower.displayDevice doesn't show all the information a "real" device would. The fetching of the right battery trough UPower was unpractical from my experience and a simple helper function did the trick.
https://quickshell.org/docs/v0.1.0/types/Quickshell.Services.UPower/UPowerDevice/ 

image

It should be ready.

### DIFF
--- a/dots/.config/quickshell/ii/modules/bar/BatteryPopup.qml
+++ b/dots/.config/quickshell/ii/modules/bar/BatteryPopup.qml
@@ -90,7 +90,7 @@ StyledPopup {
                 Layout.fillWidth: true
 
                 MaterialSymbol {
-                    text: "healing"
+                    text: "heart_check"
                     color: Appearance.colors.colOnSurfaceVariant
                     iconSize: Appearance.font.pixelSize.large
                 }


dots/.config/quickshell/ii/scripts/battery/calculate-health.sh
end-4, line 1: This is unreliable and unnecessary

https://quickshell.org/docs/v0.1.0/types/Quickshell.Services.UPower/UPowerDevice/#healthPercentage

dots/.config/quickshell/ii/modules/common/Config.qml
```diff
@@ -257,6 +257,7 @@ Singleton {
                 property int full: 101
                 property bool automaticSuspend: true
                 property int suspend: 3
+                property bool showHealth: false
```
end-4: With Quickshell's builtin way to get health we can make it reliable, so no point in providing configuration for such a minor change.

dots/.config/quickshell/ii/scripts/battery/calculate-health.sh
EinBowser: UPower does not provide the health in its display device. See following:

`    Component.onCompleted: {
        console.log("UPower.display device attributes")
        for (let key in UPower.displayDevice) {
            try {
                console.log(`${key}: ${UPower.displayDevice[key]}`)
            } catch (e) {
                console.log(`${key}: <unreadable> (${e})`)
            }
        }
    }`

returns:

` DEBUG qml: UPower.display device attributes
 DEBUG qml: objectName: 
 DEBUG qml: type: 2
 DEBUG qml: powerSupply: true
 DEBUG qml: energy: 5.776
 DEBUG qml: energyCapacity: 29.5336
 DEBUG qml: changeRate: 22.7088
 DEBUG qml: timeToEmpty: 915
 DEBUG qml: timeToFull: 0
 DEBUG qml: percentage: 0.2
 DEBUG qml: isPresent: true
 DEBUG qml: state: 2
 DEBUG qml: healthPercentage: 0
 DEBUG qml: healthSupported: false
 DEBUG qml: iconName: battery-caution-symbolic
 DEBUG qml: isLaptopBattery: true
 DEBUG qml: nativePath: 
 DEBUG qml: model: 
 DEBUG qml: ready: true`

I will commit a workaround that does not include a helper script soon.

dots/.config/quickshell/ii/modules/common/Config.qml
EinBowser: As said before, it isn't 100%  reliable. Do you still want me to remove the config option or just the visibility in the settings?

dots/.config/quickshell/ii/scripts/battery/calculate-health.sh
end-4: > healthSupported: false

We should make it show only if it's supported

I still prefer this over having an extra bash script even if it's "less reliable"

EinBowser: Made a commit to fix it.
I tried to keep it minimal, if you want I could also make a new variable to check it.
Do you want to keep the config, remove it in setting or completely?

end-4: The config option should be removed and showing health or not should only depend on whether it is supported

EinBowser: Removed them in newest commit

PR #2361, opened 2025-11-03
Try kebab over underscores for app icons

An-dz:
## Describe your changes

This tries using kebab in place of underscores.

It works, I've been using for many weeks, but I don't remember which apps had this problem. It might have fixed other stuff later too.

### DIFF
--- a/dots/.config/quickshell/ii/services/AppSearch.qml
+++ b/dots/.config/quickshell/ii/services/AppSearch.qml
@@ -86,6 +86,10 @@ Singleton {
         return str.toLowerCase().replace(/\s+/g, "-");
     }
 
+    function getUndescoreToKebabAppName(str) {
+        return str.toLowerCase().replace(/_/g, "-");
+    }
+
     function guessIcon(str) {
         if (!str || str.length == 0) return "image-missing";
 
@@ -124,6 +128,8 @@ Singleton {
         const kebabNormalizedGuess = getKebabNormalizedAppName(str);
         if (iconExists(kebabNormalizedGuess)) return kebabNormalizedGuess;
 
+        const undescoreToKebabGuess = getUndescoreToKebabAppName(str);
+        if (iconExists(undescoreToKebabGuess)) return undescoreToKebabGuess;
 
         // Search in desktop entries
         const iconSearchResults = Fuzzy.go(str, preppedIcons, {


dots/.config/quickshell/ii/services/AppSearch.qml
```diff
@@ -86,6 +86,10 @@ Singleton {
         return str.toLowerCase().replace(/\s+/g, "-");
     }
 
+    function getUndescoreToKebabAppName(str) {
+        return str.toLowerCase().replace(/_/g, "-");
+    }
+
```
end-4, line 92: Legit, but why don't we combine underscore to kebab with the normalized name? is there any icon name that have both hyphen and underscore?

An-dz: Just did a search and there are very few here: (x86_64 looks like it could be a common pattern, but I only found two for qemu)

bottles_wine-winecfg
lutris_0-ad (lots with lutris_ prefix)
org.virt_manager.virt-manager
appimagekit-Black_Chocobo
appimagekit-de.rwth_aachen.ient.YUView
appimagekit-Ripcord_Icon
Ardour-Ardour_6.0.0
Ardour-Ardour_8.0.0
qemu-system-x86_64
redhat-server_settings
wesnoth-1.14_editor-icon
yast-high_availability

end-4: uhhh
i was asking the wrong thing
i mean having both spaces and underscores
how did you search it?

An-dz: Ah ok, yeah, there are none with any spaces.

But that does not invalidate the inverse, that can't be searched. That is, how many apps have class names with spaces and underscores?

Because the class name could be like `great app_launcher`, and it could have an icon name `great-app_launcher`, but no `great-app-launcher`, and joining the regexps would result on the later string and not the earlier. And as I showed in the search above, there are icons with underscores and hyphens.

And how I found them, I made this `find` call with regular expression:
```bash
find /usr/share/icons -type f -regextype awk -regex '.*/apps/(.*/)?.*(-.*_|_.*-).*'
```
Path must contain /apps/ and the name must contain a hyphen and a underscore later, or an underscore and a hyphen later. It's not a perfect regexp because I'm not excluding `/` later, but for this icons directory it works.

end-4: well...alright
it's not like it's going to break anything, i was just wondering

PR #2362, opened 2025-11-03
Limit booru calls for tags to 10

An-dz:
## Describe your changes

This limits the amount of tags returned by Booru APIs to 10 tags so they are faster and because the UI already limits to 10 anyway.

It's all good, I checked the APIs docs.


end-4:
thanks

PR #2363, opened 2025-11-03
Improve visibility logic for Settings

0blivi0nis:
### ✨ Summary

Improves the visibility logic for configuration options in the **settings** module to ensure that settings rows and subsections are displayed only when relevant.
* Fixed incorrect use of the enabled property instead of visible in the audio protection settings.
  * Previously, the row appeared but was unresponsive.
  * Now, it correctly appears or hides based on the visible state.
* Updated background clock and battery sound settings to be conditionally visible.
* Sub-settings now only appear when their parent feature is enabled.
* Improved overall clarity and user experience by reducing visual clutter and showing only applicable options.
### 🎯 Why

With the amount of options expanding with each update, these adjustments make the settings interface more intuitive and consistent:
* Prevents inactive or irrelevant options from showing.
* Ensures related settings behave predictably and respond to parent toggles.
* Enhances usability and visual coherence in the settings screen.

https://github.com/user-attachments/assets/bab45e6d-ccf5-419a-b1fe-8561730660c9

✅ Testing

* Verified that all affected settings (audio protection, background clock, battery sounds) appear and hide as expected.

### DIFF
--- a/dots/.config/quickshell/ii/modules/settings/InterfaceConfig.qml
+++ b/dots/.config/quickshell/ii/modules/settings/InterfaceConfig.qml
@@ -971,6 +971,7 @@ ContentPage {
             buttonIcon: "center_focus_strong"
             text: Translation.tr("Center icons")
             checked: Config.options.overview.centerIcons
+            visible: Config.options.overview.enable
             onCheckedChanged: {
                 Config.options.overview.centerIcons = checked;
             }


vaguesyntax:
Any chance we could add an animation when the options appear? That way they don’t pop suddenly and hurt the eyes. _Just a tought, not a necessity_

0blivi0nis:
> Any chance we could add an animation when the options appear? That way they don’t pop suddenly and hurt the eyes. _Just a tought, not a necessity_

So i tried adding animation with the code below but it doesn't seem to make much of a difference. I have no clue what I am doing with animations tho so i could just be doing it wrong. Would be nice to have them slide down and up

```
Behavior on visible {
    PropertyAnimation {
        properties: "visible"
        duration: Appearance.animation.elementMoveEnter.duration
        easing.type: Appearance.animation.elementMoveEnter.type
        easing.bezierCurve: Appearance.animation.elementMoveEnter.bezierCurv
    }
}
```

PR #2368, opened 2025-11-03
Fix incorrect password placement when fingerprints

0blivi0nis:
## Describe your changes
The incorrect password animation messes up the placement if fingerprints are configured, this fixes that problem.

https://github.com/user-attachments/assets/924eed72-fd71-4ebd-9750-abd48489f91c

Yes it is.


end-4:
This is the wrong way, don't hard code that 42

end-4:
^ that's the correct way

PR #2371, opened 2025-11-04
Gentoo Updates

jwihardi:
- SongRec and Roboto-Flex ebuilds added.
- Quickshell, tools, and hyprland dependencies updated
- Hyprland now no longer uses live ebuilds but the testing (0.51.1 was added). I think this is better, even though Gentoo might take a bit longer to update than Arch Hyprland-wise.
- Instead of having cluttered ebuilds, they were split into smaller ebuilds in local-pkgs. This makes it easier to update.
-  import-ebuilds.sh was created to import and digest ebuilds in local-pkgs.
- Changed the overlay from _localrepo_ to _ii-dots_. I didn't name to illogical-impulse because then portage would print really long lines, but you can rename it to whatever.
- QT has its own keywords and unmasks now.


clsty:
> Do you think this is a good idea? I can make another folder called like Gentoo-AUR.

As long as you think the structure benefits future maintenance, it's a good idea. However `Gentoo-AUR` is confusing. AUR is Arch User Repository so it cannot be something for Gentoo. If you mean "local ebuilds that build non-meta but actual custom packages", I suggest to use name like `local-pkgs`.

jwihardi:
Yeah, I know what AUR is, I was just suggesting the name since it would be an equivalent of the AUR packages needed but for Gentoo. I will name it local-pkgs though 😸

clsty:
The relation between AUR and PKGBUILD is similar to the what of GitHub and Git.

AUR is one of the places holding online PKGBUILDs. However the PKGBUILDs do not necessarily need to be on AUR to be built, in our case some actual packages are also built using local PKGBUILD.

sdata/dist-gentoo/install-setups.sh
clsty, line 1: About this file, it's been moved into sdata/subcmd-install/2.setups.sh

jwihardi:
I completely forgot this PR was still open

sdata/dist-gentoo/install-setups.sh
jwihardi: 😢

jwihardi:
It should be good to go. Before moving setup, I run install on a clean Gentoo VM and it went smoothly.

clsty (approved):
LGTM

clsty:
Merging. Thanks for the updates!

Issue #2385, opened 2025-11-06
Design problems with colors


shadowe1ite:
### Step 1. Before you submit

- [x] I have read the [Troubleshooting](https://ii.clsty.link/en/ii-qs/04troubleshooting/) and [Usage](https://ii.clsty.link/en/ii-qs/02usage/) pages.
- [x] I've successfully updated to the latest version following the [guidance](https://ii.clsty.link/en/ii-qs/01setup/#updating).
- [x] I've successfully updated the system packages to the latest.
- [ ] I've ticked the checkboxes without reading their contents

### Step 2. Quick diagnose info

<details><summary>Quick diagnose</summary>

```
<!-- Run `./diagnose` inside the repo, and paste the result here! -->
```

</details>

### Step 3. Describe the issue

After updating my system, Matugen was upgraded to version 3.0.0, and its compatibility with Illogical Impulse is completely broken.

* Dark/Light mode doesn’t work.

* Color palette selection is broken.

* The generated colors look awful, especially the clock color—it’s really ugly.

<img width="425" height="290" alt="Image" src="https://github.com/user-attachments/assets/648c1976-9fee-4c50-9126-17a4fe0ed3a7" />

### Reminder

- [x] I agree that it's usually impossible for others to help me without my logs.

shadowe1ite:
The AI sidebar colors also broken

<img width="456" height="804" alt="Image" src="https://github.com/user-attachments/assets/91040cf8-9921-4d3d-b70e-026105e34ca6" />

clsty:
Can you try https://github.com/end-4/dots-hyprland/pull/2386 to see if this solves the problem?

shadowe1ite:
> Can you try [#2386](https://github.com/end-4/dots-hyprland/pull/2386) to see if this solves the problem?

Of course, downgrading Matugen will solve the issue — version 2.x.x has no problems. But I don’t use the installation script, so I can’t confirm if it works.

I just reported this since Matugen released the latest version a few weeks ago, so it needs to be fixed anyway.

end-4:
Matugen 3.0 does **not** break color generation. 

> Dark/Light mode doesn’t work.

Where does it not work?

> Color palette selection is broken.
> The generated colors look awful, especially the clock color—it’s really ugly.

The clock's changed color is intentional (matches that of the clock widget on a Pixel phone when Tonal Spot is used). Consider using Tonal Spot, Neutral or Auto (chooses between the previous two)

[...]

PR #2386, opened 2025-11-06
Use pinned version 2.4.1 for matugen (#2385)

clsty:
## Describe your changes

Use pinned version 2.4.1 for matugen

Ready but not tested since I don't have Arch distro right now.

Also if @end-4 decides to always keep up with the latest matugen then this PR may be not needed.


end-4:
matugen didn't break anything, the update just coincides with many redesigns of the shell

PR #2388, opened 2025-11-06
Feature (Overlay): MangoHud Fps Limiter widget

reakjra:
## Describe your changes

MangoHud FPS Limiter popup integrated in the new Overlay. Does the same things as https://github.com/end-4/dots-hyprland/pull/2303. I've added a focus on the overlay since it would not be usable at all in games. 

https://github.com/user-attachments/assets/e2bb9c2f-a68f-4171-930f-fe8f83d924be

It's ready. You may want to change the icon with a better one if exists and/or remove the focus. Tho, as it is shown at the end of the video, if you open the overlay within a game that uses the mouse for moving the camera, the cursor gets completely stuck (same thing happens with the overview). Is it possible to fix this? 'Cause it makes the overlay quite atrocious to use properly, since you need to switch the workspace to unlock the mouse cursor. 

EDIT:
a temporary fix I found for the cursor getting stuck is using this command in the HyprlandFocusGrab inside `Overlay.qml`: 
```bash
 command: ["bash", "-c", "ws=$(hyprctl activeworkspace -j | jq -r '.id'); hyprctl --batch \"keyword animations:enabled 0 ; dispatch workspace empty ; dispatch workspace $ws ; keyword animations:enabled 1\""]
 ```
 
It does work, but of course it causes the small flick of the workspaces being swapped, and its annoying to be fair, especially if you open the overlay when not in a game. 

Ok as far as I saw this seems a common problem within hyprland. Sucks.

### DIFF
--- a/dots/.config/quickshell/ii/modules/overlay/Overlay.qml
+++ b/dots/.config/quickshell/ii/modules/overlay/Overlay.qml
@@ -55,7 +55,7 @@ Scope {
             Connections {
                 target: GlobalStates
                 function onOverlayOpenChanged() {
-                    delayedGrabTimer.start();
+                    delayedGrabTimer.restart();
                 }
             }
 

### CODE REVIEW

File: dots/.config/quickshell/ii/modules/overlay/fpsLimiter/FpsLimiterContent.qml [hunk 1]
```diff
@@ -0,0 +1,92 @@
...
+
+        RowLayout {
+            Layout.fillWidth: true
+            spacing: 10
+
+            ToolbarTextField {
+                id: fpsField
+                Layout.fillWidth: true
+                Layout.preferredWidth: 200
+                placeholderText: qsTr("Set FPS limit (e.g. 80)")
```
end-4: Use `Translation.tr`. It's the translation system the dotfiles uses. `qsTr` doesn't work because Quickshell doesn't support it

dots/.config/quickshell/ii/modules/overlay/fpsLimiter/FpsLimiterContent.qml
```diff
@@ -0,0 +1,92 @@
...
+                inputMethodHints: Qt.ImhDigitsOnly
+                focus: true
+
+                Keys.onReturnPressed: {
+                    root.applyLimit();
+                    event.accepted = true;
+                }
```
end-4: use `onAccepted`

dots/.config/quickshell/ii/modules/overlay/OverlayContext.qml
```diff
@@ -7,6 +7,7 @@ Singleton {
     
     readonly property list<var> availableWidgets: [
         { identifier: "crosshair", materialSymbol: "point_scan" },
+        { identifier: "fpsLimiter", materialSymbol: "speed" },
```
end-4: I think `animation` would be a better icon here, since a speedometer icon can hint at performance in general and not just fps

dots/.config/quickshell/ii/modules/overlay/fpsLimiter/FpsLimiter.qml
```diff
@@ -0,0 +1,10 @@
+import QtQuick
+import Quickshell
+import qs.modules.common
+import qs.modules.overlay
+
+StyledOverlayWidget {
+    id: root
+    title: "FPS Limiter"
```
end-4: Make it "MangoHud FPS Limiter" to be clear

end-4:
Don't you think a little feedback would be nice? Maybe make the submit button next to the textbox use a `save` icon and make it briefly turn into `check` (1s timeout and it turns back to `save`) when a limit is set

dots/.config/quickshell/ii/modules/overlay/fpsLimiter/FpsLimiter.qml
end-4: or maybe just "MangoHud FPS" if that's too long

dots/.config/quickshell/ii/modules/overlay/OverlayContext.qml
reakjra: `area_chart` and `data_thresholding` looks good too if you want to check

reakjra:
Is this fine?

reakjra:
ops my bad I added my personal changes, fixing asap

dots/.config/quickshell/ii/modules/overlay/fpsLimiter/FpsLimiterContent.qml
end-4: No, replace `Keys.onReturnPressed` with `onAccepted`

dots/.config/quickshell/ii/modules/overlay/fpsLimiter/FpsLimiterContent.qml
```diff
@@ -58,13 +83,13 @@ Rectangle {
                 id: fpsField
                 Layout.fillWidth: true
                 Layout.preferredWidth: 200
-                placeholderText: qsTr("Set FPS limit (e.g. 80)")
+                placeholderText: root.showError ? Translation.tr("Insert a valid number!") : Translation.tr("Set FPS limit (e.g. 80)")
```
end-4: Don't use an exclamation mark here. Make it "Please enter a valid number"

dots/.config/quickshell/ii/modules/overlay/fpsLimiter/FpsLimiterContent.qml
```diff
@@ -34,14 +56,17 @@ Rectangle {
...
+
         // Clear the field after applying
         fpsField.text = "";
     }
 
     Keys.onPressed: event => {
         if (event.key === Qt.Key_Escape) {
             fpsField.text = "";
-            event.accepted = true;
+            event.onAccepted();
```
end-4: why does escape key...trigger an accept?

dots/.config/quickshell/ii/modules/overlay/fpsLimiter/FpsLimiterContent.qml
```diff
@@ -10,12 +11,33 @@ Rectangle {
...
+    }
+
+    Timer {
+        id: errorResetTimer
+        interval: 1000
+        onTriggered: {
+            root.showError = false;
+        }
+    }
+
```
end-4: Instead of two separate bools with two different timers, make the icon have three states and use only one timer to reset the state

dots/.config/quickshell/ii/modules/overlay/fpsLimiter/FpsLimiterContent.qml
```diff
@@ -80,7 +105,22 @@ Rectangle {
                     anchors.centerIn: parent
                     horizontalAlignment: Text.AlignHCenter
                     font.pixelSize: Appearance.font.pixelSize.title
-                    text: "keyboard_return"
+                    text: root.showError ? "close" : (root.showCheckIcon ? "check" : "save")
+                    rotation: (root.showCheckIcon || root.showError) ? 360 : 0
```
end-4: 360 is too much
For this case just use 90 or at most 180

dots/.config/quickshell/ii/modules/overlay/fpsLimiter/FpsLimiterContent.qml
```diff
@@ -80,7 +105,22 @@ Rectangle {
...
+                            duration: 200
+                            easing.type: Easing.OutCubic
+                        }
+                    }
+
+                    Behavior on color {
+                        ColorAnimation {
+                            duration: 200
+                        }
+                    }
```
end-4: for these two behaviors use these
`animation: Appearance.animation.elementMoveFast.numberAnimation.createObject(this)`
`animation: Appearance.animation.elementMoveFast.colorAnimation.createObject(this)`

reakjra:
I'll fix when I'm home. As I said in another PR, I'm far from confident with this, I mostly look at codes from other PRs and files and adjust. So be patient if there's several mistakes and inconveniences.

dots/.config/quickshell/ii/modules/overlay/OverlayContext.qml
end-4: No, that would be even more misleading, this is not a performance monitor

dots/.config/quickshell/ii/modules/overlay/fpsLimiter/FpsLimiterContent.qml
reakjra: what do you mean? You want a quick spin that looks like a flip effect? Or a tilt? 'Cause just changing that will literally make it go to 90° which doesn't really make sense for the `check` icon. Or am I missing something?

end-4: If there's any spinning, spin 90 or 180 degrees. The "close" icon should be safe with such rotations.
But it won't work well with save/check icons

reakjra: 360 degrees:

https://github.com/user-attachments/assets/55464bf3-be3a-4bb1-ab28-3f6044e62481

90 degrees:

https://github.com/user-attachments/assets/645958de-6fb0-447a-9ea4-3892c53bb1b8


This is what I meant. The cross and save icons are fine, the only one problematic is the check one.

end-4:
one last problem: whenever a widget is pinned, reopening the overlay would instantly dismiss it

end-4:
thanks for the pr

reakjra:
You're welcome! Forgive me for making you fix most of the code. Oh by the way, if you ever want to fix the cursor thing on the overview too and implement a way to add/edit entries for MangoHud config files through `~/.config/illogical-impulse/config.json`, it would be amazing

PR #2389, opened 2025-11-06
Position settings for overview/search bar

vaguesyntax:
## Describe your changes

- Added position settings for overview/search bar in to **Interface** settings

### Preview (bottom and center)

| with overview | without overview |
| ----------- | ----------- |
| image | image |
| image | image |

#### Settings

image

- Yep, it’s ready.

_Side note: I used GridLayout instead of ColumnLayout because `Layout.row` is only available for GridLayout._

### DIFF
--- a/dots/.config/quickshell/ii/modules/overview/Overview.qml
+++ b/dots/.config/quickshell/ii/modules/overview/Overview.qml
@@ -124,7 +124,8 @@ Scope {
                 Item {
                     // Filler item for centered position
                     visible: overviewScope.position === "center" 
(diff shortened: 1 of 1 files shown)


end-4:
I can understand the bottom mode, but the center is not really usable... a moving typing area isn't a good idea

end-4:
Maybe offset 1/3 or 1/4 of the screen from the top if "Center" mode is selected? 
image

vaguesyntax:
Bottom mode does not work after the last updates. I may re-open this (or in a new PR) after fixing it.

PR #2393, opened 2025-11-07
Provide Fedora support

ririko6z:
## Describe your changes

Support for the Fedora distro.

### DIFF
--- a/sdata/dist-fedora/install-deps.sh
+++ b/sdata/dist-fedora/install-deps.sh
@@ -12,7 +12,7 @@ case $SKIP_SYSUPDATE in
   *) v sudo dnf upgrade --refresh -y;;
 esac
 
-# Group installation
+# Development-tools installation
 v sudo dnf install @development-tools fedora-packager rpmdevtools fonts-rpm-macros -y
 
 # COPR repositories
@@ -69,7 +69,7 @@ v sudo dnf install python3{,.12}{,-devel} -y
 
 # Quickshell-git
 quickshell_deps=(
-  quickshell-git qt6-qtdeclarative qt6-qtbase jemalloc qt6-qtsvg pipewire-libs
+  qt6-qtdeclarative qt6-qtbase jemalloc qt6-qtsvg pipewire-libs
   libxcb wayland-devel qt6-qtwayland qt5-qtwayland libdrm breakpad
 )
 # NOTE: Below are custom dependencies of illogical-impulse
@@ -80,6 +80,15 @@ quickshell_custom_deps=(
 )
 v sudo dnf install "${quickshell_deps[@]}" -y
 v sudo dnf install "${quickshell_custom_deps[@]}" -y
+# Dynamically control the version of quickshell
+qs_version=$(dnf list --showduplicates quickshell-git 2>/dev/null |
+  grep -E "\.git${_qs_shortcommit}(-|$)" |
+  awk '{print $2}' |
+  sort -V |
+  tail -n1)
+v sudo dnf versionlock delete quickshell-git 2>/dev/null
+v sudo dnf install quickshell-git-${qs_version}
+v sudo dnf versionlock add quickshell-git
 
 # Screencapture
 v sudo dnf install hyprshot slurp swappy tesseract tesseract-langpack-eng tesseract-langpack-chi_sim wf-recorder -y
--- a/sdata/lib/environment-variables.sh
+++ b/sdata/lib/environment-variables.sh
@@ -25,3 +25,7 @@ STY_RST='\e[00m'
 
 # Used by register_temp_file()
 declare -a TEMP_FILES_TO_CLEANUP=()
+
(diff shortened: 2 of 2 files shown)


sdata/dist-fedora/README.md
```diff
@@ -0,0 +1,12 @@
+# Install scripts for Fedora Linux
+
+- Synchronize all dependencies in arch.
+- I've tested it on Fedora 43 (KDE Plasma Desktop Edition) on the x86_64 platform.
```
clsty:
```suggestion
- It has been tested on Fedora 43 (KDE Plasma Desktop Edition) on the x86_64 platform.
```

sdata/dist-fedora/install-setups.sh
clsty, line 1: Should avoid directly writing `/usr/**`. Please make custom RPM package in `install-deps.sh` instead.

sdata/dist-fedora/install-deps.sh
```diff
@@ -0,0 +1,91 @@
...
+)
+v sudo dnf install ${themes_deps[@]} -y
+
+# Hyprland 
+hyprland_deps=(
+  hypridle hyprland-qtutils hyprland hyprlock hyprpicker 
+  hyprsunset xdg-desktop-portal-hyprland wl-clipboard
+)
+v dnf download hyprland-qt-support
+v sudo rpm -ivh --nodeps hyprland-qt-support-*.fc$(rpm -E %fedora).$(uname -m).rpm
```
clsty: `hyprland-qtutils` and `hyprland-qt-support` have been removed in `dist-arch` recently. Please keep up with the **latest** `dist-arch`.

clsty (changes requested):

clsty:
About custom RPM package:
- It should be similar as PKGBUILD for Arch, as far as I know it seems to be Spec file.
- And it should be built during installation, similar to how we use `makepkg` for Arch, as far as I know it seems to be `rpmbuild`.

sdata/dist-fedora/install-setups.sh
ririko6z: Yes, I agree with you, doing it this way is more standardized.

clsty: As I said such code goes to install-deps.sh, not install-setups.sh

sdata/dist-fedora/install-deps.sh
```diff
@@ -0,0 +1,113 @@
...
+v sudo dnf install xdg-desktop-portal{,-gtk,-kde,-hyprland} -y
+
+# Python
+v sudo dnf install --setopt="install_weak_deps=False" clang uv gtk4-devel libadwaita-devel \
+  libsoup3-devel libportal-gtk4 gobject-introspection-devel -y
+v sudo dnf install python3{,.12}{,-devel} -y
+
+# Quickshell-git
+quickshell_deps=(
+  quickshell-git qt6-qtdeclarative qt6-qtbase jemalloc qt6-qtsvg pipewire-libs
```
clsty: For Quickshell, See #2302

sdata/dist-fedora/install-setups.sh
```diff
@@ -0,0 +1,21 @@
...
+v bash -c "echo uinput | sudo tee /etc/modules-load.d/uinput.conf"
+v bash -c 'echo SUBSYSTEM==\"misc\", KERNEL==\"uinput\", MODE=\"0660\", GROUP=\"input\" |
+ sudo tee /etc/udev/rules.d/99-uinput.rules'
+v systemctl --user enable ydotool
+v sudo systemctl enable bluetooth --now
+v gsettings set org.gnome.desktop.interface font-name 'Rubik 11'
+v gsettings set org.gnome.desktop.interface color-scheme 'prefer-dark'
+v kwriteconfig6 --file kdeglobals --group KDE --key widgetStyle Darkly
+v bash -c "echo exec-once = /usr/libexec/kf6/polkit-kde-authentication-agent-1 |
+ sudo tee -a ${REPO_ROOT}/dots/.config/hypr/custom/execs.conf"
```
clsty:
```suggestion
 sudo tee -a ${REPO_ROOT}/dots/.config/hypr/custom/execs.conf"
# TODO: This will modify Git repo itself which should be avoided, however it makes sense that this may be needed. Need to improve this in future.
```

clsty:
Mostly LGTM, now just need to pin commit of Quickshell (#2302).
I guess you need to create a Spec file for Quickshell to achieve this.
After that I think this PR can be merged.

ririko6z:
> Mostly LGTM, now just need to pin commit of Quickshell (#2302). I guess you need to create a Spec file for Quickshell to achieve this. After that I think this PR can be merged.

Thanks! I think Fedora users are different from Arch and Gentoo users; We need to minimize the number of builds.
Therefore, I implemented this feature using another method.

clsty:
Nice work, thanks for your contribution.

PR #2395, opened 2025-11-07
Allow pinning the left sidebar

vaguesyntax:
## Describe your changes

For me, the left sidebar was always the second option. Because, I couldn’t pin it.  _Detaching it turned it into a separate window, which made it harder to use_
So, I added an option to pin the sidebar directly instead.

- Added a keybind `Ctrl+I` to pin the left sidebar
- Improved usability when keeping the sidebar visible

## Preview

https://github.com/user-attachments/assets/fb0155fe-9977-4c75-9185-1dd3cbe725fe

I think it’s ready. The only thing that bothers me is that the sidebar _flashes/lags_ **(It's a minor issue, not too disturbing)** when I try to expand it while it’s pinned. I couldn’t find a proper solution for that yet.


dots/.config/quickshell/ii/modules/sidebarLeft/AiChat.qml
```diff
@@ -510,6 +512,8 @@ Inline w/ backslash and round brackets \\(e^{i\\pi} + 1 = 0\\)
                     placeholderText: Translation.tr('Message the model... "%1" for commands').arg(root.commandPrefix)
 
                     background: null
+                    
+
```
end-4, line 516: weird space again???

dots/.config/quickshell/ii/modules/sidebarLeft/SidebarLeft.qml
```diff
@@ -35,22 +35,41 @@ Scope { // Scope
         }
     }
 
+    
+
```
end-4, line 39: and again

dots/.config/quickshell/ii/modules/sidebarLeft/SidebarLeft.qml
```diff
@@ -35,22 +35,41 @@ Scope { // Scope
...
             function hide() {
                 GlobalStates.sidebarLeftOpen = false
             }
 
+            function updateWidth() {
+                if (pinned) {
+                    implicitWidth = sidebarRoot.extend ? Appearance.sizes.sidebarWidthExtended - Appearance.sizes.elevationMargin : Appearance.sizes.sidebarWidth - Appearance.sizes.elevationMargin
+                    exclusionMode = ExclusionMode.Auto
+                }else {
```
end-4, line 66: why do you finally put a space between `)` and `{` this time but not after `}` :woe:

dots/.config/quickshell/ii/modules/sidebarLeft/SidebarLeft.qml
```diff
@@ -71,15 +90,39 @@ Scope { // Scope
...
+                property bool currentFocus: true
+
+                TapHandler {   
+                    onTapped: {
+                        grab.active = true
+                        focusTapItem.currentFocus = false
+                    }
+                } 
+            }     
+
```
end-4, line 125: What's up with all this tap handling? Can't we just make kb focus on demand?

vaguesyntax: I am not sure how else we can detect user input and focus the sidebar. The sidebar doesn’t regain focus after losing it if we don’t use tap handling

end-4: okay, it's more complicated than i thought... hyprland is funny

dots/.config/quickshell/ii/modules/sidebarLeft/SidebarLeft.qml
vaguesyntax: Oops, I am sorry for those unintentional mistypes (weird spaces and this one). I have no access to a computer until sunday evening, so I can't fix those right now. :(

end-4:
I prefer that version of mine because the logic is flatter

end-4:
Funny thing is Ctrl+P was supposed to be for pinning, but it turned into detaching because pinning is funny and I couldn't figure it out before

vaguesyntax:
> Funny thing is Ctrl+P was supposed to be for pinning, but it turned into detaching because pinning is funny and I couldn't figure it out before

Yeah Ctrl P used to pin the sidebar in the ags version. Honestly, the detach behavior isn’t too bad either , kinda got its own use cases.

PR #2396, opened 2025-11-08
Turkish language support

PicoShot:
## Describe your changes

Added Turkish (tr_TR) language support because AI translation sucks.

Yes, it’s ready. I’ve fully translated all keys into native Turkish.


end-4:
Just a small sanity check, things look good so I'm merging now

PR #2397, opened 2025-11-08
Language dropdown

PicoShot:
## Describe your changes

  Replaced the horizontal array of language selection buttons in the settings menu with a beautiful dropdown
  button component.

  Ready for review.

## Preview 

image

image

### DIFF
--- a/dots/.config/quickshell/ii/modules/common/widgets/StyledComboBox.qml
+++ b/dots/.config/quickshell/ii/modules/common/widgets/StyledComboBox.qml
@@ -88,15 +88,27 @@ ComboBox {
     delegate: ItemDelegate {
         id: itemDelegate
         width: ListView.view ? ListView.view.width : root.width
-        height: 40
+        implicitHeight: 40
 
         required property var model
         required property int index
+        property color color: {
+            if (root.currentIndex === itemDelegate.index) {
+                if (itemDelegate.down) return Appearance.colors.colSecondaryContainerActive;
+                if (itemDelegate.hovered) return Appearance.colors.colSecondaryContainerHover;
+                return Appearance.colors.colSecondaryContainer;
+            } else {
+                if (itemDelegate.down) return Appearance.colors.colLayer3Active;
+                if (itemDelegate.hovered) return Appearance.colors.colLayer3Hover;
+                return ColorUtils.transparentize(Appearance.colors.colLayer3);
+            }
+        }
+        property color colText: (root.currentIndex === itemDelegate.index) ? Appearance.colors.colOnSecondaryContainer : Appearance.colors.colOnLayer3
 
         background: Rectangle {
             anchors.fill: parent
             radius: Appearance.rounding.small
-            color: root.currentIndex === itemDelegate.index ? Appearance.colors.colPrimary : itemDelegate.down ? Appearance.colors.colSecondaryContainerActive : itemDelegate.hovered ? Appearance.colors.colSecondaryContainerHover : ColorUtils.transparentize(Appearance.colors.colSurfaceContainerHigh)
+            color: itemDelegate.color
 
             Behavior on color {
                 animation: Appearance.animation.elementMoveFast.colorAnimation.createObject(this)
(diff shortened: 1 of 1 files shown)


dots/.config/quickshell/ii/modules/common/widgets/LanguageDropdownButton.qml
end-4, line 1: Don't you think this should be more generic, like `DropDownButton`?

end-4 (commented):
This one pops down, pushing the content below, which is not very nice (UX, UI is also debatable but less so). The long config button array is better. Either this one be made a proper combobox or we keep the old array

dots/.config/quickshell/ii/modules/common/widgets/LanguageDropdownButton.qml
PicoShot: > Don't you think this should be more generic, like `DropDownButton`?

yeah you're actually right i just do this on my free time, I'll make it a generic DropdownButton for general use and work more on the UI/UX

end-4:
You can use StyledComboBox for a proper drop down selection

PicoShot:
> You can use StyledComboBox for a proper drop down selection

i make something that look better but the current problem is it not static and move as i scroll 

image


the issue:

image

i currently working on it to maybe find a way to make it fixed

end-4:
The very clear border does not look better. Material Design typically uses colors and shadows to create elevation, and in cases where borders are used, they're not this clear

PicoShot:
> The very clear border does not look better. Material Design typically uses colors and shadows to create elevation, and in cases where borders are used, they're not this clear

i make it better and fix the issues

image

dots/.config/quickshell/ii/modules/common/widgets/DropdownButton.qml
```diff
@@ -0,0 +1,276 @@
...
+    function getCurrentDisplayName() {
+        for (let i = 0; i < options.length; i++) {
+            if (options[i].value === currentValue) {
+                return options[i].displayName;
+            }
+        }
+        return placeholder;
+    }
+
+    GroupButton {
```
end-4: Since it's not in a group, don't use a GroupButton. Use a RippleButton

dots/.config/quickshell/ii/modules/common/widgets/DropdownButton.qml
```diff
@@ -0,0 +1,276 @@
...
+        repeat: true
+        running: root.popupVisible
+        onTriggered: {
+            if (popupContent.visible) {
+                var pos = root.getGlobalPosition();
+                popupContent.x = pos.x;
+                popupContent.y = pos.y + root.height + 4;
+            }
+        }
+    }
```
end-4: what the hell... does anchoring the menu to the button not work?

hmm i guess we might not have proper click-to-dismiss that way

actually, why not style a real ComboBox? the StyledComboBox I'm having uses QtQuick's Material style but I think it should be customizable (define contentItem and stuff, like what StyledSlider does to Slider)

PicoShot: hmm i never try that. I'll switch to using Qt's built-in ComboBox with custom styling but i'm not sure how good is it

PicoShot:
I've switched from the custom dropdown menu to ComboBox with custom styling (following the StyledSlider pattern). I think i can work more on the UI and animations if needed let me know if you have any suggestions or changes you'd like to see

image
image

PicoShot:
ready for review and do we need [DropdownButton.qml](https://github.com/end-4/dots-hyprland/pull/2397/files#diff-241beb6865565796c36dba5f4dd998c7e54ba523b2d48391653c34a831f260f5) anymore?

PicoShot:
Ready for review.

PicoShot:
I think it’s ready for review I’ve done my best to refine and improve it based on what came to mind during development  

here are a few pictures of the final result:

image

image

Additionally, I experimented with updating the **Quick** tab in Settings to use a fully combo-box layout here's how that looks:

image

however I haven’t committed those changes since this PR focuses specifically on the **language combobox** feature

end-4:
Questionable contrast:

image
image

end-4:
And please use a triangle like before instead of the up down arrows

end-4:
```
WARN scene: @modules/common/widgets/StyledComboBox.qml[107:25]: Unable to assign [undefined] to QString
WARN scene: @modules/common/widgets/StyledComboBox.qml[99:17]: Unable to assign [undefined] to bool
```

dots/.config/quickshell/ii/modules/settings/GeneralConfig.qml
```diff
@@ -153,14 +152,26 @@ ContentPage {
...
+                    })]
+
+                currentIndex: {
+                    for (let i = 0; i < model.length; i++) {
+                        if (model[i].value === Config.options.language.ui) {
+                            return i;
+                        }
+                    }
+                    return 0;
+                }
```
end-4, line 160: Can't you just use `find()`?

end-4:
The popup needs a shadow
image

dots/.config/quickshell/ii/modules/common/widgets/StyledComboBox.qml
```diff
@@ -1,15 +1,166 @@
...
+        width: ListView.view ? ListView.view.width : root.width
+        height: 40
+
+        required property var model
+        required property int index
+
+        background: Rectangle {
+            anchors.fill: parent
+            radius: Appearance.rounding.small
+            color: root.currentIndex === itemDelegate.index ? Appearance.colors.colPrimary : itemDelegate.down ? Appearance.colors.colSecondaryContainerActive : itemDelegate.hovered ? Appearance.colors.colSecondaryContainerHover : "transparent"
```
end-4: Don't use "transparent". It makes the color flash black when animating. Use the appropriate color but transparentized instead.

PicoShot:
Hi @end-4,
Thanks for the review, I've addressed all the issues you mentioned, everything should be ready now.

end-4:
Thanks for your contribution
After some adjustments it's ready and I'm merging this now

PR #2403, opened 2025-11-08
Disable Bluetooth Android button for systems without Bluetooth

rsubtil:
## Describe your changes

Basically the exact same of https://github.com/end-4/dots-hyprland/pull/2167, but for the new Android style buttons

Yes


end-4:
Actually, with the current toggles placement it would be weird to totally hide the button like this. The user wouldn't even be able to remove it
image

end-4:
In 3c25d18f88937e22048451720c344d6dee33b998 I made the button grey out instead of completely hide

PR #2405, opened 2025-11-08
Hide resource details on ResourcesPopup if disabled in configuration

rsubtil:
## Describe your changes

This also hides detailed CPU & Swap info from the resources popup when they're hidden in the settings:

image


end-4:
the option is "always", not "show at all"

the tooltip is supposed to show more complete info

PR #2406, opened 2025-11-08
fix fedora install-deps crashing on hyprland-qt-support step

possum-enjoyer:
## Describe your changes

fix a missing v function invocation that leads to a crash if the hyprland-qt-support package is already installed. This can happen if the config ran in the past which would make the script not idempotent anymore.

On the same note: Is it ok that stuff like oneui/microtex... basically anything from the package-installers.sh except for the python part is missing from the fedora installation process?


clsty:
Thanks for the fix.

clsty:
> On the same note: Is it ok that stuff like oneui/microtex... basically anything from the package-installers.sh except for the python part is missing from the fedora installation process?

It depends on how you define "ok". Missing package leads to missing functionalities and even unexpected behavior.

possum-enjoyer:
> > On the same note: Is it ok that stuff like oneui/microtex... basically anything from the package-installers.sh except for the python part is missing from the fedora installation process?
> 
> It depends on how you define "ok". Missing package leads to missing functionalities and even unexpected behavior.

Yeah I thought so, i was just unsure because almost everything from package-installers.sh is absent from the fedora install script and I that looked like it was a deliberate for some reason (maybe I overlooked or so). But if that's not the case I will try to include it as best as I can and as close as it's done in dist-arch. 
If that's not possible fallback-dist has a script that could be used and it's working, as I used it in the past for fedora

clsty:
Oh, wait a minute, you mean `package-installers.sh`? Except for python things, we recommend against using it normally because it writes to `/usr/**` without being managed by system package manager which should actually be avoided if possible.

PR #2412, opened 2025-11-09
fix: Change quickshell-git and hyprland-qt-support to local builds.

ririko6z:


clsty:
LGTM, merging.

Thanks for the fix!

PR #2413, opened 2025-11-09
fix: quote overlap in locksreen

MannuVilasara:
## Describe your changes
the quote & locked text on lockscreen are at the same position causing overlap. I tried changing position but it doesn't look good. I think it would be better to show it only when we're not locked. still let me know if we need both. I'll push in the changes

image

yes, let me know if we need any change


end-4:
Nah... we should make it properly placed instead of just hiding it

end-4:
I fixed it properly. Turns out I had some weird anchors

PR #2414, opened 2025-11-09
Re-add version lock for quickshell-git.

ririko6z:
## Re-add version lock for quickshell-git.
----


sdata/dist-fedora/install-setups.sh
clsty, line 1: Why do you put it into `install-setups.sh`? By design such code should be written in `install-deps.sh`.

ririko6z: I think `install-deps` belongs in the main installation flow, and `install-setup` should be reserved for post-install tasks like version locking to disable auto-updates.
If you insist, you can put it at the end of `install-deps`.

clsty:
Merging. Thanks for the fix.

PR #2415, opened 2025-11-09
Brightness fix for multiple devices by percentage

imitoy:
## Problem

The brightness module culculates the brightness depending on the default deivce (for me, it is `nvidia_0`), which is not the truly effective one. In my case, the effective one is called `intel_backlight`, the max brightness of which is `496` while `100` for `nvidia_0`. This lead to incorrect brightness calculations because the module uses the maximum brightness of the wrong display(for me, it sets the brightness as the maximum `100` for both `nvida_0` and `intel_backlight`).

## Solution
Use percentage brightness instead, set it to all devices in the class `backlight`. This also make sure that initially it can get the right brightness.

## Changes
`dots/.config/quickshell/ii/services/Brightness.qml`

## Test
My local machine only
```
brightnessctl --list --class backlight
Available devices:
Device 'nvidia_0' of class 'backlight':
	Current brightness: 1 (1%)
	Max brightness: 100

Device 'intel_backlight' of class 'backlight':
	Current brightness: 1 (0%)
	Max brightness: 496
```

### DIFF
--- a/dots/.config/quickshell/ii/services/Brightness.qml
+++ b/dots/.config/quickshell/ii/services/Brightness.qml
@@ -152,7 +152,7 @@ Singleton {
         }
 
         function setBrightness(value: real): void {
-            value = Math.max(0, Math.min(value, 1));
+            value = Math.max(0, Math.min(1, value));
             monitor.brightness = value;
         }
 


imitoy:
There might be a bug

imitoy:
oh, the "bug" which i mentioned was `BrightnessMultiplier`😮

end-4:
> The brightness module culculates the brightness depending on the default deivce (for me, it is nvidia_0), which is not the truly effective one

Then the correct fix would be to use `--class backlight`. Please try changing this line
```qml
initProc.command = isDdc ? ["ddcutil", "-b", busNum, "getvcp", "10", "--brief"] : ["sh", "-c", `echo "a b c $(brightnessctl g) $(brightnessctl m)"`];
```
to
```qml
initProc.command = isDdc ? ["ddcutil", "-b", busNum, "getvcp", "10", "--brief"] : ["sh", "-c", `echo "a b c $(brightnessctl --class backlight g) $(brightnessctl --class backlight m)"`];
```

imitoy:
That means that the module will have to parse the full output of `brightnessctl` . A single command can only set the same brightness value for multiple devices.

Do you have any better plans?

imitoy:
> That means that the module will have to parse the full output of brightnessctl .

I'll try it

end-4:
> That means that the module will have to parse the full output of `brightnessctl` . A single command can only set the same brightness value for multiple devices.

No? If this is replying to my message... I'm saying you should test if that one line change (to upstream) also fixes your problem

image

end-4:
Setting with a percentage instead of raw value does not make it correct, it only makes failure safer.... okay well that's probably also good for the user end, but doesn't solve the problem of interacting with the wrong device

imitoy:
> No? If this is replying to my message...

I'm sorry for not using the reply function; I'm still getting familiar with the process. Thank you for review.

>I'm saying you should test if that one line change (to upstream) also fixes your problem

I've tried it before. The `brightnessctl` does list all the properties with `-c backlight`:
image

But it can not set multiple devices with different brightness values in only one command:
image

That's why I said "A single command can only set the same brightness value for multiple devices." Sorry again for not explaining what I had already tried earlier.

The source code is [here](https://github.com/Hummer12007/brightnessctl/blob/master/brightnessctl.c#L276). For referrence:
```c
return apply_operation(dev, p.operation, &p.val);
//p is a global variable
//&p.val means that it can only set brightness to a same value
```

> Setting with a percentage instead of raw value does not make it correct, it only makes failure safer.... okay well that's probably also good for the user end, but doesn't solve the problem of interacting with the wrong device

That's absolutely right, but I assumed that solving the problem requires significant changes. That's why the percentage plan had been chosen first.

> I'll try it

"It" referred to the "significant changes".

end-4:
I see. For now this works I guess

PR #2419, opened 2025-11-09
feat: Add keyboard brightness control keybindings

4urora3night:
Added keybindings for keyboard brightness control.

## Describe your changes

Added a binding in `keybindings.conf` to allow keyboard brightness to be changed via:
`Alt`+`XF86MonBrightness{Down,Up}` or
`XF86KbdBrightness{Up,Down}`

Just need to check if it works with the whole installation process, I won't be doing this.
I don't know how qs ipc works so that is another task to do


end-4 (commented):
I can't test it, but seems to be correct.
Don't you think we should use percentages instead of raw value? like `10%+` instead of `10+`

4urora3night:
yes, sounds good :+1:

4urora3night:
Been a few months, if you are not busy @end-4 ,feel free to accept the pull request and make any changes.

PR #2428, opened 2025-11-10
Add Rofi navigation & editing in search box and overview search (close #255)

### LINKED ISSUE #255: [Feature] Add a set of common keybinds to launcher widget
# **What would you like to be added?**
For the launcher widget (which keybinding is `Super`), add following keybindings:
- `Ctrl+N`: Select **n**ext item (same as `Down` arrow key).
- `Ctrl+P`: Select **p**revious item (same as `Up` arrow key).
- `Ctrl+F`: Cursor move **f**orward (same as `Right` arrow key).
- `Ctrl+B`: Cursor move **b**ackward (same as `Left` arrow key).
- `Ctrl+A`: Cursor moves to begin of line (same as `Home` key).
- `Ctrl+E`: Cursor moves to **e**nd of line (same as `End` key).
- `Ctrl+U`: Deletes from "begin of line" to cursor. (**Already implemented actually**)
- `Ctrl+K`: Deletes from cursor to "end of line".

# **How will it help**
## 1. Consistency and Routine
Such keybinds are widely used in Linux.

Here're some examples:
### Shells
- Example programs: `bash`, `zsh`, `fish`
- Behavior: `Ctrl+N/P/F/B` equals to `Down/Up/Right/Left` arrow keys, `Ctrl+A/E` moves to begin/end of line, `Ctrl+U/K` deletes between cursor and begin/end of line.
### Editors
- Example programs: Emacs
- Behavior: `Ctrl+N/P/F/B` equals to `Down/Up/Right/Left` arrow keys, `Ctrl+A/E` moves to begin/end of line, `Ctrl+K` deletes between cursor and end of line (`Ctrl+U` binds to another function though).

- Example programs: Nano
- Behavior: `Ctrl+N/P/F/B` equals to `Down/Up/Right/Left` arrow keys, `Ctrl+A/E` moves to begin/end of line (`Ctrl+U/K` binds to another function though).
### Launchers
- Example programs: Fuzzel, e.g. below command:
```bash
pkill fuzzel || fuzzel --icon-theme=candy-icons --background-color=1D1D27ee --text-color=cdd6f4ff --match-color=AC16C7FF --border-width=3 --border-radius=10 --border-color=8701E7FF	 --selection-color=585b70ff --selection-text-color=cdd6f4ff --selection-match-color=AC16C7FF --font="Torus 16"  --prompt="    "
```
- Behavior: `Ctrl+N/P/F/B` equals to `Down/Up/Right/Left` arrow keys, `Ctrl+A/E` moves to begin/end of line, `Ctrl+K` deletes between cursor and end of line (`Ctrl+U` is not implemented though).

## 2. Handy and efficient
The `Down/Up/Right/Left` arrow keys locates on the right-bottom corner,
which is far from the main type area (i.e. character keys `A`/`S`/`D`/`F` for the left hand, `J`/`K`/`L`/`;` for the right hand).
Therefore, to press those keys, the whole right **arm** (not just fingers) must move a distance, which cost a second or two (and to precisely press the key instead of the keys nearby, it's even slower). Also it's not very comfortable when wrist float in the air.

Similarly for `Home/End` keys, which are even farther.

## 3. No harm
This will not affect the people who do not use such keybinds.

locutus3009:
## Describe your changes

Add Rofi navigation in search results in overview's SearchWidget

Closes #255 

This CL is ready

- [x]    Ctrl+N: Select next item (same as Down arrow key).
- [x]    Ctrl+P: Select previous item (same as Up arrow key).
- [x]    Ctrl+F: Cursor move forward (same as Right arrow key).
- [x]    Ctrl+B: Cursor move backward (same as Left arrow key).
- [x]    Ctrl+A: Cursor moves to begin of line (same as Home key).
- [x]    Ctrl+E: Cursor moves to end of line (same as End key).
- [x]    Ctrl+U: Deletes from "begin of line" to cursor. (current keybinding deletes the whole line)
- [x]    Ctrl+K: Deletes from cursor to "end of line".
- [x]    Ctrl+D: Forward deletion.

### DIFF
--- a/dots/.config/quickshell/ii/modules/ii/overview/SearchWidget.qml
+++ b/dots/.config/quickshell/ii/modules/ii/overview/SearchWidget.qml
@@ -151,7 +151,7 @@ Item { // Wrapper
         if (event.key === Qt.Key_Escape)
             return;
 
-        // Add Emacs navigation in search results
+        // Add Rofi navigation in search results
         if (event.modifiers & Qt.ControlModifier & Config.options.interactions.searchWidget.rofiStyleNavigation) {
             switch (event.key) {
                 case Qt.Key_N:
@@ -267,7 +267,7 @@ Item { // Wrapper
                 }
             }
 
-            // Add Emacs navigation & editing inside search box
+            // Add Rofi navigation & editing inside search box
             Connections {
                 target: searchBar.searchInput
                 function onActiveFocusChanged() {


clsty:
Is it possible to implement #255 ?

locutus3009:
I think it is possible, let me try it. I will turn this to the draft

locutus3009:
Done

dots/.config/quickshell/ii/modules/ii/overview/SearchWidget.qml
```diff
@@ -249,6 +267,71 @@ Item { // Wrapper
...
+                                let textU = searchBar.searchInput.text;
+                                let posU = searchBar.searchInput.cursorPosition;
+                                searchBar.searchInput.text = textU.slice(posU);
+                                searchBar.searchInput.cursorPosition = 0;
+                                break;
+
+                            case Qt.Key_A:
+                                event.accepted = true;
+                                searchBar.searchInput.cursorPosition = 0;
+                                break;
```
end-4, line 314: This conflicts with the select all action. Either drop this or provide some "prefer emacs navigation" in config option
Maybe `interactions.text.navigationStyle` as a string, default to "normal", offer "emacs" for ctrl+a to do this

dots/.config/quickshell/ii/modules/ii/overview/SearchWidget.qml
```diff
@@ -249,6 +267,71 @@ Item { // Wrapper
...
             }
 
+            // Add Emacs navigation & editing inside search box
+            Connections {
+                target: searchBar.searchInput
+                function onActiveFocusChanged() {
+                    if (searchBar.searchInput.activeFocus) {
+                        searchBar.searchInput.Keys.pressed.connect(handleSearchKeys);
+                    }
+                }
```
end-4, line 277: Can't we just do this?
```qml
enabled: searchBar.searchInput.activeFocus
```

Wait, why not put this in SearchBar.qml directly? It's text manipulation...

Additionally, Ctrl+D from my understanding of the code deletes one char, but from my testing somehow deletes the whole word. This behavior is consistent with the code when the key handling logic is moved to SearchBar. I don't know Emacs so please clarify what's intended

clsty:
> Mostly works. I'll trust the binds are correct because I do not use Emacs

@end-4 I should clearify that Emacs uses them but it's not Emacs-like keybinding but more general thing that widely adopted just as described in the examples in #255, including shells, editors, launchers...

i.e. AFAIK it's probably Linux style, not just Emacs style. As for Ctrl-A it's windows-style.

end-4:
It's probably clearer to call it Emacs style. Graphical programs on Linux use ctrl a for select all, so calling it "Linux style" would be too broad imo

clsty:
> It's probably clearer to call it Emacs style. Graphical programs on Linux use ctrl a for select all, so calling it "Linux style" would be too broad imo


As described in #255, Emacs does not use Ctrl-U to clear the line by default. That's one reason why it's not suitable to call it Emacs style.

~~I guess something like Linux-tradition style is more suitable.~~

Maybe shell or CLI or Rofi style. Linux-tradition sounds like something outdated which is not true for those keybinds.

end-4:
Actually, since only ctrl+a is problematic here, we could just make the config option say something like "ctrlAAsHome" with description "ctrl+a to move cursor to beginning"

And we simply not mention emacs or any other program at all

dots/.config/quickshell/ii/modules/ii/overview/SearchWidget.qml
locutus3009: I will provide a config for this

dots/.config/quickshell/ii/modules/ii/overview/SearchWidget.qml
locutus3009: Let me check. C-d deletes one char, M-d deletes the whole word/semantically meaningful set of characters in case of Emacs.

locutus3009:
Ok, let me fix your suggestions. Today I don't have much time on it, I'll try to access these comments on Weekend.

locutus3009:
Add two configuration options; @clsty please check if it works for you

dots/.config/quickshell/ii/modules/ii/overview/SearchWidget.qml
locutus3009: Provided two additional configs

dots/.config/quickshell/ii/modules/ii/overview/SearchWidget.qml
locutus3009: I put it into the widget because there were already the other hotkey processing. It started from navigation in the search results, so I have decided to leave general text manipulation in the same file

locutus3009:
Please review the updated version

locutus3009:
OK closing it for now. If somebody wants this functionality, it can be used from my fork.

PR #2431, opened 2025-11-10
Feature (Bar): New Setting for Tooltip Behavior on the Bar

firminoveras:
## Describe your changes

https://github.com/user-attachments/assets/b6729da8-83ff-43ea-9fac-8f9adcd331c5

This PR introduces a new setting called **Click to show**, which allows users to control how the bar tooltips are displayed.

A new switch has been added to the shell configuration:

- **Click to Show _enabled_** — Tooltips are only displayed after the user clicks on the item.

- **Click to Show _disabled_ (default)** — Tooltips behave normally and appear automatically following the default interaction pattern: appear on hover.

This feature gives users more control and helps avoid unwanted tooltip pop-ups during normal navigation.
To manually update weather information, use the right mouse button.

Yes.

### DIFF
--- a/dots/.config/quickshell/ii/modules/settings/BarConfig.qml
+++ b/dots/.config/quickshell/ii/modules/settings/BarConfig.qml
@@ -339,7 +339,7 @@ ContentPage {
         icon: "tooltip"
         title: Translation.tr("Tooltips")
         ConfigSwitch {
-            buttonIcon: "web_traffic"
+            buttonIcon: "ads_click"
             text: Translation.tr("Click to show")
             checked: Config.options.bar.tooltips.clickToShow
             onCheckedChanged: {


dots/.config/quickshell/ii/modules/common/Config.qml
```diff
@@ -263,6 +263,9 @@ Singleton {
                         property bool showUnreadCount: false
                     }
                 }
+                property JsonObject popupClick: JsonObject {
+                    property bool enable: false
+                }
```
end-4: Let's call the category `tooltips` and the bool `clickToShow`. It makes more sense because if we add a timeout later it would also reasonably go into this category

dots/.config/quickshell/ii/modules/settings/BarConfig.qml
```diff
@@ -334,4 +334,31 @@ ContentPage {
...
+                        value: false
+                    },
+                    {
+                        displayName: Translation.tr("Click"),
+                        icon: "web_traffic",
+                        value: true
+                    }
+                ]
+            }
+        }
```
end-4, line 348: We're just toggling a bool so use a switch. A fancy config selection array is not necessary

Issue #2374, opened 2025-11-04
Allow different clock style for lock screen


HmzNasry:
### What would you like to be added?

The option  for the style of the clock (digital vs. cookie) in  the lock screen.

### How will it help?

It will help because cookie might look better with your wallpaper, but in lock screen, digital looks more appropriate and better, so it would be nice to have that option.

end-4:
solved in #2433

PR #2433, opened 2025-11-10
clock: separate style for lock screen (#2374)

Smartyguy001:
## Describe your changes

Added the option to have a separate clock style for the lock screen. This also came with adding a single variable to the Config file.
Along with it, added some handling in the settings for visibility of digital and cookie styles (eg. if regular clock style is cookie and lock screen style is digital, *but* the clock is only visible when locked, then only the settings for digital clock will be visible).

It should be pretty much ready, but I'm open to any suggestions!

### DIFF
--- a/dots/.config/quickshell/ii/modules/common/Config.qml
+++ b/dots/.config/quickshell/ii/modules/common/Config.qml
@@ -158,7 +158,7 @@ Singleton {
                         property real x: 100
                         property real y: 100
                         property string style: "cookie"        // Options: "cookie", "digital"
-                        property string styleLocked: "digital" // Options: "cookie", "digital"
+                        property string styleLocked: "cookie"  // Options: "cookie", "digital"
                         property JsonObject cookie: JsonObject {
                             property bool aiStyling: false
                             property int sides: 14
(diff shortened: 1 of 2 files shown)


dots/.config/quickshell/ii/modules/settings/BackgroundConfig.qml
```diff
@@ -55,6 +55,10 @@ ContentPage {
     ContentSection {
         icon: "clock_loader_40"
         title: Translation.tr("Widget: Clock")
+        id: settingsClock
+
+        readonly property bool digitalPresent: (!Config.options.background.widgets.clock.showOnlyWhenLocked && Config.options.background.widgets.clock.style === "digital") || Config.options.background.widgets.clock.styleLocked === "digital"
+        readonly property bool cookiePresent: (!Config.options.background.widgets.clock.showOnlyWhenLocked && Config.options.background.widgets.clock.style === "cookie") || Config.options.background.widgets.clock.styleLocked === "cookie"
```
end-4: small nitpick but these two lines could share a function

dots/.config/quickshell/ii/modules/common/Config.qml
```diff
@@ -157,7 +157,8 @@ Singleton {
                         property string placementStrategy: "leastBusy" // "free", "leastBusy", "mostBusy"
                         property real x: 100
                         property real y: 100
-                        property string style: "cookie" // Options: "cookie", "digital"
+                        property string style: "cookie"        // Options: "cookie", "digital"
+                        property string styleLocked: "digital" // Options: "cookie", "digital"
```
end-4: make both cookie by default please

PR #2436, opened 2025-11-10
Feature (Overlay): Webview

reakjra:
## Describe your changes

WebView overlay widget. 

Requires Qt6-webview dependency

https://github.com/user-attachments/assets/47c7451c-81ae-4cc8-9ac9-692448387bd9

Works fine, should be ready when quickshell will fix the QtWebView thing if interested. Just gonna leave it here 

EDIT: Might consider QtWebEngine instead for local caching (better for things like interactive maps etc....................)


laiviettrung:
This guy is going to make his own browser. :fire:

xBiei:
I think waiting for https://github.com/quickshell-mirror/quickshell/pull/351 will make this feature better.

reakjra:
in fact without that fix it simply won't work

xBiei:
> in fact without that fix it simply won't work

well yeah it crashes the whole thing, but what I meant is, it's probably better not to integrate it before that pr is merged

YerCodeWorld:
I was actually doing something similar but for a desktop widget instead of an overlay, still waiting to see if the WebView initialization in quickshell gets implemented, I have a nice project idea.

Issue #2350, opened 2025-10-31
Code snippets render weird in ai


Faithxull:
### Step 1. Before you submit

- [x] I have read the [Troubleshooting](https://ii.clsty.link/en/ii-qs/04troubleshooting/) and [Usage](https://ii.clsty.link/en/ii-qs/02usage/) pages.
- [x] I've successfully updated to the latest version following the [guidance](https://ii.clsty.link/en/ii-qs/01setup/#updating).
- [x] I've successfully updated the system packages to the latest.
- [ ] I've ticked the checkboxes without reading their contents

### Step 2. Quick diagnose info

<details><summary>Quick diagnose</summary>

```
<!-- Run `./diagnose` inside the repo, and paste the result here! -->
```
/========================\
| Checking git repo info |
\========================/

[===diagnose===] git remote get-url origin
https://github.com/end-4/dots-hyprland.git
[---SUCCESS---]

[===diagnose===] git rev-parse HEAD
43aae4ee56ac5a3e59275630a9b69ad4c3904860
[---SUCCESS---]

/=================\
| Checking distro |
\=================/

[===diagnose===] ii_check_distro_id
distro ID: arch
distro ID_LIKE: 
[---SUCCESS---]

[===diagnose===] ii_check_distro
LSB Version:	n/a
Distributor ID:	Arch
Description:	Arch Linux
Release:	rolling
Codename:	n/a
[---SUCCESS---]

/====================\
| Checking variables |
\====================/

[===diagnose===] declare -p XDG_CACHE_HOME
./diagnose: line 36: declare: XDG_CACHE_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p XDG_CONFIG_HOME
./diagnose: line 36: declare: XDG_CONFIG_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p XDG_DATA_HOME
./diagnose: line 36: declare: XDG_DATA_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p XDG_STATE_HOME
./diagnose: line 36: declare: XDG_STATE_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p ILLOGICAL_IMPULSE_VIRTUAL_ENV
declare -x ILLOGICAL_IMPULSE_VIRTUAL_ENV="~/.local/state/quickshell/.venv"
[---SUCCESS---]

/============================\
| Checking directories/files |
\============================/

[...]

PR #2471, opened 2025-11-14
Fix issue. Codeblock/Textblock/Thinkblock content can't be fully rendered

ItsSunshineXD:
## Describe your changes
fix issue #2350 

yes i've fully tested on ArchLinux Linux 6.17.7-arch1-1

### DIFF
--- a/dots/.config/quickshell/ii/modules/ii/sidebarLeft/aiChat/AiMessage.qml
+++ b/dots/.config/quickshell/ii/modules/ii/sidebarLeft/aiChat/AiMessage.qml
@@ -286,7 +286,9 @@ Rectangle {
                 }
             }
             Repeater {
-                model: root.messageBlocks
+                model: ScriptModel {
+                    values: root.messageBlocks
+                }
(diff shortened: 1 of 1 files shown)


ItsSunshineXD:
Test:
image

dots/.config/quickshell/ii/modules/ii/sidebarLeft/aiChat/AiMessage.qml
```diff
@@ -285,48 +285,34 @@ Rectangle {
...
             Repeater {
-                model: ScriptModel {
-                    values: Array.from({ length: root.messageBlocks.length }, (msg, i) => {
-                        return ({
-                            type: root.messageBlocks[i].type
-                        })
-                    });
-                }
-
+                model: root.messageBlocks
```
end-4: You should still use ScriptModel here to avoid unnecessary destruction+recreations

end-4 (commented):
seems to work, one small problem

end-4:
thank you for the fix

PR #2479, opened 2025-11-14
Added random wallpaper explicitness rating option in the menu

krispy-snacc:
Added a content rating selector for random wallpapers in the Wallpaper & Colors settings, allowing users to choose between Safe, Questionable, and Explicit ratings. The selected rating is stored as a config variable (randomWallRating) that can be reused by other random wallpaper scripts, and is currently implemented in the Konachan wallpaper fetcher to filter images based on user preference (defaults to "safe").

Preview:
image

### DIFF
--- a/dots/.config/quickshell/ii/modules/common/Config.qml
+++ b/dots/.config/quickshell/ii/modules/common/Config.qml
@@ -190,6 +190,7 @@ Singleton {
                 }
                 property string wallpaperPath: ""
                 property string thumbnailPath: ""
+                property string randomWallRating: "safe" // "safe", "questionable", "explicit"
                 property bool hideWhenFullscreen: true
                 property JsonObject parallax: JsonObject {
                     property bool vertical: false
--- a/dots/.config/quickshell/ii/modules/settings/QuickConfig.qml
+++ b/dots/.config/quickshell/ii/modules/settings/QuickConfig.qml
@@ -118,6 +118,26 @@ ContentPage {
                         text: Translation.tr("Random osu! seasonal background\nImage is saved to ~/Pictures/Wallpapers")
                     }
                 }
+                ConfigSelectionArray {
+                    currentValue: Config.options.background.randomWallRating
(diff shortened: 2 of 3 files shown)


alaotach:
+1 i needed this thankyou mf

dots/.config/quickshell/ii/modules/common/Config.qml
```diff
@@ -190,6 +190,7 @@ Singleton {
                 }
                 property string wallpaperPath: ""
                 property string thumbnailPath: ""
+                property string randomWallRating: "safe" // "safe", "questionable", "explicit"
```
end-4, line 193: This is not the place. Please have it as `wallpapers.randomWallpaperRating`

dots/.config/quickshell/ii/modules/settings/QuickConfig.qml
```diff
@@ -118,6 +118,26 @@ ContentPage {
...
+                        {
+                            "value": "questionable",
+                            "displayName": Translation.tr("Questionable")
+                        },
+                        {
+                            "value": "explicit",
+                            "displayName": Translation.tr("Explicit")
+                        },
+                    ]
+                }
```
end-4, line 140: This looks quite out of place (UI-wise, not code style). Honestly I think we shouldn't have it

end-4: I understand this is like, most of the PR and it would be extremely inconvenient without these buttons but at the same time it feels not so okay to have hentai _filling the screen_  so simply

nonetrix: I'd like if this was implemented somehow if I lived alone for laughs, but probably not the best way of going about it maybe

krispy-snacc: hmm, maybe we could just have the option available in the config only rather than in the settings menu if thats the issue

PR #2482, opened 2025-11-14
Bluetooth icon alternative

fedetrifo06:
## Description

I have added the power icon as an alternative to the bluetooth icon if bluetooth is not available.
Without a replacement only the network icon was present, making the sidebar button too small.

Before:
image

After:
image


end-4:
That power icon is weird, so I fixed the spacing instead in the above commit. Now it should look like this. It shouldn't be too small because the button on the left is smaller
image

PR #2484, opened 2025-11-15
Revert #1841

nonetrix:
## Describe your changes

This reverts #1841 since the rule is useless and doesn't fix any issues as it's a bug in the QT implementation not XDG, additionally the QT one can't be easily targeted as it doesn't have a consistent name/class. The only way I can find to fix this is using qt6ct to tell QT apps to use XDG, but this uses KDEs theming system so I have no idea how to fix this

image

Should be and uh sorry lol


nonetrix:
Nvm I'm extra stupid it happens with both 

Think before opening PRs next time maybe

PR #2489, opened 2025-11-15
Brightness fix for multiple devices

imitoy:
## Changes

Fix for multiple devices, with no-percentage. It is an alternative to [#2415](https://github.com/end-4/dots-hyprland/pull/2415)

## Test

My local machine only.


imitoy:
It seems to be conflictive with the main branch.

dots/.config/quickshell/ii/services/Brightness.qml
```diff
@@ -137,16 +150,15 @@ Singleton {
...
             } else {
-                const valuePercentNumber = Math.floor(brightnessValue * 100);
-                let valuePercent = `${valuePercentNumber}%`;
-                if (valuePercentNumber == 0) valuePercent = "1"; // Prevent fully black
-                setProc.command = ["brightnessctl", "--class", "backlight", "s", valuePercent, "--quiet"];
-                setProc.startDetached();
+                for (let i = 0; i < monitor.rawMaxBrightness.length; i++) {
+                    const rawValueRounded = Math.max(Math.floor(brightnessValue * monitor.rawMaxBrightness[i]), 1);
+                    setProc.command = ["brightnessctl", "-d", monitor.devices[i], "set", rawValueRounded, "--quiet"];
+                }
```
end-4, line 171: Wouldn't this just trigger the last command?

imitoy:
> I haven't checked the parsing yet but I'm not sure what you're trying to do in the setting

I've made it to store device name as an array and calculate brightness for each device.

Shall I add some comments?

dots/.config/quickshell/ii/services/Brightness.qml
imitoy: Yes. I'll fix this

PR #2496, opened 2025-11-16
[refactor] Updated some Fedora-related content

ririko6z:
##  Reconstructed part of the code
- Use copr repositories to replace local builds as much as possible.
- The remaining `quichshell-git` and `hyprland-qt-support` components can be effectively resolved through local builds due to compatibility issues with certain private interfaces (though this approach would consume more time).
- I adjusted the naming conventions of some software to better align with Fedora's community standards.
- Sync dist-arch updates @clsty

### DIFF
--- a/sdata/subcmd-install/2.setups.sh
+++ b/sdata/subcmd-install/2.setups.sh
@@ -28,9 +28,7 @@ showfun setup_user_group
 v setup_user_group
 
 if [[ ! -z $(systemctl --version) ]]; then
-  # TODO: Why fedora does not add i2c-dev?
-  # TODO: Why fedora add uinput and udev rules?
-  # Regarding uinput and udev rules: the former is required for the virtual keyboard to function, while the latter enables input group users to utilize it.
+  # For Fedora, uinput is required for the virtual keyboard to function, and udev rules enable input group users to utilize it.
   if [[ "$OS_GROUP_ID" == "fedora" ]]; then
     v bash -c "echo uinput | sudo tee /etc/modules-load.d/uinput.conf"
     v bash -c 'echo SUBSYSTEM==\"misc\", KERNEL==\"uinput\", MODE=\"0660\", GROUP=\"input\" | sudo tee /etc/udev/rules.d/99-uinput.rules'


sdata/subcmd-install/2.setups.sh
```diff
@@ -2,19 +2,22 @@
...
-  if [[ ! -e "${XDG_CONFIG_HOME}/systemd/user/ydotool.service" ]]; then
-    x ln -s /usr/lib/systemd/system/ydotool.service "${XDG_CONFIG_HOME}/systemd/user/ydotool.service"
+  if [[ ! -e "/usr/lib/systemd/user/ydotool.service" ]]; then
+    x ln -s /usr/lib/systemd/{system,user}/ydotool.service
   fi
 }
 
 function setup_user_group(){
-  if [[ -z $(getent group i2c) ]]; then
+  if [[ -z $(getent group i2c) ]] && [[ "$OS_GROUP_ID" != "fedora" ]]; then
```
clsty, line 11: This seems not appropriate.
- Why not just use `getent` to determine whether `i2c` exist?
- This is certainly not idempotent, as when `i2c` already exists, `sudo groupadd i2c` gives an error.

ririko6z:
Regarding i2c-dev: Its purpose is to allow direct brightness adjustment via the mouse wheel in the status bar.  However, in Fedora, I can still achieve this without adding the i2c group.  The device loads automatically without explicit declaration.  P.S. Its name should be i2c_dev in Fedora.

sdata/subcmd-install/2.setups.sh
```diff
@@ -2,19 +2,22 @@
...
     x sudo groupadd i2c
   fi
-  x sudo usermod -aG video,i2c,input "$(whoami)"
+
+  if [[ "$OS_GROUP_ID" == "fedora" ]]; then
+    # TODO: I'm not sure what the video group is for. I didn't join it, and it doesn't seem to be affecting anything. I may need to remove it later
```
clsty: If you're not sure about it, just leave it as it is, else it may potentially break something.

ririko6z:
Regarding uinput and udev rules: the former is required for the virtual keyboard to function, while the latter enables input group users to utilize it.

clsty (changes requested):

clsty:
> Regarding uinput and udev rules: the former is required for the virtual keyboard to function, while the latter enables input group users to utilize it.

It sounds reasonable. Please add this as comment near the correponding line.

sdata/dist-fedora/install-deps.sh
```diff
@@ -35,15 +34,19 @@ v sudo dnf install geoclue2 brightnessctl ddcutil -y
...
-  adw-gtk3-theme breeze-cursor-theme grub2-breeze-theme breeze-icon-theme{,-fedora} 
-  kf6-breeze-icons sddm-breeze darkly eza fish fontconfig kitty matugen starship 
-  jetbrains-mono-nl-fonts material-icons-fonts twitter-twemoji-fonts
+  adw-gtk3-theme breeze-cursor-theme grub2-breeze-theme breeze-icon-theme{,-fedora} kf6-breeze-icons
+  sddm-breeze breeze-plus-icon-theme darkly eza fish fontconfig kitty matugen florian-karsten-space-grotesk-fonts
+  starship gabarito-fonts jetbrains-mono-nerd-fonts google-material-symbols-vf-rounded-fonts material-icons-fonts
+  readex-pro-fonts-all google-roboto-flex-fonts google-rubik-vf-fonts twitter-twemoji-fonts
 )
-v sudo dnf install ${themes_deps[@]} -y
+sudo dnf install ${themes_deps[@]} -y
```
clsty: Why the removal of `v`?

sdata/subcmd-install/2.setups.sh
clsty: > However, in Fedora, I can still achieve this without adding the i2c group. The device loads automatically without explicit declaration. P.S. Its name should be i2c_dev in Fedora.

In such case, for Fedora user please just let it not joining `i2c` nor `i2c_dev`.

ririko6z: image
- Nothing seems out of the ordinary. Can you tell me what features they involve?
- It should be idempotent, so I asked the Fedora distribution not to create this group.

clsty: > Its purpose is to allow direct brightness adjustment via the mouse wheel in the status bar.

If you're using laptop this is not needed. However for monitors connected via HDMI or DP this is needed.

sdata/dist-fedora/install-deps.sh
ririko6z: 😂 I think I accidentally deleted it

sdata/subcmd-install/2.setups.sh
ririko6z: > > Its purpose is to allow direct brightness adjustment via the mouse wheel in the status bar.
> 
> If you're using laptop this is not needed. However for monitors connected via HDMI or DP this is needed.

But I'm using a desktop with an NVIDIA graphics card.

clsty: > - It should be idempotent, so I asked the Fedora distribution not to create this group.

Sorry I missed the `!` in `!=`.

> But I'm using a desktop with an NVIDIA graphics card.

In such case, we do not need to add it for Fedora.
Please as well add the comment near it like
```
# On Fedora this is not needed. Tested with desktop computer with NVIDIA video card.
```

sdata/subcmd-install/2.setups.sh
```diff
@@ -2,19 +2,23 @@
...
 function setup_user_group(){
-  if [[ -z $(getent group i2c) ]]; then
+  if [[ -z $(getent group i2c) ]] && [[ "$OS_GROUP_ID" != "fedora" ]]; then
+    # On Fedora this is not needed. Tested with desktop computer with NVIDIA video card.
     x sudo groupadd i2c
   fi
-  x sudo usermod -aG video,i2c,input "$(whoami)"
+
+  if [[ "$OS_GROUP_ID" == "fedora" ]]; then
+    # TODO: I'm not sure what the video group is for. I didn't join it, and it doesn't seem to be affecting anything. I may need to remove it later
```
clsty:
```suggestion
```

sdata/subcmd-install/2.setups.sh
```diff
@@ -27,6 +30,7 @@ v setup_user_group
 if [[ ! -z $(systemctl --version) ]]; then
   # TODO: Why fedora does not add i2c-dev?
   # TODO: Why fedora add uinput and udev rules?
+  # Regarding uinput and udev rules: the former is required for the virtual keyboard to function, while the latter enables input group users to utilize it.
```
clsty:
```suggestion
  # For Fedora, uinput is required for the virtual keyboard to function, and udev rules enable input group users to utilize it.
```

clsty:
LGTM, merging.

Thanks for the update!

PR #2504, opened 2025-11-17
fix: don't run github action on forks

MannuVilasara:
## Describe your changes

this will make sure to only run the current workflow only on the main repo, not the forks.

yes


clsty:
Sounds resonable, I'm merging this one. Thanks.

PR #2505, opened 2025-11-17
Fix darkly widget style not being applied

Sh4dowzinha:
While the install script tries to set the darkly theme, it is then overwritten by the kdeglobals file not including it. With the style included on the kdeglobals file, the theme is now applied correctly.


end-4 (approved):

end-4:
thank you

PR #2506, opened 2025-11-17
Gentoo added Kirigami dep

jwihardi:


clsty:
Thanks for the update!

Issue #2284, opened 2025-10-24
Bring back old version of update.sh


reakjra:
### Step 1. Before you submit

- [x] I have read the [Troubleshooting](https://ii.clsty.link/en/ii-qs/04troubleshooting/) and [Usage](https://ii.clsty.link/en/ii-qs/02usage/) pages.
- [x] I've successfully updated to the latest version following the [guidance](https://ii.clsty.link/en/ii-qs/01setup/#updating).
- [x] I've successfully updated the system packages to the latest.
- [ ] I've ticked the checkboxes without reading their contents

### Step 2. Quick diagnose info

!

### Step 3. Describe the issue

So genuinely, I can't figure out how to update without needing to overwrite everything? Since the `update.sh` is gone, I've read now it should be done with the `install.sh`, however it does not mention anything about updating (or at least, updating without fully replacing the files) like the old update method did. And the exp-update practically updates ONLY the latest commit. Bring back the update.sh! ( ｡ •` ⤙´• ｡)

### Reminder

- [x] I agree that it's usually impossible for others to help me without my logs.

clsty:
exp-update is update.sh literally so there is nothing we can bring back for you. Closing.

reakjra:
As I said tho, it seems it only updates for the latest commit. it does not stack up anymore like the previous one did. If it's not changed at all, I guess I'll have to figure out by myself.

reakjra:
Oh wait, I figured out it does not update the actual user config anymore, but the cloned repository one (`end4/dots/.config/quickshell`). Which doesn't make really sense to me honestly. 

Plus, now it completely replaces the file, while, I may be completely stupid but, the previous update.sh did not. So now you have to update, manually copy-paste the updated files and write everything you did on that file once again. So it's just more annoying if I'm not missing anything.

clsty:
The update.sh was introduced in #1336 and recently updated in #2220.

clsty:
I've brought back the old version of update.sh in b0d64d00b63a8695e879fcdc0ed56a48dbaffe78 but it still needs modification to work normally as a subcommand `exp-update-old`. I'll do it when I'm available.

reakjra:
Thank you a lot. Take your time with it <3

clsty:
Should be ready now. @reakjra Please give feedback.

reakjra:
I may be using it wrong but it does not work. Using `./setup exp-update-old` behaves like the normal `exp-update`. It does not touch the user config at all

[...]

PR #2510, opened 2025-11-18
Setup script subcmd-exp-merge

reakjra:
## Describe your changes

Added the `exp-merge` sub command as told in https://github.com/end-4/dots-hyprland/issues/2284. The script follows the simple logic of updating the main branch, then it checkouts a permanent exp-merge-branch where it will copy the user quickshell folder (`~/.config/quickshell`), commits those changes, then runs `git rebase main` which will try to merge the updated main with the user current config. 

I decided to not use worktrees to keep the cloned repository cleaner and safer to not mess with. Furthermore, the persistant branch will help git to perform the 3-ways merge to avoid ugly conflicts that would happen with an ephemeral branch each time. 

Review needed.


clsty:
I know only basics of Git so I can not do a complete review. The other things looks fine so I think it's ok to merge... Except for one thing: You should warn user about "this script is experimental, only continue at your own risk" (similar as `exp-update`).

If you find out more problem when using `./setup exp-merge`, you're welcomed to make futher PRs.

reakjra:
Done. As far as I tested should work fine, I will open a new PR if I find any error or someone else does

PR #2514, opened 2025-11-18
Fix file existence check in dedup_and_sort_listfile

Sh4dowzinha:
## Describe your changes

Fixed a condition check that was reversed on the function dedup_and_sort_listfile function on the function.sh file.

It's ready and tested.


clsty:
Thanks for the fix.

Issue #1664, opened 2025-07-18
Anime Boorus Not Working


nyxv3il:
### Step 1. Before you submit

- [x] I have read the [Troubleshooting](https://end-4.github.io/dots-hyprland-wiki/en/ii-qs/04troubleshooting/) and [Usage](https://end-4.github.io/dots-hyprland-wiki/en/ii-qs/02usage/) pages.
- [x] I've successfully updated to the latest version following the [guidance](https://end-4.github.io/dots-hyprland-wiki/en/ii-qs/01setup/#updating).
- [x] I've successfully updated the system packages to the latest.
- [ ] I've ticked the checkboxes without reading their contents

### Step 2. Quick diagnose info

<details><summary>Quick diagnose</summary>

```

/========================\
| Checking git repo info |
\========================/

[===diagnose===] git remote get-url origin
https://github.com/end-4/dots-hyprland.git
[---SUCCESS---]

[===diagnose===] git rev-parse HEAD
bab166baeaebbb0d68c58fb709c4cfeed95931fb
[---SUCCESS---]

/=================\
| Checking distro |
\=================/

[===diagnose===] ii_check_distro
./diagnose: line 34: lsb_release: command not found
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo
[---SUCCESS---]

/====================\
| Checking variables |
\====================/

[===diagnose===] declare -p XDG_CACHE_HOME
declare -x XDG_CACHE_HOME="/home/cyberduck/.cache"
[---SUCCESS---]

[===diagnose===] declare -p XDG_CONFIG_HOME
declare -x XDG_CONFIG_HOME="/home/cyberduck/.config"
[---SUCCESS---]

[===diagnose===] declare -p XDG_DATA_HOME
declare -x XDG_DATA_HOME="/home/cyberduck/.local/share"
[---SUCCESS---]

[===diagnose===] declare -p XDG_STATE_HOME
[...]

Issue #1534, opened 2025-06-28
Booru api configuration options


TheGreyfeather:
### What would you like to be added?

Booru API Token and/or Username options in the configuration beyond zerochan's setting of a username. 

### How will it help?

Many of the Boorus have a way to set a username and/or API Token to avoid being rate limited or to otherwise increase the limits. Having a built in way to provide these in the configuration or in the sidebar via `/key` would be a great feature to implement.
By Providing a username/API Token this can help avoid rate limiting or blacklisting through proper usage of the API services being provided.
Gelbooru is largely impacted by this, Danbooru will also allow registered users to search with more than two tags a time with a properly authorized API call

### Extra info

This is largely the result of several requests getting stuck in an unresolved Pending state with 4xx HTTP errors that would be resolved by properly authenticating with the API services

RegentsVoice:
I can confirm this, I had to make quite a few changes to the file .config/quickshell/services/Booru.qml to get gelbooru to work as expected.

TheGreyfeather:
@RegentsVoice What changes did you end making to make it work? Depending on the changes, it may be worth looking at them to make a PR. I'm willing to help make PRs, but I have no experience with QML

RegentsVoice:
@TheGreyfeather  I don't think I did it cleanly enough, but in my defense, the module structure was poorly thought out from the start.
You can simply move this code to your Booru.qml, replacing all YOUR-API-KEY and YOUR-USER-ID, this will be enough for gelbooru to work correctly.

```
pragma Singleton
pragma ComponentBehavior: Bound

import "root:/modules/common"
import Quickshell;
import Quickshell.Io;
import Qt.labs.platform
import QtQuick;

/**
 * A service for interacting with various booru APIs.
 */
Singleton {
    id: root
    property Component booruResponseDataComponent: BooruResponseData {}

    signal tagSuggestion(string query, var suggestions)

    property string failMessage: qsTr("That didn't work. Tips:\n- Check your tags and NSFW settings\n- If you don't have a tag in mind, type a page number")
    property var responses: []
    property int runningRequests: 0
    property var defaultUserAgent: Config.options?.networking?.userAgent || "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/123.0.0.0 Safari/537.36"
    property var providerList: Object.keys(providers).filter(provider => provider !== "system" && providers[provider].api)

[...]

PR #2520, opened 2025-11-19
Added command to set Gelbooru API key and user ID

nonetrix:
## Describe your changes

This adds a command to set the Gelbooru API key with a command as it requires one and doesn't work without it, the only way of working around this before was editing the Booru.qml file and adding it yourself

I made this with Antigravity IDE from Google and Gemini 3 (first try btw). However, I've looked at the code, and it seems fine to me at least, it doesn't appear all that different from anything else and indeed works. Still, I'd really like a second opinion as I don't trust my judgment whatsoever. Other Booru maybe could also benefit from the same treatment, but this PR just focuses on Gelbooru

Sorta closes (need support for other boorus): #1534 
Double closes?: #1664

### DIFF
--- a/dots/.config/quickshell/ii/modules/ii/sidebarLeft/Anime.qml
+++ b/dots/.config/quickshell/ii/modules/ii/sidebarLeft/Anime.qml
@@ -83,6 +83,28 @@ Item {
                 Persistent.states.booru.allowNsfw = true;
             }
         },
+        {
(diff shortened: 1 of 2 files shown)


nonetrix:
I hate to be that person, but could someone please review this? I've used this patch without issue for a while, I plan to add other features too like being able to view full resolution version without opening browser

PR #2533, opened 2025-11-21
Added apply to IpcHandler

EphraimShikanga:
This pull request adds a new method to the wallpaper selector module, allowing wallpapers to be applied directly from the terminal by specifying a path.

Enhancement to wallpaper selection:
* Added an `apply(path: string)` function to the `Scope` object in `WallpaperSelector.qml`, enabling wallpapers to be applied programmatically by path.


end-4:
This goes in `Wallpapers` service

end-4:
and can we even pass arguments to the IpcHandler? afaik no

end-4:
turns out we can... that's new to me. thank you and i will merge this now

PR #2536, opened 2025-11-21
feat: increase decrease sliders on scroll

MannuVilasara:
## Describe your changes

adds the feature of increasing or decreasing the volume and brightness by scroll similar to windows

yes here's the video

https://github.com/user-attachments/assets/e5fe3c5f-af69-4d0d-ba13-d6e9fbd352a0


end-4:
This should be implemented in WSlider instead because it's quite a common action. I did it in the above commit.

PR #2545, opened 2025-11-22
Fix: Flatpak app doesnt appear on app launcher

Nasrullaunais:
## Describe your changes
add flatpak paths to XDG_DATA_DIRS 

Adds standard Flatpak export paths to XDG_DATA_DIRS to ensure applications appear in the Quickshell launcher. Explicitly includes /usr/local/share and /usr/share as fallbacks to prevent system applications from disappearing if the variable was previously unset.

it works for me, pls review


end-4:
I feel this is more the user's fault for using flatpak, but the default is `/usr/local/share/:/usr/share/` so doesn't hurt I guess

PR #2548, opened 2025-11-22
Enhance The Translate Service 

ah51k:
# 1-Reverse translation
I have modified the service to make the translation reverse, instead of modifying the source and target language every time. Now all you have to do is write freely between the two chosen languages, and the service will recognize the language and translate to the opposite.
| en -> ar | ar->en |
|--------|--------|
|  image|image| 

like you see actully I wrote this massage with the service 🤗

# 2-local-Translate-System 
I have designed a system to automatically save the words used in translation daily in the service. If you cancel it, nothing will be saved in it. If the button is activated, any word you write will be saved in the service. 
```
.local/share/quickshell_translator/
```
It is in the form of a tree folder system, and if the size of the index files reaches 200BM, the system automatically cleans part of them, and if you want to delete them manually, you can through the interface, through the bottom button, delete only a specific language.
| on | off | dir-manager |
|--------|--------|--------|
|image|image|image| 
# 3-I just modified the interface arrangement a little, I think it's clear

# Is it Ready 
Yes, at least for me, I am waiting for your feedback's

### DIFF
--- a/dots/.config/quickshell/ii/modules/ii/sidebarLeft/Translator.qml
+++ b/dots/.config/quickshell/ii/modules/ii/sidebarLeft/Translator.qml
@@ -5,246 +5,313 @@ import qs.modules.common.functions
 import qs.modules.ii.sidebarLeft.translator
 import QtQuick
 import QtQuick.Layouts
+import QtQuick.Controls 2.15
 import Quickshell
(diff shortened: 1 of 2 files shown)


tokKurumi:
AI 🤖

PR #2549, opened 2025-11-23
Add setting for choosing brightnessctl device

viridivn:
## Describe your changes
I noticed that brightness control wasn't working on my hybrid laptop, and it turned out it was trying to adjust the nvidia backlight instead of the intel_backlight, which is the one that works. I've added a configuration option in settings under services to either leave it at the default, or force brightnessctl to use a specific device. It should autohide the section if there are no brightnessctl compatible devices, but I cannot test that so I would appreciate someone checking it out before merge.

I'd appreciate someone making sure the autohide functionality works on a device with no internal displays. It's also my first time writing QML so a quick review would be appreciated as well.

### DIFF
--- a/dots/.config/quickshell/ii/modules/common/Config.qml
+++ b/dots/.config/quickshell/ii/modules/common/Config.qml
@@ -360,6 +360,7 @@ Singleton {
             }
 
             property JsonObject light: JsonObject {
+                property string device: ""
                 property JsonObject night: JsonObject {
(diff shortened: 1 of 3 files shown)


viridivn:
Anything that needs to be done to get this merged?

viridivn:
I think https://github.com/end-4/dots-hyprland/pull/3623 is a cleaner implementation of this. Closing in favor.

PR #2557, opened 2025-11-23
Add Gemini 3 Pro to model list

viridivn:
## Describe your changes
Small PR. Adds Google's `gemini-3-pro-preview` model to the builtin model list.

## Is it ready? Questions/feedback needed?
Seems good. I've tested it with text and multimodal and everything works as expected. Tool calling works as well, shell commands, google search, etc.

PR #2558, opened 2025-11-23
Revert "Add Gemini 3 Pro to model list"

end-4:
Reverts end-4/dots-hyprland#2557


Azazel1124:
add a qwen plzzzzzzzzzzz

PR #2562, opened 2025-11-24
Ddcci fixes

moetayuko:
### brightness: use connector name to distinguish different monitors

One may have multiple monitors of the same model

### brightness: adjust DDCCI initialize timings

* Delay DDCCI monitor initialization until after `ddcutil detect` to
  ensure `isDdc`, `busNum`, and `ddcMonitors` are properly set.
* Run `ddcutil getvcp` sequentially to avoid failures from concurrent
  executions.

yes


end-4:
thank you

PR #2569, opened 2025-11-25
Add Firefox Nightly to Browser Shortcut

yav12:
## add firefox nightly to the list of browsers

it works fine as far as i can tell, when i press super+w it opens without problems


vaguesyntax:
I don't think you should open a PR to change a **personal** file.
The owner of this repository is already very busy with other waffle style, so there is no need to push them further.

majaahh:
> **personal** file.

Its literally not "personal"...

PicoShot:
> > **personal** file.
> 
> Its literally not "personal"...

most common browsers is already there
if you want to use firefox nightly just add bind in your custom configs, there is no need for PR.

end-4:
Honestly, I'm like, whatever, but there would be no end to adding more obscure stuff if I (keep) accept(ing) prs like this

PR #2575, opened 2025-11-25
Korean translation

jwdjk392:
## Describe your changes
Add korean translation.
While some keys are currently not translated, the main UI is almost fully translated - settings, side bars, etc

### Font check needed
Korean often needs a dedicated font on linux. While it worked well on my computer, which has korean fonts installed (```adobe-source-han-sans-kr-fonts``` and etc), I'm unsure if it will on other computers. I've set font to default on the illogical-impulse settings (Google Sans Flex)


end-4:
> While some keys are currently not translated, the main UI is almost fully translated - settings, side bars, etc

So are you going to keep working on it or this is all you're willing to do? Either is fine and there's no pressure, I just want your confirmation

> Korean often needs a dedicated font on linux

Google Sans Flex appears to not support korean: [See on Google Fonts](https://fonts.google.com/specimen/Google+Sans+Flex?preview.text=%E3%84%B1%ED%9B%84%E3%85%9C%E3%85%A0%E3%85%87%EB%82%B4%E3%85%94%E3%85%90%E3%85%90%E3%85%90). However Korean text should show up as long as you have a font supporting it installed, which is something I will leave to the user

Copilot review is a sanity check, accepting its suggestions or not is your choice. I trust a real person more than an AI.

jwdjk392:
I'm willing to keep working on this translation.
Also, I checked out the copilot review, and found out that I made some typo.

uiovasot:
it seems good. what ime are you using?

jwdjk392:
@uiovasot I'm currently using fcitx5 with fcitx5-hangul. They work pretty well with KDE, qt apps, but sometimes have issues with electron based apps like VSCode or Chrome.

end-4:
> @uiovasot I'm currently using fcitx5 with fcitx5-hangul. They work pretty well with KDE, qt apps, but sometimes have issues with electron based apps like VSCode or Chrome.

I remember correctly someone PR'd an environment variable that makes Electron apps launch in Wayland, and the dotfiles include a chrome flags file which should make Chrome work, with these flags:

```
--ozone-platform-hint=wayland
--enable-wayland-ime
```

PR #2580, opened 2025-11-26
Fix: finger print will still keep running when use password to login

wooze-pao:
## Describe your changes

Use "active" instead of "running" to check if the fingerprint module is still operating.

https://quickshell.org/docs/master/types/Quickshell.Services.Pam/PamContext/

yes


end-4:
thanks

PR #2588, opened 2025-11-27
added gentoo uninstall

jwihardi:


clsty:
Thanks for the update!

PR #2591, opened 2025-11-28
Add Qalculate as Dependency for Fedora

EisregenHaha:
fixes math

## Describe your changes

uh yeah fixes math in search

nothing changed besides that

### DIFF
--- a/sdata/dist-fedora/feddeps.toml
+++ b/sdata/dist-fedora/feddeps.toml
@@ -181,6 +181,7 @@ packages = [
   "hyprpicker",
   "songrec",
   "translate-shell",
+  "qalculate",
   "wlogout"
 ]
 


clsty:
https://github.com/end-4/dots-hyprland/blob/2fd25af353fd963c13ef35d07325b9143f3b4a48/dots/.config/quickshell/ii/modules/ii/overview/SearchWidget.qml#L135

It seems that `qalc` is used here, which is provided by `libqalculate`.

It's interesting that `dist-arch` and `dist-gentoo` are also missing that...

For Arch Linux it's [libqalculate](https://archlinux.org/packages/extra/x86_64/libqalculate/), which I'll add later (also in `deps-info.md`).

For Gentoo @jwihardi please help.

**Update:** as @EisregenHaha mentioned in the comment,
> libqalculate already gets installed as a dependency for stuff like plasma-workspace, but for whatever reason the calculations dont work for me unless i install qalc in the searchbar

So @jwihardi maybe it's already installed as dependency of another package. Just be sure that the package provides the command `qalc`. For example, `libqalculate` in Arch Linux provides `qalc`, but in Fedora `libqalculate` only provides some libs, not the command `qalc`.

sdata/dist-fedora/feddeps.toml
```diff
@@ -181,6 +181,7 @@ packages = [
   "hyprpicker",
   "songrec",
   "translate-shell",
+  "qalculate",
```
clsty, line 184: Since we only need the executable `qalc` and not Qt/Gtk frontend, why not install [libqalculate](https://packages.fedoraproject.org/pkgs/libqalculate/libqalculate/) instead?

clsty: Oh, ok the `libqalculate` does not provide `qalc`, which `qalculate` does.

- https://packages.fedoraproject.org/pkgs/libqalculate/libqalculate/fedora-43.html#files
- https://packages.fedoraproject.org/pkgs/libqalculate/qalculate/fedora-43.html#files

EisregenHaha: libqalculate already gets installed as a dependency for stuff like plasma-workspace, but for whatever reason the calculations dont work for me unless i install qalc in the searchbar

clsty:
LGTM, thanks for the fix!

PR #2593, opened 2025-11-28
feat(clock): Improve digital clock style

vaguesyntax:
## Describe your changes

With the ability to change fonts through settings we can now change the digital clock font. However some fonts look great at lighter _weights_, while others are better in **bold**. So I've added this ability, along with other improvements for the digital clock.

- Added `ConfigSlider` component to add slider to settings app. (Can also be used in different places in settings app in the future)
- Added `vertical` switch for digital clock
- Added `weight` , `width` and `size` sliders for digital clock to settings (`width` and `roundness` property is only available for some variable fonts like `Google Sans Flex`)
- And other improvements

### Settings:

image

### Possible combinations:
|  |  |
| ----------- | ----------- |
| image | image |
| image | image |
| image| image |

### And also very close recreation of Android 16 style lock screen is possible:

image

It's currently ready. But I think more additions and settings can be added to digital clock (like in the cookie clock). Let me know if you have got any issues or suggestions!

### DIFF
--- a/dots/.config/quickshell/ii/modules/settings/BackgroundConfig.qml
+++ b/dots/.config/quickshell/ii/modules/settings/BackgroundConfig.qml
@@ -53,9 +53,9 @@ ContentPage {
     }
 
     ContentSection {
+        id: settingsClock
         icon: "clock_loader_40"
         title: Translation.tr("Widget: Clock")
-        id: settingsClock
 
         function stylePresent(styleName) {
             if (!Config.options.background.widgets.clock.showOnlyWhenLocked && Config.options.background.widgets.clock.style === styleName) {
@@ -172,7 +172,7 @@ ContentPage {
         ContentSubsection {
             visible: settingsClock.digitalPresent
             title: Translation.tr("Digital clock settings")
-            tooltip: Translation.tr("Font width and roundness settings are only available for some fonts like 'Google Sans Flex'")
+            tooltip: Translation.tr("Font width and roundness settings are only available for some fonts like Google Sans Flex")
 
             ConfigRow {
                 uniform: true
@@ -196,7 +196,6 @@ ContentPage {
 
             ConfigRow {
                 uniform: true
-                
 
                 ConfigSwitch {
                     buttonIcon: "date_range"
@@ -221,7 +220,7 @@ ContentPage {
 
             MaterialTextArea {
                 Layout.fillWidth: true
-                placeholderText: Translation.tr("Font family (e.g., Google Sans Flex)")
+                placeholderText: Translation.tr("Font family")
                 text: Config.options.background.widgets.clock.digital.font.family
                 wrapMode: TextEdit.Wrap
                 onTextChanged: {
@@ -229,7 +228,6 @@ ContentPage {
                 }
             }
 
-            
             ConfigSlider {
                 text: Translation.tr("Font weight")
                 value: Config.options.background.widgets.clock.digital.font.weight


vaguesyntax:
I've updated my system and it somehow broke the weight functionality... Closing for now, I may open it later.

end-4:
haha cum

end-4:
I doubt quickshell affects the font, and my recent changes (I mean very recent, because you seem very active) shouldn't affect the font... Are you sure you didn't change the font or anything?

vaguesyntax:
> haha cum

My favourite day btw..

> I doubt quickshell affects the font, and my recent changes (I mean very recent, because you seem very active) shouldn't affect the font... Are you sure you didn't change the font or anything?

Uhh I updated my system after opening the PR and the weight property on Text (at higher `pixelSizes`) stopped working. It is updated from Qt 6.10.0 to Qt 6.10.1 and that’s the reason behind it. ~~I think it messed the text emboldening.~~

vaguesyntax:
Oh I finally found the problem..
Somehow `font.styleName` was the problem and I have to set it to `""` in order to prevent some conflicts..

end-4:
~~If we want flexibility, maybe it's best to force Roboto Flex. Then we get a ton of axes and they work all the time~~ okay maybe not, roboto flex can't have kiddy rounded ends in case it's needed

If it becomes too much we can just have presets like "fat", "serif", etc

dots/.config/quickshell/ii/modules/common/Config.qml
```diff
@@ -186,7 +186,16 @@ Singleton {
...
+                            property bool adaptiveAlignment: true
+                            property bool showDate: true
                             property bool animateChange: true
+                            property bool vertical: false
+                            property JsonObject font: JsonObject {
+                                property real weight: 350
+                                property real width: 100
+                                property real size: 90
+                                property real roundness: 0
+                            }
```
end-4, line 199: Please add a family option (also make sure you add a text field for that in the settings)
I think we should default to Google Sans Flex for the kiddy rounded ends to work ootb, but open to discussion.

dots/.config/quickshell/ii/modules/ii/background/widgets/clock/ClockText.qml
```diff
@@ -0,0 +1,20 @@
...
+    Layout.fillWidth: true
+    horizontalAlignment: root.textHorizontalAlignment
+    font {
+        family: Appearance.font.family.expressive
+        pixelSize: 20
+        weight: 350
+        variableAxes: ({})
+        styleName: ""
+    }
+    color: root.colText
```
end-4: `root.colText` is not defined in this file. Please make this get passed where ClockText is used

dots/.config/quickshell/ii/modules/ii/background/widgets/clock/DigitalClock.qml
```diff
@@ -0,0 +1,74 @@
...
+            font {
+                pixelSize: Config.options.background.widgets.clock.digital.font.size
+                weight: Config.options.background.widgets.clock.digital.font.weight
+                variableAxes: ({
+                    "wdth": Config.options.background.widgets.clock.digital.font.width,
+                    "ROND": Config.options.background.widgets.clock.digital.font.roundness
+                })
+            }
+        }
+    }
```
end-4: Don't reinvent the `Column`

vaguesyntax: What was i thinking 🤨

dots/.config/quickshell/ii/modules/common/Config.qml
vaguesyntax: I have added the option and defaulted to 'Google Sans Flex'. 

I have got an idea: maybe we could have 1-2 default sliders for font option (popular in most fonts),  then we could give users control over font variable axes (`ROND` for google sans flex, `YTAS` `YTDE` `opsz` etc. for roboto flex). and even add new sliders with the preferred variable axes.

end-4:
~~problem with 12h formats~~ fixed below
image

dots/.config/quickshell/ii/modules/common/Config.qml
end-4: only if we knew what axes are available...

end-4: it seems printing font.variableAxes gives nothing... if there's no builtin way we'll just ignore it. i don't want to go all the way and get some external tool for this

vaguesyntax: I was thinking about user entering the variable axes themselves. For example they can add ROND and the value of this variable axe to a list of objects on the config file. (We can have sliders dynamically added/removed later on). Users will be able to control all of the variable axes this way.

But we can just ignore it if we want to keep it simple.

end-4:
Man, I've been so focused on that windows clone
I'm merging this now, thank you for your work and sorry for taking so long

PR #2620, opened 2025-12-02
Make Hyprlock display wallpaper

LIghtJUNction:
This pull request updates the Hyprlock configuration to enhance the lock screen appearance by adding a custom background image and improving the visual effects.

Visual and appearance improvements:

* Added a new `$background_image` variable in `colors.conf` to specify the default wallpaper path for the lock screen background.
* Updated the `background` section in `hyprlock.conf` to use the `$background_image` variable, and enabled blur with `blur_passes = 2` for a smoother look.## Describe your changes

### DIFF
--- a/dots/.config/hypr/hyprland.conf
+++ b/dots/.config/hypr/hyprland.conf
@@ -19,6 +19,7 @@ source=custom/execs.conf
 source=custom/general.conf
 source=custom/rules.conf
 source=custom/keybinds.conf
+source=custom/colors.conf
 
 # nwg-displays support
 source=workspaces.conf


bhimio1:
You mean Hyprlock(misstyped as Hyorlock)

LIghtJUNction:
> You mean Hyprlock(misstyped as Hyorlock)

Corrected

dots/.config/hypr/hyprlock.conf
```diff
@@ -1,7 +1,10 @@
 source=~/.config/hypr/hyprlock/colors.conf
 
 background {
+    monitor =
+    path = $background_image
     color = rgba(181818FF)
+    blur_passes = 2
```
end-4: This should be 3 to be consistent with the blur of hyprland

end-4 (commented):
It was intentional that the lock screen does not use a background due to readability concerns. 
However, we can have this if it's possible to add a color overlay to the background. It's probably possible, I'm just not sure since it's been a while since I've used hyprlock.

dots/.config/hypr/hyprlock.conf
LIghtJUNction: ok

LIghtJUNction: > This should be 3 to be consistent with the blur of hyprland

It's definitely much more effective now; I can clearly see how many characters I typed when entering my password.

dots/.config/hypr/custom/colors.conf
```diff
@@ -0,0 +1 @@
+# exec = export SLURP_ARGS='-d -c FFDAD4BB -b 673B3444 -s 00000000'
```
end-4: actually, this can be removed. we no longer use slurp as the main region picker

PR #2622, opened 2025-12-03
Added Katakana on CheatSheet module

chewwallwa:
## Describe your changes

I don't want to actively study katakana, so I made this page for a cheat sheet. It's the writing that appears least often when studying Japanese, and it has more characters that look similar to each other, so it's common to need a table.

image

Just as in the case of the Ptable, it can be disabled  by commenting two lines in the file. (I didn't saw anything about Ptable in the GUI settings app in my version, that is recent).

It seems like lazy work, but it turned out cool. I didn't change anything in ElementTile.qml, and very little in Cheatsheet.qml, to keep everything compatible.

### DIFF
--- a/dots/.config/quickshell/ii/modules/common/Config.qml
+++ b/dots/.config/quickshell/ii/modules/common/Config.qml
@@ -301,6 +301,8 @@ Singleton {
                     property int key: Appearance.font.pixelSize.smaller
                     property int comment: Appearance.font.pixelSize.smaller
                 }
+                property bool showKatakana: false
+                property bool showPeriodicTable: false
             }
 
             property JsonObject conflictKiller: JsonObject {


end-4:
Honestly, I think this is too specific. It would make sense to also add hiragana and kanji if we have this and it might become too much
However I as a weeb like this... let's make this show only when the weeb policy thing is set to yes, like the anime tab in the left sidebar
Also make it go after the periodic table so it's easier to hide for the "closet" weeb policy

end-4:
The copilot review is just to make sure the characters are correct

masthierryi:
Thank you so much for the reviews. I will also change the ones suggested by copilot.

> However I as a weeb like this... let's make this show only when the weeb policy thing is set to yes, like the anime tab in the left sidebar
> Also make it go after the periodic table so it's easier to hide for the "closet" weeb policy

Really, I've disable the weeb settings and the leftbar :) so I will suggest another idea:

On the InterfaceConfig, to put buttons for each Cheat page, on a "More cheatsheets" subsection. I've already made that, but I will make it more dynamic. 

I think it would be nice to have a cheatsheet library, but Idk if the users would use it. What do you think? 

& sorry for the extra commits on other things, I'm still figuring out the PRs

masthierryi:
I have made button to show/hide the extra cheatsheets on super+I > interface > cheatsheet

- Added a dynamic loading for the tab and page on the cheatsheet.qml to a certain point (If more cheatpages are to be added, they need to be added manually there).
- Added the to the InterfaceConfig.qml a ConfigRow in a ContentSubsection for the extra shets (also manual adding for new sheets).
- Added properties to the Config.qml so the GUI works and send it to the config.json. (Also manual, my bad;( )
- Adjusted the things coplitot suggested.

image

I will try to find a way for these files to relate dynamically so that new cheatsheets can be added (I want a lot of them lol).

PR #2624, opened 2025-12-03
Loop sudo in background during installation

Perdixky:
## Describe your changes

Eliminate repeated sudo password prompts during installation.

Issue #823, opened 2024-09-19
Use only one authentication for sudo for install.sh


b4shful:
- I have read the Usage and Troubleshooting pages of the wiki: https://end-4.github.io/dots-hyprland-wiki/en/i-i/03troubleshooting/
- I have made sure that both my config and system packages are up to date
- Linux distro: Arch Linux

---
## The issue

I ran the install script to update stuff today, and I'm being prompted for a password for every single metapackage, like:
```
==> WARNING: A package has already been built, installing existing package...
==> Installing package illogical-impulse-audio with pacman -U...
[sudo] password for joel: 
```

I'm not sure if it's something I did or not. I did recently edit my `/etc/makepkg.conf` to set `BUILDDIR=/tmp/makepkg` as per https://wiki.archlinux.org/title/Makepkg#Building_from_files_in_memory

Not entirely sure if that affects anything. A full log of one metapackage install is below, but the warning and password prompt are repeated for every single one.

<details>
<summary>Logs</summary>

```
[./install.sh]: Next command:
install-local-pkgbuild ./arch-packages/illogical-impulse-audio --needed
Execute? 
  y = Yes
  e = Exit now
  s = Skip this command (NOT recommended - your setup might not work correctly)
  yesforall = Yes and don't ask again; NOT recommended unless you really sure
====> y
OK, executing...
~/.cache/dots-hyprland/arch-packages/illogical-impulse-audio ~/.cache/dots-hyprland
[./install.sh]: Command "pushd ./arch-packages/illogical-impulse-audio" finished.
Sync Dependency (5): swww-0.9.5-2, pavucontrol-1:6.1-1, wireplumber-0.5.6-1, libdbusmenu-gtk3-16.04.0.r498-2, playerctl-2.4.1-4
warning: playerctl-2.4.1-4 is up to date -- skipping
warning: swww-0.9.5-2 is up to date -- skipping
warning: pavucontrol-1:6.1-1 is up to date -- skipping
warning: wireplumber-0.5.6-1 is up to date -- skipping
warning: libdbusmenu-gtk3-16.04.0.r498-2 is up to date -- skipping
 there is nothing to do
[./install.sh]: Command "yay -S --needed --asdeps pavucontrol wireplumber libdbusmenu-gtk3 playerctl swww" finished.
==> WARNING: A package has already been built, installing existing package...
==> Installing package illogical-impulse-audio with pacman -U...
[sudo] password for joel: 
```

</details>

[...]

PR #2630, opened 2025-12-03
Set PACMAN_AUTH to prevent repeated sudo password prompts

tmcnulty387:
## Describe your changes

- Default `PACMAN_AUTH` to `sudo` in the Arch deps script so makepkg skips its `sudo -k` path and reuses the sudo session kept alive by commit end-4/dots-hyprland@8a9f105, eliminating repeated password prompts (#2624).
- Only set `PACMAN_AUTH` when it was unset, preserving any user-defined value.

Yes, it is ready.

Tested by running `./setup install` and selecting the option to execute all commands automatically. I only needed to input my password once. Without my change, I needed to input my password multiple times.

### DIFF
--- a/sdata/dist-arch/install-deps.sh
+++ b/sdata/dist-arch/install-deps.sh
@@ -43,6 +43,11 @@ if ! command -v pacman >/dev/null 2>&1; then
   exit 1
 fi
 
+# Keep makepkg from resetting sudo credentials
+if [[ -z "${PACMAN_AUTH:-}" ]]; then
+  export PACMAN_AUTH="sudo"
+fi
+
 showfun remove_deprecated_dependencies
 v remove_deprecated_dependencies
 


clsty:
I think it will override user config if `PACMAN_AUTH` is defined in `makepkg.conf` (?)

tmcnulty387:
No, I think `makepkg.conf` is sourced everytime `makepkg` is ran. Tested it by copying `/etc/makepkg.conf` to `~/.makepkg.conf` and modifying it so PACMAN_AUTH is defined as
```
PACMAN_AUTH=("echo" "makepkg.conf" "works")
```
Then I ran install with my version of the repo and I got
```
==> Installing package illogical-impulse-audio with pacman -U...
makepkg.conf works /usr/bin/pacman --noconfirm -U /home/bruv/Code/dots-hyprland/sdata/dist-arch/illogical-impulse-audio/illogical-impulse-audio-1.0-2-any.pkg.tar.zst
[./setup]: Command "makepkg -Afsi --noconfirm" finished.
```

clsty:
I'm merging this one. Thanks for contribution.

clsty:
#823 should be solved now.

PR #2638, opened 2025-12-04
Fix. top bar time&battery text not centered

ItsSunshineXD:
## Describe your changes

the time text isn't centered. lowering it by 1px prob looks neater :3
btw the commit description is generated by github copilot which is not accurate

yes

### DIFF
--- a/dots/.config/quickshell/ii/modules/ii/bar/BatteryIndicator.qml
+++ b/dots/.config/quickshell/ii/modules/ii/bar/BatteryIndicator.qml
@@ -47,6 +47,7 @@ MouseArea {
                     Layout.alignment: Qt.AlignVCenter
                     font: batteryProgress.font
                     text: batteryProgress.text
+                    transform: Translate { y: 1 }
                 }
             }
         }


ItsSunshineXD:
i fixed battery indicator text as well

ItsSunshineXD:
btw. the dot in the siderbarRight calendar marking 'not current month' makes sense, but it looks kinda off
Relevant code:
quickshell/ii/modules/ii/sidebarRight/calendar/CalendarWidget.qml:50

ItsSunshineXD:
western and Eastern date/time formats vary. Eastern ones tend to favor styles like these:
```QML
    property string time: Qt.locale().toString(clock.date, "hh:mm")
    property string shortDate: Qt.locale().toString(clock.date, "MM/dd")
    property string date: Qt.locale().toString(clock.date, "yyyy/MM/dd")
    property string longDate: Qt.locale().toString(clock.date, "MM/dd dddd")
    property string collapsedCalendarFormat: Qt.locale().toString(clock.date, "yyyy MMMM dd")
```
i've tried implementing a time/date format setting, but unable to locate the config initializes Config.options.time.format.

end-4:
This way the specifically adjusted numbers misalign with the other text. I made the battery indicator look slightly more accurate by making it shift down in case of an odd pill height in 1a3cc9b4d33445096d62a4d831e11882d05b0681

end-4:
> western and Eastern date/time formats vary. Eastern ones tend to favor styles like these:

This is why we should not make specific adjustments like these

My fix above is somewhat acceptable because it only works around rounding (and it's certain that the affected content is just numbers)

PR #2640, opened 2025-12-04
audio: protection: fix weird volume limit behaviors

F-fengzi:
## Describe your changes
Small fixes on the volume limit feature:
1. newVolume is often a weird non-whole float number, making e.g. 1.0002... detect as exceeded the limit of 1, so I rounded it to the hundredth in the check
2. If it goes over, just set it to the limit value instead of the last volume, which fits (I think) most people's intuition better

| Before | After |
|--------|--------|
| image | image (1) |
|  | image (2) |

Yes, although I found a [previous commit on the line of my second change](https://github.com/end-4/dots-hyprland/commit/92051b88ec3e7778145270913eb88c43a817ecb9). Doesn't make sense to me, but was it intentional?


end-4:
thanks

end-4:
whoops, i slipped my finger onto the shift key there... ctrl+shift+enter turns out to close

Issue #2646, opened 2025-12-05
Implement graceful shutdown


nonetrix:
### What would you like to be added?

A graceful shutdown that waits for programs to close like on Windows, this could be done via Quickshell likely or using [Hyprshutdown](https://github.com/hyprwm/hyprshutdown). Also, I'm sure some just want everything to just forcefully shutdown, so a setting for this would be great as well

### How will it help?

This would allow programs to save data before shutdown and programs like Brave or Firefox won't complain about being shutdown incorrectly on next use

ghost:
Made a quick PR for this suggestion.

PR #2647, opened 2025-12-05
Hyprshutdown before shutting down.

ghost:
## Describe your changes

This simply adds hyprshutdown to any session management buttons. Adds #2646 

Yes, could be styled, but it uses hyprtoolkit

### DIFF
--- a/dots/.config/quickshell/ii/modules/common/functions/Session.qml
+++ b/dots/.config/quickshell/ii/modules/common/functions/Session.qml
@@ -22,7 +22,7 @@ Singleton {
 
     function logout() {
         closeAllWindows();
-        Quickshell.execDetached(["hyprshutdown", "&&, "pkill", "-i", "Hyprland"]);
+        Quickshell.execDetached(["hyprshutdown", "&&", "pkill", "-i", "Hyprland"]);
     }
 
     function launchTaskManager() {
--- a/dots/.config/quickshell/ii/modules/common/functions/Session.qml
+++ b/dots/.config/quickshell/ii/modules/common/functions/Session.qml
@@ -17,12 +17,12 @@ Singleton {
     }
 
     function suspend() {
-        Quickshell.execDetached(["bash", "-c", "systemctl suspend || loginctl suspend"]);
+        Quickshell.execDetached(["hyprshutdown", "bash", "-c", "systemctl suspend || loginctl suspend"]);
     }
 
     function logout() {
         closeAllWindows();
-        Quickshell.execDetached(["pkill", "-i", "Hyprland"]);
+        Quickshell.execDetached(["hyprshutdown", "&&, "pkill", "-i", "Hyprland"]);
     }
 
     function launchTaskManager() {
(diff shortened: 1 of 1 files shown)


nonetrix:
It's actually not QT it's Hyprtoolkit which exists now apparently:P 

https://github.com/hyprwm/hyprshutdown/blob/main/CMakeLists.txt 

Might worth remaking it in Quickshell eventually later, and just using this as a fallback in the future. Theming Hyprtoolkit would also help in another PR if it doesn't already. Finally, probably want to add it to install script and replace default physical power button behavior

ghost:
> It's actually not QT it's Hyprtoolkit which exists now apparently:P
> 
> https://github.com/hyprwm/hyprshutdown/blob/main/CMakeLists.txt
> 
> Might worth remaking it in Quickshell eventually later, and just using this as a fallback in the future. Theming Hyprtoolkit would also help in another PR if it doesn't already. Finally, probably want to add it to install script and replace default physical power button behavior

A script that would replace the physical power button would require the modification of systemd. I have done it on my laptop (because i keep hitting the power button so it shows nyan cat instead). However, every power button is slightly different and could cause issues.

dots/.config/quickshell/ii/modules/common/functions/Session.qml
```diff
@@ -17,34 +17,34 @@ Singleton {
...
 
     function suspend() {
-        Quickshell.execDetached(["bash", "-c", "systemctl suspend || loginctl suspend"]);
+        Quickshell.execDetached(["hyprshutdown", "bash", "-c", "systemctl suspend || loginctl suspend"]);
     }
 
     function logout() {
         closeAllWindows();
-        Quickshell.execDetached(["pkill", "-i", "Hyprland"]);
+        Quickshell.execDetached(["hyprshutdown", "&&, "pkill", "-i", "Hyprland"]);
```
masthierryi: `        Quickshell.execDetached(["hyprshutdown", "&&", "pkill", "-i", "Hyprland"]);`
you forgot a `"` after `&&`

ghost: 💀. It was midnight, leave me alone

nonetrix:
> > It's actually not QT it's Hyprtoolkit which exists now apparently:P
> > https://github.com/hyprwm/hyprshutdown/blob/main/CMakeLists.txt
> > Might worth remaking it in Quickshell eventually later, and just using this as a fallback in the future. Theming Hyprtoolkit would also help in another PR if it doesn't already. Finally, probably want to add it to install script and replace default physical power button behavior
> 
> A script that would replace the physical power button would require the modification of systemd. I have done it on my laptop (because i keep hitting the power button so it shows nyan cat instead). However, every power button is slightly different and could cause issues.

I mean KDE does it without anything crazy unless I'm missing something?

ghost:
> > > It's actually not QT it's Hyprtoolkit which exists now apparently:P
> > > https://github.com/hyprwm/hyprshutdown/blob/main/CMakeLists.txt
> > > Might worth remaking it in Quickshell eventually later, and just using this as a fallback in the future. Theming Hyprtoolkit would also help in another PR if it doesn't already. Finally, probably want to add it to install script and replace default physical power button behavior
> > 
> > 
> > A script that would replace the physical power button would require the modification of systemd. I have done it on my laptop (because i keep hitting the power button so it shows nyan cat instead). However, every power button is slightly different and could cause issues.
> 
> I mean KDE does it without anything crazy unless I'm missing something?

They do. You have to modify systemd configuration and disable the power button. I don't think a script that installs a desktop environment should be modifying a file like that, if it fails, nothing loads.

nonetrix:
> > > > It's actually not QT it's Hyprtoolkit which exists now apparently:P
> > > > https://github.com/hyprwm/hyprshutdown/blob/main/CMakeLists.txt
> > > > Might worth remaking it in Quickshell eventually later, and just using this as a fallback in the future. Theming Hyprtoolkit would also help in another PR if it doesn't already. Finally, probably want to add it to install script and replace default physical power button behavior
> > > 
> > > 
> > > A script that would replace the physical power button would require the modification of systemd. I have done it on my laptop (because i keep hitting the power button so it shows nyan cat instead). However, every power button is slightly different and could cause issues.
> > 
> > 
> > I mean KDE does it without anything crazy unless I'm missing something?
> 
> They do. You have to modify systemd configuration and disable the power button. I don't think a script that installs a desktop environment should be modifying a file like that, if it fails, nothing loads.

Ah I see sorry misunderstood

PR #2648, opened 2025-12-05
Fix(thumbnail): Make thumbnail generator script work

vaguesyntax:
## Describe your changes

- Fixed `thumbgen.py` using wrong gnome version causing script failure.
- Added the `GIO_USE_VFS` environment variable to fix a startup timeout bug. This reduces script startup time from 10-15 seconds to instant.
```
(process:199643): GVFS-WARNING **: 23:40:51.021: Error: Error calling StartServiceByName for org.gtk.vfs.Metadata: Timeout was reached
```

- Added thumbnail loading indicator:

https://github.com/user-attachments/assets/092f5f62-84a9-4feb-b35a-0034549bd2cb

Yep, it's ready. I hope there will be no more GH issues or discord posts after this.

### DIFF
--- a/dots/.config/quickshell/ii/modules/ii/wallpaperSelector/WallpaperDirectoryItem.qml
+++ b/dots/.config/quickshell/ii/modules/ii/wallpaperSelector/WallpaperDirectoryItem.qml
@@ -11,7 +11,6 @@ MouseArea {
     required property var fileModelData
     property bool isDirectory: fileModelData.fileIsDir
     property bool useThumbnail: Images.isValidImageByName(fileModelData.fileName)
-    property bool showLoadingIndicator: false
 
     property alias colBackground: background.color
     property alias colText: wallpaperItemName.color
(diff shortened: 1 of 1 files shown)


Ahorts:
Are there some dependencies which are required for the thumbnail generation? I haven't had any issue for a long while but I just reinstalled my system (arch)  with the dotfiles and suddenly no image thumbnails are being generated. Tried the fix here too. Doesn't seem to be loading anything.

vaguesyntax:
> Are there some dependencies which are required for the thumbnail generation? I haven't had any issue for a long while but I just reinstalled my system (arch) with the dotfiles and suddenly no image thumbnails are being generated. Tried the fix here too. Doesn't seem to be loading anything.

There shouldn't be an **extra** dependency. And if one is required for thumbnail generation, it is probably managed by the install scripts

Can you run `./thumbgen-venv.sh --img_dirs /home/vaguesyntax/Pictures/Wallpapers/` (with your own wallpaper path) in `/.config/quickshell/ii/scripts/thumbnails` and look what terminal says.


### Thumbnail scripts
(maybe you can troubleshoot it)

There is two scripts for thumbnail generation. One `generate-thumbanils-magick.sh` and another one (fallback) is `thumbgen.py`.

Second one runs when first script fails (and it mostly fails or gets skipped idk). _I have fixed `thumben.py` scripts in this PR bc it was using wrong version._

Ahorts:
Got the image thumbnails working. It was missing some gnome variable so I had to use `sudo pacman -S gnome-desktop` to install it. However I had to reduce the version from 4 to 3 for it work. Video thumbnails not working though

Ahorts:
@vaguesyntax Fixed the video wallpaper thumbnail generation with `sudo pacman -S ffmpegthumbnailer gstreamer gst-plugins-ugly   ` and deleting the cached thumbnails. Then I run the generation script again. Thanks

PR #2658, opened 2025-12-08
Feature (Overlay): Buttons toggler

reakjra:
## Describe your changes

Just makes it possible to change positions and disable/enable buttons in the overlay's bar.

Of course, now to add a overlay widget it should be mentioned in the `widgetSymbols` object plus added in the default array in `Config.qml`.

image
image

Yes.

### DIFF
--- a/dots/.config/quickshell/ii/modules/ii/overlay/OverlayTaskbar.qml
+++ b/dots/.config/quickshell/ii/modules/ii/overlay/OverlayTaskbar.qml
@@ -167,20 +167,18 @@ Rectangle {
     }
 
     component BatteryWidget: Row {
+        id: batteryWidget
         Layout.alignment: Qt.AlignVCenter
         Layout.leftMargin: 6
         Layout.rightMargin: 6
         spacing: 2
-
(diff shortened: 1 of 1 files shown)


end-4:
menu needed, something like this

image

reakjra:
how do you want it? A separate whole button for it? An expanding arrow?  a symbol (like next the clock or idk)?

end-4:
The menu needs more padding
image

The menu moving as buttons getting toggled is also bad UX. We should somehow make it not move when it's open

reakjra:
Huh, I have no idea honestly. I thought about fixing the left side, but ignoring the burden of making it centered, if someone adds more widgets it'll overflow to the right which is obnoxious. Maybe moving it into OverlayContent.qml and give it a fixed position at the center under the taskbar? Yeah it won't stay/follow under the gear but it won't move ig. Lemme know

reakjra:
ig?

https://github.com/user-attachments/assets/5c8e2707-3e73-461d-9b4c-27c990fcf0ed

PR #2659, opened 2025-12-08
Add persistent accent color

altrup:
## Describe your changes

Adds default off persistent accent color (in default config file next to palette type)
- Using `/accentcolor` automatically sets the color in the config file
- Using `/accentcolor clear` or clicking escape while selecting a color will unset it 

Pretty much done, although haven't edited the settings page at all


end-4:
thanks

PR #2663, opened 2025-12-09
Allow typing with osk into search

altrup:
## Describe your changes

- On Screen Keyboard and Task bar can now be interacted with even when search is open
- Allows you to type into the search bar with the on screen keyboard

I believe it should be working, although I haven't tested much with a second monitor
There could be a better way of managing keyboard focus that I don't know of though


altrup:
Closing because feature got added

PR #2674, opened 2025-12-11
Fix. Topbar Clock Widget text not centered

ItsSunshineXD:
## Describe your changes
The time text isn't centered because of its larger font size, but still sharing the same baseline as the surrounding text, causing vertical misalignment.

yes i've fully tested


ItsSunshineXD:
hmm i set up the window scale to 1.5x in hyprland.conf, resulting the issue. i'll close the pr instead

PR #2680, opened 2025-12-12
Added Notes

tnakirhs:
Added a new Notes tab to "cheatsheet" that allows creating and managing personal notes.
- It includes support for encrypted files .
- Basic CRUD operations have been implemented.

image
image

Let me know if anything should be improved or adjusted.

### DIFF
--- a/dots/.config/quickshell/ii/modules/ii/cheatsheet/NotesBrowser.qml
+++ b/dots/.config/quickshell/ii/modules/ii/cheatsheet/NotesBrowser.qml
@@ -1,0 +1,291 @@
+import QtQuick
+import QtQuick.Controls
+import QtQuick.Layouts
+import Qt5Compat.GraphicalEffects
+import qs.modules.common
+import qs.modules.common.widgets
+
+Item {
+    id: root
+    
+    property ListModel model
+    property string currentFilter: "All"
+    property bool isRefreshing: false
+    
+    signal filterChanged(string mode)
+    signal noteOpened(string path, bool isEncrypted)
+    signal noteFlipped(int index, string path)
+    signal createRequested()
+    signal refreshRequested()
+
+    property var tileColors: [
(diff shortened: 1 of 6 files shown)


carabistouflette:
that looks sooooo coooool. Does it respect theming?

dots/.config/quickshell/ii/modules/ii/cheatsheet/NotesBrowser.qml
masthierryi, line 1: I think you should use more the color properties on `~/.config/quickshell/ii/modules/common/Appearance.qml`, so it will match the color scheme. For `NotesBrowser.qml` and `NotesEditor.qml` files.

tnakirhs:
@carabistouflette @masthierryi The tile colors are fixed, I wanted light colors on dark background, some elements do follow scheme like modals.

vaguesyntax:
I don't think we need another notes section in this dotfiles as we currently have it in the overlay.

And also I don't think we need such a complex notes section. Shell's purpose is not to be able to do everything (especially when  **apps** exists for it).
(There will be no end to this and someone will eventually **vibe code** a huge IDE for the shell).

tnakirhs:
@vaguesyntax I didn't know if we had one. where is it in the overlay?

vaguesyntax:
> @vaguesyntax I didn't know if we had one. where is it in the overlay?

`Super + G` opens the overlay menu

image

Toggle notes from here 

image

tnakirhs:
@vaguesyntax I wanted a notebook-style setup, had no idea Notes existed tbh and its more like a todo list anyway.

end-4:
I'm not really against file read/writing/encryption, but any addition should be made to the notes widget in the overlay

Simplicity is the goal as it's not supposed to be a full blown text editor. That's not the main responsibility of the shell.

PR #2682, opened 2025-12-12
Add dismiss button to notification groups and refactor header buttons into a new container

carabistouflette:
Just added a little dismiss button to groups of notification because working with specific apps which sends a lot of notification can be wonky when you don't want to erase every notification

Yes, maybe some feedback about the code. It's actually simple code but i'm new to quickshell.


carabistouflette:
image
image

end-4:
But you can already swipe it, like individual notifications

https://github.com/user-attachments/assets/6eea2451-b63f-46bc-a2a3-cbd862c83ac6

carabistouflette:
Yes but having a button is better when you have a trackpad. With a trackpad swiping is sometimes a bit wonky especially when you have a bad trackpad lol

carabistouflette:
@end-4 i don't really mind if the feature isn't merged. If you think that's not a good idea let's just close the PR

Zatch07:
I think its a good feature i dont want to flick my hand twice once to get there and then to close it
this makes it simple and in terms with how android does it
and we can also have the swipe as a secondary

PR #2685, opened 2025-12-13
Use rm -rf for removing paths in uninstall script

feytrace:
... most of these are directories

## Describe your changes

changed to rm -rf instead of rm, tested on system. otherwise gave the error of "rm: cannot remove $file: Is a directory". might be a random error, just creating this pull request 

probably ready, tested on two machines, and it worked great. without it, it would error out. Feedback needed, implementation maybe not needed

### DIFF
--- a/sdata/subcmd-uninstall/0.run.sh
+++ b/sdata/subcmd-uninstall/0.run.sh
@@ -45,7 +45,7 @@ function delete_targets(){
       printf "${STY_YELLOW}Target \"$path\" inexists, skipping...${STY_RST}\n"
       continue
     elif [[ "$path" == "$HOME"* ]]; then
-      x rm -- "$path"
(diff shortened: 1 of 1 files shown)


feytrace:
might be safer to do something like 

```
x rm -r -- "$path" || x rm -rf -- "$path"
```

or 

```
if [ -d "$path" ]; then
  rm -r -- "$path"
elif [ -e "$path" ]; then
  rm -- "$path"
else
  echo "Warning: path does not exist: $path" >&2
fi
```

clsty:
`-rf` should be avoided. Closing.

reakjra:
@clsty consider reworking the uninstall script slightly, if you don't want to use use `-r`/`-rf` for folders, maybe a better-looking prompt to guide the user to do so? 'cause currently the user has to do it manually anyway

PR #2687, opened 2025-12-13
Disable custom config sources in hyprland.conf

MysticalMike60t:
Commented out custom configuration sources in hyprland.conf since they produce globbing error if custom folder isn't created correctly (doesn't contain the files).

If someone is configuring far enough to need custom files, they can also uncomment each source line. But for people who don't care, it will reduce possible errors :)

## Describe your changes

Commented out custom configuration sources in hyprland.conf since they produce globbing error if custom folder isn't created correctly (doesn't contain the files).

Yes, it is ready.

### DIFF
--- a/dots/.config/hypr/hyprland.conf
+++ b/dots/.config/hypr/hyprland.conf
@@ -14,11 +14,11 @@ source=hyprland/colors.conf
 source=hyprland/keybinds.conf
 
 # Custom 
-source=custom/env.conf
-source=custom/execs.conf
-source=custom/general.conf
-source=custom/rules.conf
-source=custom/keybinds.conf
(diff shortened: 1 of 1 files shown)


clsty:
>  if custom folder isn't created correctly (doesn't contain the files).

Custom folder should be created correctly. If this is the case then it should be dealt with from the first place.

Closing.

MysticalMike60t:
Not necessary to fix an issue only present in some installs. It is better to just comment out sourcing of CUSTOM files that not everyone uses. Anyone who is modifying will read it, and see that they need to uncomment, while anyone who wants to immediately use it can just use it without getting errors no matter what issues arise with the custom folder.

PR #2700, opened 2025-12-16
Gentoo minor modifications

jwihardi:
- I removed the initial install-deps.sh message because GCC 15 no longer requires keywording now that it is stable. Clang 19 can also be used as an alternative, and is likewise stable.
- Added ``hyprland-qtutils-private.patch``, I'm not sure if everyone needs this, but I did. The install script doesn't handle this; it's something the user would need to copy over if they needed the patch. I think it's because of using QT 6.10.1 instead of 6.9.3, but I may be wrong.
- Updated the ``dist-gentoo`` README


clsty:
Thanks for the update!

PR #2709, opened 2025-12-18
Added Live Lyrics 

SMANahian:
## Describe your changes

Shows time-synced lyrics for the currently playing track in the bar's media widget, fetched from [lrclib.net](https://lrclib.net). The current line sits in the middle with the previous and next lines dimmed above and below, scrolling as the song plays.

**Disabled by default.** Toggle it from *Settings → Bar → Lyrics*, or with the button in the media widget. While it's off, nothing extra is loaded and no requests are made — the selector popup is behind a `LazyLoader` and the service only fetches when `bar.media.showLyrics` is on and a track is playing.

lrclib usually has several entries for the same song and the automatic pick isn't always the right one, so left-clicking the lyrics opens a selector listing the candidates with a live preview of where each one is in the track. The choice is remembered per track in the config. Other mouse buttons keep their existing behaviour (play/pause, next, previous) and fall through to the media controls as before.

Inspiration from https://github.com/LabyStudio/plasmoid-spotify

### Files

| File | What changed |
|---|---|
| `services/LyricsService.qml` | New. Fetches, scores, parses and syncs lyrics. |
| `modules/ii/bar/LyricsSelector.qml` | New. The match-picker popup, a `LazyLoader` alongside `ResourcesPopup`/`BatteryPopup`. |
| `modules/ii/bar/Media.qml` | Renders the lyric lines, hosts the selector, adds the toggle button. |
| `modules/ii/bar/BarContent.qml` | Widens the center-left group when lyrics are on. |
| `modules/settings/BarConfig.qml` | Enable switch. |
| `modules/common/Config.qml`, `GlobalStates.qml` | Config option and open-state. |

### Since the last review

This has been rebased onto current `main` and squashed to a single commit. It was previously 19 commits at +932/−47; it is now **+630/−10** with the same functionality — mainly by replacing a four-attempt fetch cascade with a single request, dropping some layout churn, and folding the "use automatic match" button into the same list delegate as the candidates (it now sits at the top of the list rather than the bottom).

All five review comments are addressed — see the replies on each thread. In short:

- **curl timeout** — now `--connect-timeout 5 --max-time 10` plus `-f` and a non-zero-exit handler, so a stalled network surfaces "Couldn't reach lrclib" instead of hanging on "Fetching lyrics…". Verified against an unroutable host.
- **Untranslated strings** — all user-facing text now goes through `Translation.tr`.
- **Negative popup radius** — clamped with `Math.max(0, …)`.
- **`ActiveWindow` wrapper reserving space** — the wrapper `Item` is gone; `ActiveWindow` gets `clip: true` and `Layout.preferredWidth: 0` directly, so it yields to the center section instead of leaving a gap.
- **`QsWindow` attached property** — this one is a false positive, `QsWindow` is provided by Quickshell and is used in ~15 places in this repo already (`StyledPopup.qml:38`, `BarContent.qml:15`, `Workspaces.qml:19`). The restructure removed that call site anyway.

@BharathBala21 there's now an enable switch under *Settings → Bar → Lyrics*. I left width/appearance knobs out to keep this PR to one feature — happy to follow up with those separately.

Ready.

**Tested** on Arch, Hyprland, Quickshell 0.2.1 / Qt 6.11.1, single 1920x1200 monitor, horizontal top bar, Spotify via MPRIS: line rendering and sync, the selector opening and anchoring under the lyrics, switching to a different match, the choice surviving a shell restart, toggling on/off (bar expands/collapses, `Resources` returns, no `ActiveWindow` gap), and the network-failure path. No new QML warnings versus a clean `main`.

**Not tested** — I don't have the hardware/config: multi-monitor selector placement (it is gated on the screen whose bar was clicked, so it should open only there) and the bottom-bar layout.

Lyrics are horizontal-bar only — `VerticalMedia.qml` is untouched, so nothing appears on a vertical bar and the selector has no vertical case to handle.

One note for anyone who ran the earlier version of this branch: `bar.media.lyricsSelection` changed from an object to a list. It was declared as a `JsonObject` with dynamic keys before, which silently never persisted anything. Since `bar.media` is new in this PR, no existing user is affected.


BharathBala21:
Working
can you also add settings to customize it via shell?like width and stuff

PR #2710, opened 2025-12-18
update dist-fedora

ririko6z:
- Removed some unnecessary packages.
- Adjust the `install-deps.sh` script to make it easier for humans to understand.
- <del>No content was actually updated; this was just done to prevent it from being marked as outdated .</del>

### DIFF
--- a/sdata/dist-fedora/install-deps.sh
+++ b/sdata/dist-fedora/install-deps.sh
@@ -1,6 +1,5 @@
 # This script is meant to be sourced.
 # It's not for directly running.
-set -ex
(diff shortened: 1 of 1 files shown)


sdata/dist-fedora/install-deps.sh
```diff
@@ -1,9 +1,16 @@
 # This script is meant to be sourced.
 # It's not for directly running.
+set -ex
```
clsty:
```suggestion
```

clsty:
Thanks for the update.

PR #2717, opened 2025-12-20
Add ApplicationDrawer and improve overview scrolling

hotamago:
## Describe your changes

- Introduces ApplicationDrawer.qml for a grid-based application drawer with search, context menu, and properties dialog. Refactors Overview.qml to use a flickable layout, enabling smooth expansion/collapse of the drawer based on scroll position and user interaction, and improves UI transitions and state management for the overview panel.
- Using scroll mouse to top to open full application list

## Sample Screenshots

image
image

### DIFF
--- a/dots/.config/quickshell/ii/modules/ii/overview/ApplicationDrawer.qml
+++ b/dots/.config/quickshell/ii/modules/ii/overview/ApplicationDrawer.qml
@@ -1,0 +1,657 @@
+import qs
+import qs.services
+import qs.modules.common
+import qs.modules.common.widgets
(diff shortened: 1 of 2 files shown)


0blivi0nis:
Would be nice to have this available though a Global state like GlobalStates.appDrawerOpen for user who dont want it to open using the overview. That way it can be called from the dock apps button as well and only open this app drawer

Gezraf:
In line 304, for those encountering a quickshell load error regarding the panel window (workspace management window):

Change:
``panelWindow: root``

To:
``screen: panelWindow.screen``

PR #2730, opened 2025-12-22
Fix background clock widget's color being obnoxious

reakjra:
## Describe your changes

Fixes the hideous clock widget's color when draggable.

**Before:**
swappy-20251222_232839

**Now:**

https://github.com/user-attachments/assets/8e3b2551-bf80-4bd9-87cb-1c913885394b


end-4:
I'm not sure how to reproduce your problem but the fix makes sense

Issue #2750, opened 2025-12-26
timer in topbar


Nareshix:
### What would you like to be added?

A timer that countdowns live in the topbar would be useful. Can stay in the same window while looking at the timer

PR #2757, opened 2025-12-27
Stopwatch and Pomodoro on top bar

masthierryi:
## Description

A dynamic top bar widget that displays Pomodoro and Stopwatch states with real-time updates and dynamic resizing.

- If both are running: Shows both widgets side-by-side.
- If Stopwatch is running: Shows Stopwatch only (prioritized).
- If Pomodoro is active: Shows Pomodoro (unless Stopwatch is running).
- If Paused: Keeps the widget visible if either timer has stored time (> 0).
- If Idle: Widget collapses completely (width: 0).

image

Solves #2750 
 
Seems so

### DIFF
--- a/dots/.config/quickshell/ii/modules/ii/bar/BarContent.qml
+++ b/dots/.config/quickshell/ii/modules/ii/bar/BarContent.qml
@@ -315,6 +315,8 @@ Item { // Bar content region
                 }
             }
 
+            TimersTray {}
+
             SysTray {
                 visible: root.useShortenedForm === 0
                 Layout.fillWidth: false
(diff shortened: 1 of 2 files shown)


BISWA63:
it works nice but I think that when you click on the timer, it should open the Timer tab in right sidebar, it might be kinda tough to implement but that would be a really nice addition

masthierryi:
Bruh, sorry, I'm not using ii now. I've broke it too much :(
But I think this should be easy. Maybe joining the trigger from hover/cilck on the bar, and the trigger from choosing the timer tab. I will take a look, but I cannot commit without testing.

therapisttt:
was looking for this feature, works great thank you

PR #2771, opened 2025-12-29
fix(ai): add Gemini 2.5 & 3 Flash support and improve AI sidebar stability

JainamOswal18:
## Describe your changes

Fixes AI sidebar function calling errors with Gemini 2.5 & 3 Flash Preview models:

**Core fixes:**
- Added `thoughtSignature` property to `AiMessageData` to store reasoning metadata
- Modified `GeminiApiStrategy` to capture and echo `thought_signature` in function call history
- Added validation to skip malformed function calls with missing names or empty arguments

**Improvements:**
- Enhanced system prompt with tool/function execution guidelines for more direct responses
- Fixed `MessageTextBlock` property access to use local `done` property instead of `messageData.done`

This resolves **"Error 400: Function call is missing a thought_signature"** when using Gemini 3 Flash Preview with local tools (run_shell_command, get_shell_config, etc.). The model requires preserving and echoing the thought_signature field for multi-turn function calling.

preview (1)

The MessageTextBlock fix prevents undefined property errors when messageData is initialized as an empty object.

✅ Ready for review. Tested locally with Gemini 2.5 & 3 Flash Preview using function calling features (shell commands, config management). All function calls now complete successfully with proper thought signature handling.

image


gabectlz:
what's the default? i am curious. i do would like this

JainamOswal18:
> what's the default? i am curious. i do would like this

Default as in? You're asking about the llm model?

PR #2773, opened 2025-12-29
Fixed SearchWidget.qml to use proper implicitHeight calculation

radicalgale:
## Describe your changes

The `SearchWidget.qml` has the wrong `implicitHeight` value. Instead of using `searchWidgetContent.implicitHeight` it uses `searchBar.implicitHeight`. This causes the search bar to ignore mouse clicks and only allows the user to launch apps by pressing Return (Enter). 

This issue isn't present if Overview is enabled, because it expands `columnLayout.implicitHeight`, then gets hidden if search isn't equal to "", allowing the mouse clicks to go through. So, to reproduce this issue, try turning off Overview and launching any app in search results by using mouse.

Yes, just a small fix


end-4:
valid, thanks

PR #2779, opened 2025-12-30
Update auto-Hypr.sh/auto-Hypr.fish to use start-hyprland instead for hyprland 0.53

VietNguyenVN:
## Describe your changes

Hyprland 0.53 introduced start-hyprland as the recommended launcher when starting Hyprland from a TTY.

This PR updates the TTY autostart in relevant files to use start-hyprland instead of Hyprland in order to avoid messages on login.

It is a simple fix, so yes it's ready


clsty:
Thanks for the fix.

PR #2812, opened 2026-01-03
Fix when wifi having space in between

PandaKewt:
…modification

## Describe your changes
This change from qt variable to shell environment to prevent code execution like `;shutdown 0;` also adding a quote to SSID to prevent if wifi name have space in it.

Yes, this change is ready


end-4 (approved):
thank you

Issue #1902, opened 2025-08-30
Wallpaper selector missing previews


sansmoraxz:
### Step 1. Before you submit

- [x] I have read the [Troubleshooting](https://end-4.github.io/dots-hyprland-wiki/en/ii-qs/04troubleshooting/) and [Usage](https://end-4.github.io/dots-hyprland-wiki/en/ii-qs/02usage/) pages.
- [x] I've successfully updated to the latest version following the [guidance](https://end-4.github.io/dots-hyprland-wiki/en/ii-qs/01setup/#updating).
- [x] I've successfully updated the system packages to the latest.
- [ ] I've ticked the checkboxes without reading their contents

### Step 2. Quick diagnose info

<details><summary>Quick diagnose</summary>

Running my fork

```

/========================\
| Checking git repo info |
\========================/

[===diagnose===] git remote get-url origin
https://github.com/sansmoraxz/dots-hyprland.git
[---SUCCESS---]

[===diagnose===] git rev-parse HEAD
2f6a793fbd034682ca8ce44b8ea20fe3c4d8e645
[---SUCCESS---]

/=================\
| Checking distro |
\=================/

[===diagnose===] ii_check_distro
./diagnose: line 34: lsb_release: command not found
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo
[---SUCCESS---]

/====================\
| Checking variables |
\====================/

[===diagnose===] declare -p XDG_CACHE_HOME
declare -x XDG_CACHE_HOME="/home/souyamad/.cache"
[---SUCCESS---]

[===diagnose===] declare -p XDG_CONFIG_HOME
declare -x XDG_CONFIG_HOME="/home/souyamad/.config"
[---SUCCESS---]

[===diagnose===] declare -p XDG_DATA_HOME
./diagnose: line 30: declare: XDG_DATA_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p XDG_STATE_HOME
[...]

PR #2818, opened 2026-01-03
Fix: Wallpaper selector missing previews

Ech0-jp:
## Describe your changes
Fixes issue where thumbnails in the Wallpaper Selector would fail to generate/load.
Should fix/help to fix issue #1902 

- WallpaperDirectoryItem.qml:
  - Issue:
  directory would have "file://" prepended to the path. 
  - Fix:
  trimmed the file protocol.
- thumbgen-venv.sh:
  - Issue:
  `deactivate` would always result in `thumbgen-venv.sh` having a status code of 0, preventing Wallpapers.qml generateThumbnail command to never fallback on `generate-thumbnails-magick.sh`
  - Fix:
  Stored the exit code of `thumbgen.py` and exited `thumbgen-venv.sh` with that code. Allowing the shell to see that exit code for the python script.

✅ Ready for review. Tested locally for the `generate-thumbnails-magick.sh` fallback since `thumbgen.py` does not work for me (missing GnomeDesktop exception)


end-4 (approved):
thanks

PR #2819, opened 2026-01-04
feat: add last refresh timestamp to weather popup

TheGreyfeather:
## Describe your changes
Adds the date time to the footer of the Weather Popup. 
It refreshes when the data gets refined, so this should only update on an successful data fetch.

This should be ready, but I don't know if there might be a better way to get the date + time in the right format for this project. I think that it might be worth adding a config option for a datetime value in addition to the existing configurations for date formats.

Happy to make additional changes as directed or requested

### DIFF
--- a/dots/.config/quickshell/ii/modules/ii/bar/weather/WeatherPopup.qml
+++ b/dots/.config/quickshell/ii/modules/ii/bar/weather/WeatherPopup.qml
@@ -101,34 +101,15 @@ StyledPopup {
             }
         }
 
-        // Footer
-        ColumnLayout {
-            id: footer
+        // Footer: last refresh
+        StyledText {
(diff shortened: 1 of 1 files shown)


dots/.config/quickshell/ii/modules/ii/bar/weather/WeatherPopup.qml
```diff
@@ -100,5 +100,35 @@ StyledPopup {
                 value: Weather.data.sunset
             }
         }
+
+        // Footer
+        ColumnLayout {
+            id: footer
+            Layout.alignment: Qt.AlignHCenter
+            spacing: 2
```
end-4: useless column layout

end-4:
thanks

PR #2839, opened 2026-01-07
Hyprpm update error message solved

daniel0alves:
## Describe your changes

- Added hyprpm update to execs.conf because when I was booting I was having to much of the hyprpm update warning
- Added cpio as dependencie/package to install as @narendnp suggested

- It works on arch-linux, needs intervention to add cpio to dist-nix, the other ones I thnk it works

### DIFF
--- a/sdata/dist-arch/illogical-impulse-hyprland/PKGBUILD
+++ b/sdata/dist-arch/illogical-impulse-hyprland/PKGBUILD
@@ -8,4 +8,5 @@ depends=(
   hyprland
   hyprsunset
   wl-clipboard
+  cpio
 )
--- a/sdata/dist-fedora/feddeps.toml
+++ b/sdata/dist-fedora/feddeps.toml
@@ -85,7 +85,8 @@ packages = [
   "hyprland",
   "hyprland-qtutils",
   "hyprsunset",
-  "wl-clipboard"
+  "wl-clipboard",
+  "cpio"
 ]
 install_opts = ["--setopt=install_weak_deps=False"]
 
--- a/sdata/dist-gentoo/illogical-impulse-hyprland/illogical-impulse-hyprland-1.0-r3.ebuild
+++ b/sdata/dist-gentoo/illogical-impulse-hyprland/illogical-impulse-hyprland-1.0-r3.ebuild
@@ -16,4 +16,5 @@ RDEPEND="
 	gui-apps/hyprsunset
 	gui-wm/hyprland
 	gui-apps/wl-clipboard
+  app-arch/cpio
 "


narendnp:
This might not work on all systems since `hyprpm` requires `cpio`, which is not installed by default. Maybe include it on the install script too?

daniel0alves:
Since there was an exec-once hyprpm reload already being used I just made that change, but since you mentioned it, I'm gonna search how to and try to add it

An-dz:
hyprpm doesn't even need to run if you don't have any hyprland plugins. And as far as I'm aware none is installed with illogical-impulse. I'd remove this entirely and leave it for the user to include it if they decide to install one.

daniel0alves:
It used to just work to me, but then I started getting an error message saying to update hyprpm every time I boot so I tried to solve it

An-dz:
I think it never worked, it's just that previous versions of hyprland didn't show a visible warning on the screen

daniel0alves:
So I didn't want to see that error message and this fixes it

An-dz:
You can also fix it by removing `hyprpm reload` from running. It's better to not run something most users don't need, especially one that downloads stuff without asking.

daniel0alves:
Ok, in my logic, the update is good, because if you have an hyprland update and it stops working? I think this is more future proof

An-dz:
It won't stop working, unless you manually installed a hyprland plugin. Because illogical-impulse script doesn't install any for you during the install. And if you do install one yourself, the hyprland wiki and every plugin readme will tell you to add the `exec-once=hyprpm reload` line to your config.

That's why I said this should be removed. Since the install script does not install a hyprland plugin, it means that the majority of the users don't need to run this command, because they don't even have a plugin to load. And those that do use plugins, have already been told to add that to their config.

Also, since plugins basically can run any arbitrary code, it's a security problem having them update automatically. Though it's only a problem for those that do use plugins.

narendnp:
Doesn't ii use hyprexpo plugin for the quick overview?

An-dz:
@narendnp No, it's entirely made in quickshell using [ScreencopyView](https://quickshell.org/docs/v0.2.0/types/Quickshell.Wayland/ScreencopyView/) to draw the windows.

clsty:
> It won't stop working, unless you manually installed a hyprland plugin. Because illogical-impulse script doesn't install any for you during the install. And if you do install one yourself, the hyprland wiki and every plugin readme will tell you to add the `exec-once=hyprpm reload` line to your config.
> 
> That's why I said this should be removed. Since the install script does not install a hyprland plugin, it means that the majority of the users don't need to run this command, because they don't even have a plugin to load. And those that do use plugins, have already been told to add that to their config.
> 
> Also, since plugins basically can run any arbitrary code, it's a security problem having them update automatically. Though it's only a problem for those that do use plugins.

This sounds reasonable. So let's just remove hyprpm both from config and also dependency, is that so?

clsty:
@end-4 Are you fine with removing `hyprpm` (and let users install & load it themselves if they want it)?

An-dz:
> This sounds reasonable. So let's just remove hyprpm both from config and also dependency, is that so?

I think so. But there are no hyprpm dependency, it's installed with hyprland itself on all supported distros.

end-4:
> @end-4 Are you fine with removing `hyprpm` (and let users install & load it themselves if they want it)?

absolutely fine, i actually don't use it either

PR #2840, opened 2026-01-07
Add animation for hovering over a program icons in the dock

PRoX2011:
Added animation when hovering over program icons in the Dock, similar to that used in macOS

Ready to use


0blivi0nis:
Nice, but its making the icons blurry on hover

PRoX2011:
So, merge or no?

PR #2850, opened 2026-01-08
fix wallpaper not switching when setting a custom accent color

reakjra:
## Describe your changes

Since lots of people have been encountering this "problem", this should address that issues when the wallpaper won´t switch if you have set a custom accent color.


Pogdog0:
Love you bro i was so worked up after my wallpaper picker randomly broke

end-4:
This also makes #2659 no longer do what it's supposed to do, but that persistence also seems to cause lots of confusion so not having it persist is better I guess. Also at least that pr still prevents light/dark switching resetting the manual accent color

PR #2854, opened 2026-01-08
Weird Behaviour in Wifi Network Fix with Recent nmcli Versions

create-juicey-app:
## Describe your changes

Simple change fixing Network.qml to avoid network showing as "Connected" while it really isn't, even showing lo for loopback, you don't connect to loopback as a wifi network.. it just doesn't work like that.

This pull request fixes it on recent nmcli versions.

Yes, it's quite a simple change and i am pretty sure it won't break anything.

### DIFF
--- a/dots/.config/quickshell/ii/services/Network.qml
+++ b/dots/.config/quickshell/ii/services/Network.qml
@@ -35,23 +35,23 @@ Singleton {
     property string networkName: ""
     property int networkStrength
     property string materialSymbol: root.ethernet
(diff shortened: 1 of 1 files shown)


juiceydev:
How do i turn copilot off

An-dz:
Good fix, now it works as it used to.

end-4:
I missed this but merged #3063, which seems to also fix this?

create-juicey-app:
closed cuz issue stink

PR #2865, opened 2026-01-10
Add support for custom hyprlock and hypridle config

KAGEYAM4:
## Describe your changes

This adds `source` keyword in hypridle.conf and hyprlock.conf to source custom/{hypridle,hyprlock}.conf

### DIFF
--- a/dots/.config/hypr/custom/hyprlock.conf
+++ b/dots/.config/hypr/custom/hyprlock.conf
@@ -1,0 +1,1 @@
+# You can put custom hyprlock config here


clsty:
Hmm... It sounds reasonable to have custom config for hyprlock and hypridle generally but I do not like the way it's implemented here. I'll probably look into it again when I'm available.

KAGEYAM4:
> Hmm... It sounds reasonable to have custom config for hyprlock and hypridle generally but I do not like the way it's implemented here. I'll probably look into it again when I'm available.

if you are talking about `source=` key being at bottom of file, then i tried moving it at top of file and the config broke, probably because hyprland reads config top to bottom

clsty:
Well the thing is that they should be handled similar as hyprland config.

end-4:
let's have the custom files in custom_hypridle and custom_hyprlock folders instead of also dumping them in the custom folder

clsty:
> let's have the custom files in custom_hypridle and custom_hyprlock folders instead of also dumping them in the custom folder

A folder is normally to contain more than one files, but if we have only one file for each folder, it's kinda meaningless.

I suggest to use a structure like this:
```plain
./config/hypr/
- hyprland.conf (contains only "source" lines)
- hyprlock.conf (contains only "source" lines)
- hypridle.conf (contains only "source" lines)
- illogical-impulse/ (contains configs from illogical-impulse)
  - colors.conf
  - env.conf
  - execs.conf
  - general.conf
  - keybinds.conf
  - rules.conf
  - hyprlock.conf
  - hypridle.conf
- custom/
  - (similar as "illogical-impulse/", but for user customization)
```

KAGEYAM4:
Yes the abovee looks good. 

Can we have another file inside custom named conflicts.conf sourced by illogical-impulse as the first config.

This file will contain all the unset/unbind rules that user want to reset from illogical impulse.

Having this file seprates the user custom code that actually matters to the user, otherwise unset/unbind rules get mixed in custom code.

And if yes, then maybe one step further "conflict-keybind.conf" & "conflict-rules.conf"

clsty:
> This file will contain all the unset/unbind rules that user want to reset from illogical impulse.

Such thing needs manual update when the config from illogical-impulse updates. This will add burden to maintenance.

end-4:
> A folder is normally to contain more than one files, but if we have only one file for each folder, it's kinda meaningless.
> 
> I suggest to use a structure like this:
> [...]

Sure, this is also nice. We can have this, just make sure the install script gets updated accordingly

clsty:
> Sure, this is also nice. We can have this, just make sure the install script gets updated accordingly

That's it. @KAGEYAM4 If you'd like to finish this PR please implement the structure mentioned above and also `sdata/subcmd-install/3.files-legacy.sh` accordingly, thanks in advance.

KAGEYAM4:
I have no experience with install script. Can someone else do this PR .. i will close this one .. i am still learning all commands and syntax.

KAGEYAM4:
> > let's have the custom files in custom_hypridle and custom_hyprlock folders instead of also dumping them in the custom folder
> 
> A folder is normally to contain more than one files, but if we have only one file for each folder, it's kinda meaningless.
> 
> I suggest to use a structure like this:
> 
> ```
> ./config/hypr/
> - hyprland.conf (contains only "source" lines)
> - hyprlock.conf (contains only "source" lines)
> - hypridle.conf (contains only "source" lines)
> - illogical-impulse/ (contains configs from illogical-impulse)
>   - colors.conf
>   - env.conf
>   - execs.conf
>   - general.conf
>   - keybinds.conf
>   - rules.conf
>   - hyprlock.conf
>   - hypridle.conf
> - custom/
>   - (similar as "illogical-impulse/", but for user customization)
> ```

Do i create a new issue for this? I really like the new structure.

clsty:
> > > let's have the custom files in custom_hypridle and custom_hyprlock folders instead of also dumping them in the custom folder
> > 
> > 
> > A folder is normally to contain more than one files, but if we have only one file for each folder, it's kinda meaningless.
> > I suggest to use a structure like this:
> > ```
> > ./config/hypr/
> > - hyprland.conf (contains only "source" lines)
> > - hyprlock.conf (contains only "source" lines)
> > - hypridle.conf (contains only "source" lines)
> > - illogical-impulse/ (contains configs from illogical-impulse)
> >   - colors.conf
> >   - env.conf
> >   - execs.conf
> >   - general.conf
> >   - keybinds.conf
> >   - rules.conf
> >   - hyprlock.conf
> >   - hypridle.conf
> > - custom/
> >   - (similar as "illogical-impulse/", but for user customization)
> > ```
> 
> Do i create a new issue for this? I really like the new structure.

Sure.

PR #2882, opened 2026-01-13
Add fallback to xcb for QT_QPA_PLATFORM

tetrois:
## Describe your changes

Use Wayland if available, fall back to X11 if not.
hyprland [docs](https://wiki.hypr.land/Configuring/Environment-variables/)

Ready to merge


end-4:
uh.. sure i guess? 
though i wonder when might this fallback even be used

tetrois:
Thx.
This is necessary for applications written in QT that do not support Wayland (X11 only). Without this setting, the application will fail to launch with an error like 
```
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, vkkhrdisplay, vnc, wayland-egl, xcb.
```

PR #2895, opened 2026-01-17
Custom time for pomodoro timer

pralad-p:
## Describe your changes

Currently, the Pomodoro timer is locked at 25 minutes (as per [definition of a pomodoro timer](https://en.wikipedia.org/wiki/Pomodoro_Technique#:~:text=typically%2025%20minutes%20in%20length)). However, I would love to have a custom timer that is still a bit rigid to preserve the principle. This commit introduces additional functionality to modify the timer in 10 minutes changes. 
Middle-Mouse-click the Timer text area for incrementing 10 minutes.
Shift+Middle-Mouse-click  the Timer text area for decrementing 10 minutes.

It is ready. But the exact interaction can be up for discussion. I would love to hear back how you would perhaps implement this. Waybar uses the mouse wheel to do the increment/decrement on a minute-wise basis. Would love to hear what would be ideal. I understand this is also quite a cosmetic change, so I don't want to burden the contributors with this minor feature. Thanks.

### DIFF
--- a/dots/.config/quickshell/ii/modules/ii/sidebarRight/pomodoro/PomodoroTimer.qml
+++ b/dots/.config/quickshell/ii/modules/ii/sidebarRight/pomodoro/PomodoroTimer.qml
@@ -32,15 +32,38 @@ Item {
                 anchors.centerIn: parent
                 spacing: 0
 
-                StyledText {
+                Item {
                     Layout.alignment: Qt.AlignHCenter
-                    text: {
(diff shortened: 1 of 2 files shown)


nyx-4:
> I would love to hear back how you would perhaps implement this.

https://github.com/end-4/dots-hyprland/blob/6075835d85110c3c0ef61c3e4ce3a229b700169b/dots/.config/quickshell/ii/modules/common/Config.qml#L566-L571

end-4:
Don't add a new adjustment param. Use the existing config option as above

Issue #2867, opened 2026-01-10
Option to set floating windows to take smaller part of the screen


Qorvil:
### What would you like to be added?

Option to set floating windows to occupy smaller portion of the screen in the settings app.

### How will it help?

Right now, setting a window to be in floating mode requires resizing to make it look neat. A feature in the settings where we can choose if a floating window takes full of the screen or 1/4th of the screen with it positioning in the center would be useful. (Please excuse the HORRIBLE quality, my laptop is old)

https://github.com/user-attachments/assets/1da3ff80-235f-4c21-9d45-ef5e60ba1ee9

end-4:
isn't this just normal hyprland behavior? I'm not sure if there's a config option for that

as for centering, possible but it's worse than just letting the window float in place if we consider fitt's law

end-4:
You can unbind the default keybind and bind it like in the above PR for the desired behavior. Since this is a hyprland option and it deviates from the (well, my) expected default behavior there's nothing to add here

Qorvil:
Oh, thanks...

PR #2899, opened 2026-01-18
toggle floating will detach, center and resize to 50% resolution

ar-Raqmi:
## Describe your changes
* **Toggle Floating (`Super + Alt + Space`)** detaches window, moves it to the center of the screen and resizes it to **50%** of the user's screen resolution.
* https://github.com/end-4/dots-hyprland/issues/2867

* yes, you may change the resizing if the config didn't suit on most user's screen resolution.
* The config can also be resized using exact pixel values (e.g., `exact 500 500` for a  window) instead of percentages.

### DIFF
--- a/sdata/subcmd-exp-update/0.run.sh
+++ b/sdata/subcmd-exp-update/0.run.sh
@@ -113,7 +113,16 @@ safe_read() {
 
   echo -n "$prompt"
   
-  # Try to read from terminal with better detection
+  # First, try reading from stdin (supports piped input like "yes 1 |")
+  if read -r -t 0.1 input_value 2>/dev/null; then
+    # Successfully read from stdin (piped input)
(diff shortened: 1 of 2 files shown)


unixr77:
nice

end-4:
Toggle floating should not center the window. It should pop up where it was so the mouse can grab it directly

ar-Raqmi:
Then, would it be better if we used this command? or this just not going to fit in this main dotfiles?

```
bash -c "if hyprctl activewindow | grep -q 'floating: 0'; then hyprctl --batch 'dispatch togglefloating; dispatch resizeactive exact 50% 50%'; else hyprctl dispatch togglefloating; fi"

end-4:
No, there is no reason to force the window to be 50% screen size. What if someone with an ultrawide floats a clock app which only displays a few digits?

Issue #2816, opened 2026-01-03
Launcher does not perform word splitting


antinomie8:
### Step 1. Before you submit

- [x] I have read the [Troubleshooting](https://ii.clsty.link/en/ii-qs/04troubleshooting/) and [Usage](https://ii.clsty.link/en/ii-qs/02usage/) pages.
- [x] I've successfully updated to the latest version following the [guidance](https://ii.clsty.link/en/ii-qs/01setup/#updating).
- [x] I've successfully updated the system packages to the latest.
- [ ] I've ticked the checkboxes without reading their contents

### Step 2. Quick diagnose info

<details><summary>Quick diagnose</summary>

```
/========================\
| Checking git repo info |
\========================/

[===diagnose===] git remote get-url origin
https://github.com/end-4/dots-hyprland
[---SUCCESS---]

[===diagnose===] git rev-parse HEAD
d5b599da3de8e879745945e5e74de74c32a6e23d
[---SUCCESS---]

[===diagnose===] git status
On branch main
Your branch is up to date with 'origin/main'.

nothing to commit, working tree clean
[---SUCCESS---]

[===diagnose===] git submodule status --recursive
 7f0f0709ec5bbe4c3158c7f5fc68fd890af46618 dots/.config/quickshell/ii/modules/common/widgets/shapes (remotes/origin/HEAD)
[---SUCCESS---]

/=================\
| Checking distro |
\=================/

[===diagnose===] _check_distro_id
distro ID: arch
distro ID_LIKE: 
[---SUCCESS---]

[===diagnose===] cat os-release
cat: os-release: No such file or directory
[---EXIT 1---]

/====================\
| Checking variables |
\====================/

[===diagnose===] declare -p XDG_CACHE_HOME
declare -x XDG_CACHE_HOME="/home/antoine/.cache"
[---SUCCESS---]

[===diagnose===] declare -p XDG_CONFIG_HOME
declare -x XDG_CONFIG_HOME="/home/antoine/.config"
[---SUCCESS---]

[===diagnose===] declare -p XDG_DATA_HOME
declare -x XDG_DATA_HOME="/home/antoine/.local/share"
[---SUCCESS---]

[===diagnose===] declare -p XDG_STATE_HOME
declare -x XDG_STATE_HOME="/home/antoine/.local/state"
[---SUCCESS---]

[===diagnose===] declare -p ILLOGICAL_IMPULSE_VIRTUAL_ENV
[...]

PR #2908, opened 2026-01-20
fix(LauncherSearch): don't concatenate command in a single string

antinomie8:
## Describe your changes

See #2816.
 [Here in LauncherSearch.qml](https://github.com/end-4/dots-hyprland/blob/20ec7717a4c67bcc074dec0cd28638f4f22516f0/dots/.config/quickshell/ii/services/LauncherSearch.qml#L254), a shell is spawned for seemingly no reason and is asked to execute the terminal emulator which in turn is asked to execute a command, but the command is concatenated in a single string; but terminals (or at least some of them) expect the command they should execute not to be concatenated in a space-separated string, but to be splitted across different arguments instead, preventing any `.desktop` file whose `Exec` contains several words (e.g. a command with a flag) and which runs in the terminal (for other apps it's fine) from running correctly.

Currently there are still issues I'm running into: the main one being that single quotes are removed from `entry.command`. For instance, with a desktop file containing the following:
```desktop
Exec=nvim -c "lua require('notmuch.send').compose(string.sub( %u, 8, -1))"
```
If I add a logging statement in `LauncherSearch.qml`:
```qml
console.log(JSON.stringify(entry.command))
```
then I get the following in my logs:
```
DEBUG qml: ["nvim","-c","lua require(notmuch.send).compose(string.sub( %u, 8, -1))"]
```
Single quotes are nowhere to be found, and the command fails to execute as expected (despite working fine with e.g. fuzzel).
That's another issue that needs to be fixed, but which is not directly related to this PR.

### DIFF
--- a/dots/.config/quickshell/ii/services/LauncherSearch.qml
+++ b/dots/.config/quickshell/ii/services/LauncherSearch.qml
@@ -250,7 +250,7 @@ Singleton {
                     if (!entry.runInTerminal)
                         entry.execute();
                     else {
-                        Quickshell.execDetached([Config.options.apps.terminal, "-e", ...entry.command]);
+                        Quickshell.execDetached([...Config.options.apps.terminal.trim().split(/\s+/), "-e", ...entry.command]);
                     }
                 },
                 comment: entry.comment,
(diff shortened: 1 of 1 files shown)


end-4:
The terminal has to go through a shell because Config.options.apps.terminal can also contain flags (the default is `kitty -1`)
Maybe spreading it (`...Config.options.apps.terminal`) would do? 🤔

end-4:
Uh nope not that simple, will gotta split by space, but then it would also be annoying to handle which space is in quotes and which space isn't...

PR #2913, opened 2026-01-22
fix: improve media player deduplication and sorting logic

ToRvaLDz:
This PR improves how media players are processed to prevent duplicates and ensure the active player is correctly prioritized.

### Changes
- **Refactor**: Renamed and rewritten `filterDuplicatePlayers` to `processPlayers` for clarity and better logic.
- **Active Player Priority**: The `activePlayer` is now explicitly added to the distinct list first. This ensures the player currently being controlled is never filtered out by the deduplication logic.
- **Smart Deduplication**:
  - Players with valid metadata are now deduplicated based on a normalized `Title + Artist` key.
  - Players without metadata (ghost players) are deduplicated by grouping them based on approximate duration and position (rounded to 10s intervals).

This fixes issues where duplicate entries would appear for the same media source or where the active player might be hidden by a background instance.

### DIFF
--- a/dots/.config/quickshell/ii/modules/ii/mediaControls/MediaControls.qml
+++ b/dots/.config/quickshell/ii/modules/ii/mediaControls/MediaControls.qml
@@ -72,7 +72,7 @@ Scope {
                  // Round to 10s to catch lagging instances
                  let d = Math.round(p.length / 10);
                  let pos = Math.round(p.position / 10);
-                 key = "empty:" + d + "|" + pos;
+                 key = "empty:" + clean(p.identity) + "|" + d + "|" + pos;
              }
              
              // Since 'active' (playing) is first in list, it claims the key first.


unixr77:
Better than default but not perfect

ToRvaLDz:
> Better than default but not perfect

What else is there to improve?

dots/.config/quickshell/ii/modules/ii/mediaControls/MediaControls.qml
```diff
@@ -17,40 +17,72 @@ Scope {
...
+             if (t) {
+                 // Valid title: Filter duplicates by Title + Artist
+                 key = "content:" + t + "|" + a;
+             } else {
+                 // No title: Filter duplicates by rough duration/position (Ghost players)
+                 // Round to 10s to catch lagging instances
+                 let d = Math.round(p.length / 10);
+                 let pos = Math.round(p.position / 10);
+                 key = "empty:" + d + "|" + pos;
+             }
```
end-4, line 76: Don't you think this would be too sensitive? What if there are many players without time position?

ToRvaLDz: You're right! I've fixed this by including player identity in the deduplication key:

```javascript
key = "empty:" + clean(p.identity) + "|" + d + "|" + pos;
```

Now different players (Firefox vs VLC) stay distinct even without metadata, while true ghost duplicates are still filtered.

Fixed in `fa681f0a`. Thanks for the feedback!

end-4: did you push that to the wrong branch? i don't see it

ToRvaLDz: Yes I pushed to the wrong branch, not it's fixed.

PR #2915, opened 2026-01-22
feat(exp-update): add --default-choice option

BeshoyEhab:
Adds a `--default-choice=N` option to `exp-update` that allows users to specify the default action for file conflicts (1-8).

**Changes:**

- Added `--default-choice=N` flag to set the default conflict resolution choice
- `--non-interactive` now acts as if user pressed Enter for every prompt
- The prompt now shows the current default: `Enter your choice (1-8) [N]:`

**Usage:**

```bash
./setup exp-update -s -f --non-interactive --default-choice=1
```

### DIFF
--- a/sdata/subcmd-exp-update/0.run.sh
+++ b/sdata/subcmd-exp-update/0.run.sh
@@ -371,23 +371,23 @@ handle_file_conflict() {
     echo
 
     while true; do
-      if ! safe_read "Enter your choice (1-8) [${default_val}]: " choice "$default_val"; then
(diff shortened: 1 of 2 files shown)


clsty:
What if the choice number changes in future? I do not think it's a good idea to use numbers.

BeshoyEhab:
So would it be better to choose from flags like 'ignore, replace, keep'?

clsty:
> So would it be better to choose from flags like 'ignore, replace, keep'?

Sure, that's better.

BeshoyEhab:
Is it now better?

clsty:
Yeah, I'll merge it. Thanks for the update.

PR #2931, opened 2026-01-27
Gentoo -> Hyproverlay

jwihardi:
- Switched from using ``hyprland::gentoo`` to ``hyprland::hyproverlay`` since the maintainers moved to that overlay for faster updates.
- QT is no longer masked on unstable, so I removed that
- ``glaze`` and ``pugixml`` are hard coded to install.
  - I made a PR https://codeberg.org/hyproverlay/hyproverlay/pulls/9 to add pugixml to ``hyprwire`` BDEPEND, so I will remove it once that is added.
  - Not sure what's happening with ``glaze``. The ebuild has it as a dependency only when ``hyprpm`` is built, but ``Nix.cpp`` is being built and pulling it in even with hyprpm disabled. I think this is a hyprland issue rather than the ebuild (https://github.com/hyprwm/Hyprland/discussions/13074).

### DIFF
--- a/sdata/dist-gentoo/install-deps.sh
+++ b/sdata/dist-gentoo/install-deps.sh
@@ -30,7 +30,6 @@ if [[ -z $(eselect repository list | grep -E ".*hyproverlay \*.*") ]]; then
 	v sudo eselect repository enable hyproverlay
 fi
 
-
 arch=$(portageq envvar ACCEPT_KEYWORDS)
 
 # Exclude hyprland, will deal with that separately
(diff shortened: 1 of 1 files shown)


clsty:
Thanks for the update!

logrusx:
>       * Not sure what's happening with `glaze`. The ebuild has it as a dependency only when `hyprpm` is built, but `Nix.cpp` is being built and pulling it in even with hyprpm disabled. I think this is a hyprland issue rather than the ebuild ([Undeclared dependency on glaze hyprwm/Hyprland#13074](https://github.com/hyprwm/Hyprland/discussions/13074)).

https://github.com/logrusx/gentoo/tree/main/gui-wm/hyprland/files

logrusx:
>       * I made a PR https://codeberg.org/hyproverlay/hyproverlay/pulls/9 to add pugixml to `hyprwire` BDEPEND, so I will remove it once that is added.

This is wrong, it's DEPEND, not BDEPEND. Hyprwire-scanner links against it.

Issue #2817, opened 2026-01-03
max int workspace when locking


kevinwaxi:
### Step 1. Before you submit

- [x] I have read the [Troubleshooting](https://ii.clsty.link/en/ii-qs/04troubleshooting/) and [Usage](https://ii.clsty.link/en/ii-qs/02usage/) pages.
- [ ] I've successfully updated to the latest version following the [guidance](https://ii.clsty.link/en/ii-qs/01setup/#updating).
- [ ] I've successfully updated the system packages to the latest.
- [ ] I've ticked the checkboxes without reading their contents

## my diagnose results

``` sh
$ ./diagnose
rm: cannot remove 'diagnose.result': No such file or directory

/========================\
| Checking git repo info |
\========================/

[===diagnose===] git remote get-url origin
https://github.com/end-4/dots-hyprland
[---SUCCESS---]

[===diagnose===] git rev-parse HEAD
d5b599da3de8e879745945e5e74de74c32a6e23d
[---SUCCESS---]

[===diagnose===] git status
On branch main
Your branch is up to date with 'origin/main'.

nothing to commit, working tree clean
[---SUCCESS---]

[===diagnose===] git submodule status --recursive
 7f0f0709ec5bbe4c3158c7f5fc68fd890af46618 dots/.config/quickshell/ii/modules/common/widgets/shapes (remotes/origin/HEAD)
[---SUCCESS---]

/=================\
| Checking distro |
\=================/

[===diagnose===] _check_distro_id
distro ID: cachyos
distro ID_LIKE: 
[---SUCCESS---]

[===diagnose===] cat os-release
cat: os-release: No such file or directory
[---EXIT 1---]

/====================\
| Checking variables |
\====================/

[===diagnose===] declare -p XDG_CACHE_HOME
./diagnose: line 36: declare: XDG_CACHE_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p XDG_CONFIG_HOME
./diagnose: line 36: declare: XDG_CONFIG_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p XDG_DATA_HOME
./diagnose: line 36: declare: XDG_DATA_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p XDG_STATE_HOME
./diagnose: line 36: declare: XDG_STATE_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p ILLOGICAL_IMPULSE_VIRTUAL_ENV
[...]

PR #2942, opened 2026-01-30
fix(quickshell): bug temp lock screen workspaces

atdma:
I solved this issue, I saw open issues reporting this as well.

https://github.com/end-4/dots-hyprland/issues/2817

## Describe your changes

Lock screen workspace handling: On lock we move all monitors to a temp workspace (2147483647 − N). That ID was leaking into the bar and overview. We now clamp/filter workspace IDs to 1–100 in HyprlandData, Workspaces.qml, and OverviewWidget.qml so the temp workspace never shows in the UI.

Lock/unlock batching: Lock and unlock now use a single batched hyprctl --batch instead of one call per monitor/delegate, so we don’t race. On unlock we restore the saved workspace per monitor after a short delay.

Looks good to me, tested well.

### DIFF
--- a/dots/.config/quickshell/ii/modules/ii/bar/Workspaces.qml
+++ b/dots/.config/quickshell/ii/modules/ii/bar/Workspaces.qml
@@ -18,8 +18,7 @@ Item {
     property bool borderless: Config.options.bar.borderless
     readonly property HyprlandMonitor monitor: Hyprland.monitorFor(root.QsWindow.window?.screen)
     readonly property Toplevel activeWindow: ToplevelManager.activeToplevel
-    // Clamp to avoid lock-screen temp workspace (2147483647 - N) leaking into UI
-    readonly property int effectiveActiveWorkspaceId: Math.max(1, Math.min(100, monitor?.activeWorkspace?.id ?? 1))
+    readonly property int effectiveActiveWorkspaceId: monitor?.activeWorkspace?.id ?? 1
     
     readonly property int workspacesShown: Config.options.bar.workspaces.shown
     readonly property int workspaceGroup: Math.floor((effectiveActiveWorkspaceId - 1) / root.workspacesShown)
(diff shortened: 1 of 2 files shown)


dots/.config/quickshell/ii/modules/ii/bar/Workspaces.qml
```diff
@@ -18,9 +18,11 @@ Item {
     property bool borderless: Config.options.bar.borderless
     readonly property HyprlandMonitor monitor: Hyprland.monitorFor(root.QsWindow.window?.screen)
     readonly property Toplevel activeWindow: ToplevelManager.activeToplevel
+    // Clamp to avoid lock-screen temp workspace (2147483647 - N) leaking into UI
+    readonly property int effectiveActiveWorkspaceId: Math.max(1, Math.min(100, monitor?.activeWorkspace?.id ?? 1))
```
end-4: Clamping 1-100 is arbitrary. Don't do that

dots/.config/quickshell/ii/modules/ii/lock/Lock.qml
```diff
@@ -11,11 +11,57 @@ import Quickshell.Hyprland
...
+        function onScreenLockedChanged() {
+            if (GlobalStates.screenLocked) {
+                // Lock: save workspace per monitor and move all to temp workspace in one batch
+                var next = {}
+                var batch = "keyword animation workspaces,1,7,menu_decel,slidevert; "
+                for (var i = 0; i < Quickshell.screens.length; ++i) {
+                    var mon = Quickshell.screens[i].name
+                    var mData = HyprlandData.monitors.find(m => m.name === mon)
+                    var ws = (mData?.activeWorkspace?.id ?? 1)
+                    if (ws < 1 || ws > 100) ws = 1
```
end-4: Why the restriction to 100 here

atdma: This made it easier to get back to your workspaces in cases where the bug still occurred. I can remove the arbitrary limit.

KyongSik-Yoon:
+1

end-4:
thank you

Pulsar33550336:
@atdma This PR seems to cause the current window to slide in from the left after unlocking (it used to come from the top), which is a bit unusual

PR #2952, opened 2026-02-01
fix(dock): add robust retry mechanism for desktop entry lookup

Itsmeakash248:
## Describe your changes

Resolves an issue where dock buttons would fail to function on startup if the system's application database wasn't fully loaded. Adds a self-healing timer to DockAppButton.

Ready & tested!

### DIFF
--- a/dots/.config/quickshell/ii/modules/waffle/bar/tasks/TaskAppButton.qml
+++ b/dots/.config/quickshell/ii/modules/waffle/bar/tasks/TaskAppButton.qml
@@ -16,12 +16,12 @@ AppButton {
 
     Timer {
         // Retry looking up the desktop entry if it failed (e.g. database not loaded yet)
-        property int retryCount: 0
-        interval: 500
-        running: !root.isSeparator && root.desktopEntry === null && retryCount < 30
+        property int retryCount: 5
+        interval: 1000
+        running: !root.isSeparator && root.desktopEntry === null && retryCount > 0
         repeat: true
         onTriggered: {
-            retryCount++;
+            retryCount--;
             root.desktopEntry = DesktopEntries.heuristicLookup(root.appEntry.appId);
         }
     }


Itsmeakash248:
anyone here ⏱️?

Itsmeakash248:
@end-4 please review

dots/.config/hypr/hyprland/general.conf
end-4, line 1: unrelated change

dots/.config/quickshell/ii/modules/ii/dock/DockAppButton.qml
```diff
@@ -18,7 +18,18 @@ DockButton {
...
 
     readonly property bool isSeparator: appToplevel.appId === "SEPARATOR"
-    readonly property var desktopEntry: DesktopEntries.heuristicLookup(appToplevel.appId)
+    property var desktopEntry: DesktopEntries.heuristicLookup(appToplevel.appId)
+
+    Timer {
+        // Retry looking up the desktop entry if it failed (e.g. database not loaded yet)
+        interval: 500
+        running: !root.isSeparator && root.desktopEntry === null
+        repeat: true
```
end-4, line 28: Does this just run forever? Please add a max # of retries

dots/.config/hypr/hyprland/general.conf
Itsmeakash248: i have reverted the changes, but its good to keep scrolling enabled when typing because its usefull when gaming

dots/.config/quickshell/ii/modules/ii/dock/DockAppButton.qml
Itsmeakash248: Done i have added 30 retries, is it ok or should i tweak it?

end-4: I think 5 retries with an interval of 1000ms would be more appropriate. And decrement the count to 0 instead of increasing. If you have to do 30 retries then you probably should run something more lightweight

Itsmeakash248: Done! I've refined the logic to use 5 retries with a 1000ms interval, and it now correctly decrements the count to 0.

Itsmeakash248:
Hi @end-4 ! Just a gentle bump on this. I’ve addressed the requested changes in my latest commit. Could you please take another look when you have a moment? Thanks!

end-4:
you didnt adjust the retry count for both timers

end-4:
thanks and sorry for not responding sooner

PR #2964, opened 2026-02-04
Update window rules in colors.conf

niyakipham:
Comment out the old windowrule and add a new windowrulev2 with different colors.


miguelfm:
The use of `windowrulev2` is deprecated since Hyprland 0.53.2. It was replaced by  `windowrule` .

end-4:
closing for the above reason

PR #2970, opened 2026-02-05
feat(screenshot): add option to auto-open annotation tool on copy

ToRvaLDz:
This PR introduces a user-configurable option to control the automatic opening of the annotation tool (e.g., Swappy) when copying a screenshot to the clipboard.

### Changes:
- Added `regionSelector.annotation.autoOpenTool` to `Config.qml` (default: `true`).
- Updated `ScreenshotAction.qml` to conditionally pipe output to the annotation command based on this setting.
- Added a toggle switch in the "Region selector" section of `InterfaceConfig.qml`.
- Added translation keys for "Annotation" and "Open annotation tool after copy" in all supported languages.

### Motivation:
Some users prefer to copy the raw screenshot directly to the clipboard without the interruption of an annotation tool. This option provides that flexibility while maintaining the existing behavior as the default.

### DIFF
--- a/dots/.config/quickshell/ii/modules/common/Config.qml
+++ b/dots/.config/quickshell/ii/modules/common/Config.qml
@@ -449,6 +449,7 @@ Singleton {
                 }
                 property JsonObject annotation: JsonObject {
                     property bool useSatty: false
+                    property bool autoOpenTool: true
                 }
             }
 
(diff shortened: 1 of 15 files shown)


end-4:
the default is not open
check what the ai is yapping on your behalf if there's one...

DRKCTRLDEV:
this is already a thing is it not? by simply right clicking rather than copying...

PR #3000, opened 2026-02-13
make the custom-tela

mohammedaouamri5:
---

### **New file:**

`dots/.config/quickshell/ii/scripts/colors/custom-tela`

This is a **compiled binary** that bundles the Tela icon assets and the logic required to generate a color-adaptive icon theme. see  [custom-tela](https://github.com/mohammedaouamri5/Tela-circle-icon-theme-custom/releases/tag/1.0)

---

### **New file:**

`dots/.config/quickshell/ii/scripts/colors/set-icons.sh`

A simple script that applies a given icon theme system-wide.

---

### **Modified file:**

`dots/.config/quickshell/ii/scripts/colors/applycolor.sh`

I added a new function: `apply_icon`.

It reads the following configuration values:

```bash
enable_icon=$(jq -r '.appearance.wallpaperTheming.enableIcon // false' "$CONFIG_FILE")
user_icons=$(jq -r '.appearance.wallpaperTheming.userIcons // ""' "$CONFIG_FILE")
```

* If `enableIcon` is `false` and `userIcons` is defined →
  The script applies the user-provided icon theme using `set-icons.sh`.

* Otherwise →
  The script extracts the **primary color** from:

```bash
"$STATE_DIR/user/generated/material_colors.scss"

primary_color=$(awk -F': ' '/^\$primary:/ {gsub(/;/,"",$2); print $2}' \
  "$STATE_DIR/user/generated/material_colors.scss")

if [ -z "$primary_color" ]; then
  echo "Primary color not found. Skipping icon generation."
  return
fi

primary_color="${primary_color#\#}"
```

It then generates a color-adaptive Tela theme using:

```bash
"$CONFIG_DIR/scripts/colors/custom-tela" custom-tela "$primary_color"
```

Finally, it applies the generated theme:

```bash
"$CONFIG_DIR/scripts/colors/set-icons.sh" \
  "$HOME/.local/share/icons/custom-tela"
```

---

## Questions / Feedback Needed

* I haven’t worked with Qt before, so I may need some guidance integrating this into the dotfiles configuration.
* About the  [ `custom-tela`](https://github.com/mohammedaouamri5/Tela-circle-icon-theme-custom/releases/tag/1.0)  binary included in this contribution:

  * It is **3.6 MB**.
  * I’ve published the compilation steps and provided **CRC and MD5 checksums** so it can be fully verified.
  * So what shall I do :
    * Commit the binary directly,
    * Include only the source code with build instructions, or
    * Add a build step to generate it during installation?

I’m happy to adjust this contribution however we think is best, including providing a build script or an AUR package if that’s helpful.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "dots/.config/quickshell/ii/modules/common/widgets/shapes"]
 	path = dots/.config/quickshell/ii/modules/common/widgets/shapes
 	url = https://github.com/end-4/rounded-polygon-qmljs.git
+[submodule "dots/.config/quickshell/ii/assets/tela-custom"]
+	path = dots/.config/quickshell/ii/assets/tela-custom
+	url = https://github.com/mohammedaouamri5/tela-custom.git
--- a/dots/.config/quickshell/ii/scripts/colors/applycolor.sh
+++ b/dots/.config/quickshell/ii/scripts/colors/applycolor.sh
@@ -8,6 +8,7 @@ CONFIG_DIR="$XDG_CONFIG_HOME/quickshell/$QUICKSHELL_CONFIG_NAME"
 CACHE_DIR="$XDG_CACHE_HOME/quickshell"
 STATE_DIR="$XDG_STATE_HOME/quickshell"
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+TELA_ICONS="$CONFIG_DIR/assets/tela-custom"
 
 term_alpha=100 #Set this to < 100 make all your terminals transparent
 # sleep 0 # idk i wanted some delay or colors dont get applied properly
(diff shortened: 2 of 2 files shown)


mohammedaouamri5:
**I removed the  the  binary file and i just clone it as a `git submodule` so it should be fine**

end-4:
I'm not interested in adding Tela icons. But I'll point out this:
Adding a binary without the source is unacceptable. That last commit didn't remove it, just adds a submodule, which 
- Leads to your own fork of Tela which adds a bash script. If it's a script for color applying then why not include it in the PR?
- Contains only bash, which does not need to be compiled.

I believe that this is just a naive attempt in good faith, but this could look like a malicious attempt to add a backdoor.

mohammedaouamri5:
not a problem 
for the icons i just want to add the ability to add any icons because the scripts of wallpaper switching reset the default breeze icons
and for the binary actually i removed in the next commit 
any way thank you for you time

PR #3002, opened 2026-02-15
fix(quickshell): mpris not detecting browser players when plasma-browser-integration is installed

GokuPlay609:
## Describe your changes

The bar media widget (and media controls) wouldn't detect browser MPRIS players (Firefox, Chromium) if `plasma-browser-integration-host` was installed on the system — even if Plasma integration wasn't actually running.

The old code spawned a process to check if the binary exists on disk. If it did, it filtered out all native browser MPRIS buses, assuming Plasma would provide duplicates. But just having the package installed (e.g. pulled as a dependency) doesn't mean it's active, so the filter would kill the only player available.

This replaces the subprocess check with a reactive property that looks at whether a `plasma-browser-integration` player is actually registered on D-Bus among the current `Mpris.players`. Native browser buses only get filtered when there's genuinely a duplicate to filter.

Also gets rid of the bash subprocess spawn on startup which is nice.

Tested on Arch with Firefox + `plasma-browser-integration` installed but not running. Media shows up properly now.

### DIFF
--- a/dots/.config/quickshell/ii/services/MprisController.qml
+++ b/dots/.config/quickshell/ii/services/MprisController.qml
@@ -25,22 +25,16 @@ Singleton {
 
 	property var activeTrack;
 
-	property bool hasPlasmaIntegration: false
-    Process {
-        id: plasmaIntegrationAvailabilityCheckProc
-        running: true
-        command: ["bash", "-c", "command -v plasma-browser-integration-host"]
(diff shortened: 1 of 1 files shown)


GokuPlay609:
i forget to tell , this issue was recognized and fix by claude opus 4.6 .

end-4:
Interesting. Thanks

PR #3003, opened 2026-02-15
Add a script that only allows one browser to run

wkr111:
Add a script that only allows one browser to run, improving work efficiency. This is merely an optional script for users.

### DIFF
--- a/dots/.config/hypr/custom/scripts/browser_toggle.sh
+++ b/dots/.config/hypr/custom/scripts/browser_toggle.sh
@@ -1,0 +1,82 @@
+#!/usr/bin/env bash
+
+# browser_toggle.sh - Hyprland browser toggler
+# Usage: ./browser_toggle.sh <browser-commands...>
+# Selects first installed browser from args, focuses existing window,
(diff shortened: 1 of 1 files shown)


wkr111:
An example of usage:
`
bind = Super, W, exec, ~/.config/hypr/custom/scripts/browser_toggle.sh "zen-browser" "floorp" "waterfox" "brave" "google-chrome-stable" "firefox" "microsoft-edge-stable" "vivaldi-stable" "librewolf" "chromium" "opera" "helium" # Browser
`

end-4:
I will not add a script that is not used, and won't use this by default because it deviates from the keybind-to-open-window you'd normally expect

wkr111:
> I will not add a script that is not used, and won't use this by default because it deviates from the keybind-to-open-window you'd normally expect

Alright, but when the browser is already running, it does bring the browser into focus.

Anyway, thanks for the explanation.

wkr111:
Or how about moving the browser to the current window?

wkr111:
Thank you, I understand. Regardless, this feature goes against the user's original intent of opening a window.

PR #3028, opened 2026-02-21
[Simple] Add support for RTL in the translator

Majoramari:
## Describe your changes

Before:
image

After:
image

I used this fix for the months, but I’m not sure if it’s the most optimal solution.


end-4:
Both look like noodles to me so I can't tell 100% but makes sense I guess, thanks

PR #3040, opened 2026-02-25
fix: sudo keepalive cleanup leaking exit code 143

GokuPlay609:
## Describe your changes

The EXIT trap running `sudo_stop_keepalive` causes `./setup install` to exit with code 143 instead of 0, even on a fully successful install.

`wait` on a SIGTERM-killed background process returns 143 (128+15). With `set -e` active, this leaks into the script's final exit code, so the user's shell prompt shows it as a failure.

Fix: add `|| true` after `kill` and `wait` in `sudo_stop_keepalive()`, matching the existing convention used elsewhere in the codebase.

Ready. Tested across multiple scenarios (normal exit, Ctrl-C, already-dead process, real failures still propagate correctly).


clsty:
Thanks for the fix.

PR #3071, opened 2026-03-04
Gentoo updates

jwihardi:
Updated ``illogical-impulse-quickshell-git`` git commit version. Added ``vulkan-headers`` build dependency. 

Slightly changed ``keywords``. Removed the unnecessary specification of ``::hyproverlay`` and added a few new packages.


clsty:
Thanks for update!

PR #3082, opened 2026-03-09
change the weather service

### LINKED ISSUE #1918: replace wttr.in with openweathermap
### **why?**

in short, it's just hot garbage.

for example, right now it's 24 C where i'm at, both google weather and OWM show that.
but the bar for whatever reason shows 28 C & sunny (it's the dead of night)
basically anyone can reproduce this, wttr is known to be stupid

cons: 

- api key needed (takes 2 clicks, the popup could guide you through it if it detects a key is not set)

pros:

- more stats you can fetch
- actually accurate
- is not shit
- easy to implement
- does not go down every 5 seconds

issues related to this while we're at it :

- weather indicator going poof if the bar is vertical
- temp basically never refreshing unless you reboot or click on it (might be wrong)

### How will it help?

gives you actually accurate data.

### Extra info

_No response_

rijaluddina:
1. Total Overhaul of services/Weather.qml
   * New Data Source: Replaced wttr.in with a two-stage flow: * Location: Use ip-api.com to get coordinates (Lat/Lon) and city name directly
         automatically.
       * Weather: Uses Open-Meteo API which is more stable and accurate
with coordinate data
         that.
   * Icon Mapping: Created wmoToWwo function to convert Open-Meteo weather codes to icon format
     existing ones, so that the icon display remains accurate.
   * Unit Conversion: Added automatic conversion logic for Metric (°C, km/h, mm) and units
     Imperial (°F, mph, in) based on user settings.
   * Reactivity: Changed the interval property to respond immediately to changes from the GUI Settings
     no restart required (automatically restarts timer when set).

  2. Feature & Settings Integration
   * Manual Refresh: Ensures the Right Click feature on the Weather Bar still functions to trigger
     Instant data retrieval with notification.
   * GUI Settings: Enables the Polling Interval setting in the Settings > Services menu
     actually control the timer in the background in real-time.
     
fix: end-4/dots-hyprland#1918

### DIFF
--- a/dots/.config/quickshell/ii/services/Weather.qml
+++ b/dots/.config/quickshell/ii/services/Weather.qml
@@ -2,166 +2,186 @@ pragma Singleton
 pragma ComponentBehavior: Bound
 
 import Quickshell
-import Quickshell.Io
 import QtQuick
-import QtPositioning
-
 import qs.modules.common
+import qs.services
 
 Singleton {
     id: root
-    // 10 minute
+
     readonly property int fetchInterval: Config.options.bar.weather.fetchInterval * 60 * 1000
-    readonly property string city: Config.options.bar.weather.city
     readonly property bool useUSCS: Config.options.bar.weather.useUSCS
+    
+    // For backward compatibility and UI settings
     property bool gpsActive: Config.options.bar.weather.enableGPS
+    readonly property string city: Config.options.bar.weather.city
 
-    onUseUSCSChanged: {
-        root.getData();
+    onUseUSCSChanged: getData()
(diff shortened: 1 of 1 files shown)


ayushjaipuriyar:
when we have geoclue why do we need to ping an ip ?

rijaluddina:
> when we have geoclue why do we need to ping an ip ?

because I tried using geoclue but it didn't work and ip-api was the easiest solution. maybe you can help solve the Dbus geoclue problem I would really appreciate it

end-4:
geoclue works for me... 
using an api would give wrong results for those using a vpn

PR #3094, opened 2026-03-15
Fix zen-browser startup command

hwc2102:
## Describe your changes

Fix zen-browser startup command
The latest version of zen-brower(zen-browser-1.17.15b-2), it cannot start by pressing SUPER+W.It's startup command is "zen", not "zen-browser"


end-4:
App launch commands are now overridable in `~/.config/hypr/custom/variables.conf`, so please config that for yourself

PR #3103, opened 2026-03-20
click-to-mute on app icons in the volume mixer

PicoShot:
## Describe your changes
This PR introduces the ability to selectively mute individual app audio streams directly from the Volume Mixer overlay without lowering their volume sliders. This preserves their original volume levels when unmuting.

image
image

Yes, it is fully tested and ready to merge

### DIFF
--- a/dots/.config/quickshell/ii/modules/ii/sidebarRight/volumeMixer/VolumeMixerEntry.qml
+++ b/dots/.config/quickshell/ii/modules/ii/sidebarRight/volumeMixer/VolumeMixerEntry.qml
@@ -7,10 +7,8 @@ import Quickshell
 import Quickshell.Services.Pipewire
 import Qt5Compat.GraphicalEffects
 
-MouseArea {
+Item {
     id: root
-    cursorShape: Qt.PointingHandCursor
-    onClicked: root.node.audio.muted = !root.node.audio.muted
     required property PwNode node
     PwObjectTracker {
         objects: [root.node]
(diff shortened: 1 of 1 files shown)


end-4:
Since it's not destructive, don't use red icons
I think it's better to tint the app icon colors a bit towards the background color then have the mute icon use text color

dots/.config/quickshell/ii/modules/ii/sidebarRight/volumeMixer/VolumeMixerEntry.qml
```diff
@@ -20,19 +21,51 @@ Item {
...
+                text: root.node?.isSink ? "volume_off" : "mic_off"
+                font.pixelSize: 22
+                color: Appearance.colors.colError
+            }
+
+            MouseArea {
+                anchors.fill: parent
+                cursorShape: Qt.PointingHandCursor
+                onClicked: root.node.audio.muted = !root.node.audio.muted
             }
```
end-4: Instead of adding a MouseArea inside, just make the root Item a MouseArea

PicoShot:
Thanks for the review, I've updated the PR:

Changed the mute icon color from colError to colOnLayer1 (text color)
Changed the root Item to a MouseArea instead of nesting one inside
Let me know if anything else needs adjustment.

It's now looking like this:
imageimage

end-4:
uh
my bad the mouse area should be on the icon not the whole thing

end-4:
i fixed that last bit
thanks for your work

PR #3119, opened 2026-03-22
Gentoo Update

jwihardi:
- Updated _quickshell_ commit version 
- Updated the revision number (should not have 9999, it's not a live ebuild)
- Instead of deleting breakpad, just disabled the useflag by default
  - breakpad, i3, i3-ipc disabled by default
- Added := slot operations to most qt dependencies ( rebuild if dependency slots change, shouldn't hurt )
- Use ``get_libdir`` instead of hardcoded ``lib64``
- ``kde-frameworks/kimageformats`` with the ``avif`` USE flag is the equivalent to ``qt6-avif-image-plugin``
- Added ``cpptrace`` with ``unwind`` USE flag
  - I added the ``unwind`` USE flag to the gentoo overlay [here](https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=a682ee1f5e0a7ce11cd2fd9228e113b11b0c98e3), it will take a while to sync.
- Moved ``depclean`` to the end because if someone's install failed, and they fixed something then reran the installer, it would erase any partial progress ( installed deps but not the main package ) 😭

### DIFF
--- a/sdata/dist-gentoo/useflags
+++ b/sdata/dist-gentoo/useflags
@@ -115,5 +115,5 @@ dev-libs/glib dbus elf introspection mime xattr
 #gui-apps/wlogout (no use flags)
 media-gfx/imagemagick xml
 
-################### WIDGETS ###################
+################### OTHER ###################
 dev-cpp/cpptrace unwind


jwihardi:
Also, I have the version as 0.1.0 since the PKGBUILD has that, but looking at the GitHub I think we're actually on 0.2.0. I'm not sure which version number you would like to use.

clsty:
> Also, I have the version as 0.1.0 since the PKGBUILD has that, but looking at the GitHub I think we're actually on 0.2.0. I'm not sure which version number you would like to use.

That `0.1.0` actually does not matter since it's just a meta info of the local PKGBUILD instead of upstream. The commit hash is the real point.

clsty:
I'll merge this. Thanks for update!

jwihardi:
> > Also, I have the version as 0.1.0 since the PKGBUILD has that, but looking at the GitHub I think we're actually on 0.2.0. I'm not sure which version number you would like to use.
> 
> That `0.1.0` actually does not matter since it's just a meta info of the local PKGBUILD instead of upstream. The commit hash is the real point.

I know, I thought maybe it would look nicer 😃

PR #3120, opened 2026-03-22
Clear text input on searchWidget.cancelSearch();

vananh2801:
I sometimes used windows + V to open history panel. After a while, if I pressed windows key and used arrow keys to change workspace, it would move cursor on text input panel instead of changing workspace. I think we should disable one of them when we use arrow keys: changing workspace or editing input text. This change will remove input text automatically.

### DIFF
--- a/dots/.config/quickshell/ii/modules/ii/overview/SearchWidget.qml
+++ b/dots/.config/quickshell/ii/modules/ii/overview/SearchWidget.qml
@@ -37,7 +37,7 @@ Item { // Wrapper
     }
 
     function cancelSearch() {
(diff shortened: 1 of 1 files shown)


end-4:
The text not being cleared and being selected is by design. This helps when you accidentally close the overview or have to close it to see more stuff to type into it (for example scrolling down a table to see entries for calculation)

Arrow keys switching workspaces was added just because it was possible; it's not the primary function. If you already have your left hand at the windows key and your right hand at the arrows you can just also hold ctrl and use the keybind (ctrl+windows+arrows) to switch workspaces anyway

PR #3121, opened 2026-03-22
Add XWayland configuration to fix blurry fonts while using fractional scale

vananh2801:
It will disable windows scaling on xwayland software to prevent blurry fonts. Source: https://wiki.hypr.land/Configuring/XWayland/


end-4:
Whether or not this is better really depends on specific setups but always having a clean look is good I guess

PR #3122, opened 2026-03-22
Decrease gamma when attempting to decrease brightness below 0

altrup:
## Describe your changes

- Attempting to decrease brightness when brightness is 0 will result in gamma being reduced
- Increasing brightness prioritizes increasing gamma back to 100 before increasing actual screen brightness
- Adds slider to night light config for gamma
- Updates quick slider, so button 30% is used to adjust gamma
- Adds on screen display for gamma

image
image
image

- Seems to work fine for me, not sure how intuitive it is though, but figured I'd make the PR in case people wanted it
- Will note that I switched hyprsunset to be always running, not sure if that breaks something, but it seems to be fine so far

### DIFF
--- a/dots/.config/quickshell/ii/modules/ii/onScreenDisplay/indicators/GammaIndicator.qml
+++ b/dots/.config/quickshell/ii/modules/ii/onScreenDisplay/indicators/GammaIndicator.qml
@@ -9,5 +9,6 @@ OsdValueIndicator {
 
     icon: "wb_twilight"
     name: Translation.tr("Gamma")
+    from: Hyprsunset.gammaLowerLimit / 100
     value: Hyprsunset.gamma / 100 ?? 0.5
 }
--- a/dots/.config/quickshell/ii/modules/ii/onScreenDisplay/OsdValueIndicator.qml
+++ b/dots/.config/quickshell/ii/modules/ii/onScreenDisplay/OsdValueIndicator.qml
@@ -11,6 +11,8 @@ Item {
     required property string name
     property bool rotateIcon: false
     property bool scaleIcon: false
+    property alias from: valueProgressBar.from
+    property alias to: valueProgressBar.to
 
     property real valueIndicatorVerticalPadding: 9
     property real valueIndicatorLeftPadding: 10
--- a/dots/.config/quickshell/ii/modules/ii/sidebarRight/QuickSliders.qml
+++ b/dots/.config/quickshell/ii/modules/ii/sidebarRight/QuickSliders.qml
@@ -42,9 +42,9 @@ Rectangle {
             sourceComponent: QuickSlider {
                 materialSymbol: "light_mode"
                 secondaryMaterialSymbol: "wb_twilight"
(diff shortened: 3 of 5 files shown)


end-4:
This is interesting

> Will note that I switched hyprsunset to be always running, not sure if that breaks something, but it seems to be fine so far

It should be fine as long as hyprsunset doesn't randomly break. When I first set it up the ipc would randomly fail and need relaunch; idk how it is now

end-4:
the scrolling on the bar currently doesn't change gamma

dots/.config/quickshell/ii/modules/ii/onScreenDisplay/indicators/GammaIndicator.qml
```diff
@@ -0,0 +1,14 @@
...
+import QtQuick
+import Quickshell
+import Quickshell.Hyprland
+import qs.modules.ii.onScreenDisplay
+
+OsdValueIndicator {
+    id: rotateIcon
+
+    icon: "wb_twilight"
+    scaleIcon: true
```
end-4: don't scale it, it doesn't make sense for this icon (why would the land on the horizon shrink)

dots/.config/quickshell/ii/modules/common/widgets/StyledSlider.qml
```diff
@@ -94,71 +96,100 @@ Slider {
...
-                topRightRadius: root.unsharpenRadius
-                bottomRightRadius: root.unsharpenRadius
+        Repeater {
+            model: background.leftWidths.length
+
+            Loader {
+                required property real index
+                anchors {
+                    verticalCenter: background.verticalCenter
+                }
```
end-4: don't nest if it's just one value

dots/.config/quickshell/ii/modules/common/widgets/StyledSlider.qml
```diff
@@ -177,7 +208,7 @@ Slider {
 
         implicitWidth: root.handleWidth
         implicitHeight: root.handleHeight
-        x: root.handleMargins + (root.visualPosition * root.effectiveDraggingWidth) - (root.handleWidth / 2)
+        x: leftPadding + (root.visualPosition * root.effectiveDraggingWidth) - (root.handleWidth / 2)
```
end-4: `root.leftPadding` for qualified access

end-4:
It's currently daytime and I'm seeing a lower bound for gamma of 25 would be appropriate, but we should probably see how it is at night

altrup:
> the scrolling on the bar currently doesn't change gamma

Oh I forgot about that, I'll add it for that as well

Also for lower bound, I think that's reasonable, I'll also say I was trying it and once you get too low it gets hard to read just because the contrast also gets lower when you decrease gamma

edit: actually nvm I didn't see you already did it, i'll add the icon though

end-4:
tested in a dark room with lights off, 25 seems fine, might even be too dark, but linux users are weird anyway so doesn't hurt having the limit a lil generous

PR #3127, opened 2026-03-23
Update gtk4 styles for Material 3 Expressive

fb24m:
Currently some GTK4 components (like shapes/spacing) differ from the rest of the theme, which makes them visually stand out compared to the overall Material design.

This PR improves that consistency without changing the project's focus or design direction. I’ve included a few screenshots for demonstration, though they may differ slightly

nautilus
old vs new
additional windows

### DIFF
--- a/sdata/subcmd-install/2.setups.sh
+++ b/sdata/subcmd-install/2.setups.sh
@@ -68,6 +68,5 @@ if [[ "$OS_GROUP_ID" == "gentoo" ]]; then
 fi
 
 v gsettings set org.gnome.desktop.interface font-name 'Google Sans Flex Medium 11 @opsz=11,wght=500'
-v gsettings set org.gnome.desktop.wm.preferences button-layout ":"
 v gsettings set org.gnome.desktop.interface color-scheme 'prefer-dark'
 v kwriteconfig6 --file kdeglobals --group KDE --key widgetStyle Darkly
--- a/sdata/subcmd-install/2.setups.sh
+++ b/sdata/subcmd-install/2.setups.sh
@@ -68,5 +68,6 @@ if [[ "$OS_GROUP_ID" == "gentoo" ]]; then
 fi
 
 v gsettings set org.gnome.desktop.interface font-name 'Google Sans Flex Medium 11 @opsz=11,wght=500'
+v gsettings set org.gnome.desktop.wm.preferences button-layout ":"
 v gsettings set org.gnome.desktop.interface color-scheme 'prefer-dark'
 v kwriteconfig6 --file kdeglobals --group KDE --key widgetStyle Darkly
--- a/dots/.config/matugen/templates/gtk-4.0/gtk.css
+++ b/dots/.config/matugen/templates/gtk-4.0/gtk.css
@@ -266,6 +266,7 @@ banner widget {
 }
 
 .boxed-list {
+  box-shadow: none;
   background-color: @window_bg_color;
 }
 
--- a/dots/.config/matugen/templates/gtk-4.0/gtk.css
+++ b/dots/.config/matugen/templates/gtk-4.0/gtk.css
@@ -7,6 +7,7 @@
     /* Accents */
     @define-color accent_color {{colors.primary.light.hex}};
     @define-color accent_hover_color rgba({{colors.primary.light.red}}, {{colors.primary.light.green}}, {{colors.primary.light.blue}}, 0.08);
+    @define-color accent_vibrant_hover_color rgba({{colors.primary.light.red}}, {{colors.primary.light.green}}, {{colors.primary.light.blue}}, 0.14);
     @define-color accent_active_color rgba({{colors.primary.light.red}}, {{colors.primary.light.green}}, {{colors.primary.light.blue}}, 0.1);
     @define-color accent_fg_color {{colors.on_primary.light.hex}};
     @define-color accent_bg_color {{colors.primary.light.hex}};
@@ -53,6 +54,7 @@
     /* Accents */
     @define-color accent_color {{colors.primary.dark.hex}};
     @define-color accent_hover_color rgba({{colors.primary.dark.red}}, {{colors.primary.dark.green}}, {{colors.primary.dark.blue}}, 0.08);
+    @define-color accent_vibrant_hover_color rgba({{colors.primary.dark.red}}, {{colors.primary.dark.green}}, {{colors.primary.dark.blue}}, 0.14);
     @define-color accent_active_color rgba({{colors.primary.dark.red}}, {{colors.primary.dark.green}}, {{colors.primary.dark.blue}}, 0.1);
     @define-color accent_fg_color {{colors.on_primary.dark.hex}};
     @define-color accent_bg_color {{colors.primary.dark.hex}};
--- a/dots/.config/matugen/templates/gtk-4.0/gtk.css
+++ b/dots/.config/matugen/templates/gtk-4.0/gtk.css
@@ -378,7 +378,7 @@ switch:checked slider {
 
 toast {
   border-radius: 999px;
-  padding: 3px 6px;
+  padding: 3px 3px 3px 6px;
   background-color: @inverse_surface;
   color: @inverse_on_surface;
 }
--- a/dots/.config/matugen/templates/gtk-4.0/gtk.css
+++ b/dots/.config/matugen/templates/gtk-4.0/gtk.css
@@ -1,6 +1,6 @@
 /*
-* GTK Colors
-* Generated with Matugen
+* GTK colors generated with Matugen
+* The source template is here: ~/.config/matugen/templates/gtk-4.0/gtk.css
 */
 
 @media (prefers-color-scheme: light) {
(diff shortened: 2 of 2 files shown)


end-4:
This looks sick
Though don't you think that last piece of breadcrumb is too prominent?

fb24m:
> This looks sick Though don't you think that last piece of breadcrumb is too prominent?

image

how about this one?

fb24m:
Updated tabs style 

image

end-4:
I think breadcrumb pieces should not have varying colors
As for tabs, using the underline style is okay, but it's really weird when the [x] button is way on the right... I think we should just stick to the default tab but make it fully round so it feels material

end-4:
for breadcrumb color just use the color of the rightmost piece right now

fb24m:
alright

fb24m:
image

end-4:
Scrollbar looks really weird

image

Switches have white pixels in the Off state

image

Hover color seems to be wrong for vertical groups: I don't think it's supposed to use a more vibrant color role

image

Breadcrumbs' hover state looks barely different from default

image

end-4:
Nautilus tab close button is vertically misaligned
...actually, the text too

image

fb24m:
fixed that

i also stylized tooltips and right-click menus

fb24m:
We could also hide all window buttons with command:

```bash
gsettings set org.gnome.desktop.wm.preferences button-layout ""
```

but I'm not sure if I should add that to the install script or how to do it

clsty:
> We could also hide all window buttons with command:
> 
> ```shell
> gsettings set org.gnome.desktop.wm.preferences button-layout ""
> ```
> 
> but I'm not sure if I should add that to the install script or how to do it

https://github.com/end-4/dots-hyprland/blob/aea6216a7fc15bddbf547e4fbe6040ba8632fdbc/sdata/subcmd-install/2.setups.sh#L70

fb24m:
> > We could also hide all window buttons with command:
> > ```shell
> > gsettings set org.gnome.desktop.wm.preferences button-layout ""
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > but I'm not sure if I should add that to the install script or how to do it
> 
> https://github.com/end-4/dots-hyprland/blob/aea6216a7fc15bddbf547e4fbe6040ba8632fdbc/sdata/subcmd-install/2.setups.sh#L70

thanks

end-4:
The switch still looks bad when toggling... Maybe transparentize that border instead of making it black?

https://github.com/user-attachments/assets/b79618c4-2b08-491a-b4b8-535fd39218dc

end-4:
Scrollbar is still weird, just that the border is now less noticeable

image

end-4:
> fixed that
> 
> i also stylized tooltips and right-click menus

now the body part and the arrow part of the bubble are of different colors

image

end-4:
Nautilus breadcrumb hover got better but still not good enough. And pressing makes no difference
image

end-4:
button next to address bar is not properly themed (bad corners)

image

fb24m:
> Scrollbar is still weird, just that the border is now less noticeable
> 
> image
> 

it seems like the gtk default, not sure if i can remove this shadow

or the style is clashing with the switch

end-4:
> it seems like the gtk default, not sure if i can remove this shadow

fine then i guess

fb24m:
> The switch still looks bad when toggling... Maybe transparentize that border instead of making it black?
> 
>  recording_2026-04-05_19.52.22.mp4

it's literally the same with transparent border
i actually can't animate the toggle size using just css and it looks like border smoothly becomes darker

https://github.com/user-attachments/assets/a2f7e41f-bf19-40d1-ab2a-8e66fe15f833

fb24m:
> Scrollbar is still weird, just that the border is now less noticeable
> 
> image

make the scrollbar similar to the one in settings

image

end-4:
Now the scrollbar is like this :sob:
edit: for now i un-styled it

image

end-4:
This padding is really bad
edit: resolved

image

end-4:
Another problem is that the color variables are mixed together with the default Adwaita variables. New variables we define should be separated to avoid confusion

end-4:
would you happen to know why dropdown items that are of the "bubble" type have this weird grouping?

image

fb24m:
> would you happen to know why dropdown items that are of the "bubble" type have this weird grouping?
> 
> image
> 

because gtk for some reason uses vertical list layout for some context menus
will fix it

fb24m:
should be ok now

fb24m:
> Now the scrollbar is like this 😭 edit: for now i un-styled it
> 
> image

is it in nautilus?

fb24m:
> This padding is really bad edit: resolved
> 
> image

don't you think we should use a toast just like in android?

image

end-4:
> don't you think we should use a toast just like in android?
> 
> [image]

hmmm interesting... i like this one. i was following the m3 specs which had the inverted color: https://m3.material.io/components/snackbar/overview

EDIT: nevermind, it works there because it's over a wallpaper, which has a different color. not sure if it'd work well for the app

end-4:
there's also this problem. see the white corners?
(not caused by that above commit btw, it's just something i only noticed now)

image

end-4:
I tried surface container high for the toast, not super great (i mean the background, ignore the white Undo text)
Let's just stick with inverse surface for now, but have the full rounded corner

image

fb24m:
> > don't you think we should use a toast just like in android?
> > [image]
> 
> hmmm interesting... i like this one. i was following the m3 specs which had the inverted color: https://m3.material.io/components/snackbar/overview
> 
> EDIT: nevermind, it works there because it's over a wallpaper, which has a different color. not sure if it'd work well for the app

they use a different color and shadow in Messages, not sure if it's contrast enough, but i really don't like the big white toast in a dark app

image

fb24m:
image

what about secondary container?

end-4:
> image
> what about secondary container?

It feels off... I don't think primary/secondary/tertiary are supposed to be used for whole containers except in a "widget boxes" layout

fb24m:
> > image
> > what about secondary container?
> 
> It feels off... I don't think primary/secondary/tertiary are supposed to be used for whole containers except in a "widget boxes" layout

Google uses "primary container" for buttons, switch containers in settings, etc, so i think it's fine for toast

end-4:
The whole toast is not a button

fb24m:
I decreased paddings to match with android and font weight to match with material 3 

image

sdata/subcmd-install/2.setups.sh
```diff
@@ -68,5 +68,6 @@ if [[ "$OS_GROUP_ID" == "gentoo" ]]; then
 fi
 
 v gsettings set org.gnome.desktop.interface font-name 'Google Sans Flex Medium 11 @opsz=11,wght=500'
+v gsettings set org.gnome.desktop.wm.preferences button-layout ":"
```
end-4: Let's not do this
Having no [x] button feels weird

fb24m: isn't it normal for hyprland?

end-4: Browsers (or chromium at least) would not have anything on the top-right corner this way (and would feel empty). The button is still useful if you're somehow using your system with only one hand holding the mouse.

fb24m: it's weird that there’s a button in some apps but not in others
actually, you can hide the button in chromium using system titlebar and borders (an option in settings/appearance)

fb24m: if i want consistent borders and x button in all apps, i'd use border++

end-4: Plugins are a pain. I'd rather not rebuild every Hyprland release. Having "consistent" close buttons is not really the goal

> it's weird that there’s a button in some apps but not in others

But not having any button is also weird... Take dolphin for example, it doesn't have client-side decoration and thus no [x] but still has this hamburger menu button on the right
image

Chromium without the close button is just this
image


> actually, you can hide the button in chromium using system titlebar and borders (an option in settings/appearance)

And hiding is not showing

end-4:
> I decreased paddings to match with android and font weight to match with material 3

The padding is messed up again... look how the side padding is larger than vertical padding
I think what's most important is the colors. We don't have to try too hard to adapt the spacings, as that would take a lot of work

image

fb24m:
> > I decreased paddings to match with android and font weight to match with material 3
> 
> The padding is messed up again... look how the side padding is larger than vertical padding I think what's most important is the colors. We don't have to try too hard to adapt the spacings, as that would take a lot of work
> 
> image

for text, the horizontal padding should be larger than vertical one, but yes, right looks bad

fb24m:
image
image

end-4:
what app is that?

fb24m:
> what app is that?
> 

gnome tweaks

end-4:
image

now the toast has nearly no padding...

end-4:
fixed that last part; i'm merging now

image

PR #3135, opened 2026-03-24
Update Quick Settings dialogs with android 16 style

fb24m:
## Describe your changes

New Android 16–style design for Quick Settings dialogs. 
Color generation is still under discussion (#3133), so I’m open to any ideas or approaches. **Suggestions for translating into other languages are also welcome**
Any other suggestions or improvements are welcome as well

bluetooth
other


wfogame:
Cool

fb24m:
[43f6870](https://github.com/end-4/dots-hyprland/pull/3135/commits/43f687091baa779f73bdb827a6de724e36d87fb3) is broken, trying to solve the conflict

fb24m:
okay, now it works fine and is ready to go
also fixed the issues caused by my force pushes

omaku2006:
Superb!! Coool!!!

fb24m:
to solve conflict with [67fbaf1](https://github.com/end-4/dots-hyprland/commit/67fbaf1d27b30ff5aac097eabc8ff4e74ab28215)

MostafaSensei106:
cool

end-4:
- `Appearance.font.pixelSize.smaller` is too small. It makes sense for the sections in the Eye protection menu but isn't very easy on the eyes in other sections
- The padding of the bottom buttons is way too large

end-4:
Please also remove the horizontal line on the Wifi menu when it has finished scanning to be consistent with other popups

end-4:
This is Gmail. Two things we can learn from here:
- Wifi and Bluetooth entries could be in groups like these email conversations. I think for Bluetooth we can have a section of devices with recognizable names and a section of MAC address ones. For an example of how to use sections in a ListView, see `dots/.config/quickshell/ii/modules/waffle/startMenu/searchPage/SearchResults.qml`
- The section header should be left-padded (roughly) the same amount as the outer rounding of the section

image

fb24m:
> This is Gmail. Two things we can learn from here:
> - Wifi and Bluetooth entries could be in groups like these email conversations. I think for Bluetooth we can have a section of devices with recognizable names and a section of MAC address ones. For an example of how to use sections in a ListView, see `dots/.config/quickshell/ii/modules/waffle/startMenu/searchPage/SearchResults.qml`
> - The section header should be left-padded (roughly) the same amount as the outer rounding of the section
> 
> image
> 

do you mean wrapping them in bubbles like in gmail?

PR #3137, opened 2026-03-24
Gentoo: songrec update

jwihardi:
- ``songrec`` is nowin the ``guru`` overlay so I switched it over to that, instead of a local ebuild.


clsty:
Thanks for the update.

PR #3141, opened 2026-03-25
fix: set min height for Notification Group to fix visual bug on  low-res screens

vananh2801:
I have a 14" screen, 1920x1080, scale x1.2, when i edited quick toggles, the notification group was like in the attached screenshot. I guess the height seems equal to 0. 

image

### DIFF
--- a/dots/.config/quickshell/ii/modules/ii/sidebarRight/SidebarRightContent.qml
+++ b/dots/.config/quickshell/ii/modules/ii/sidebarRight/SidebarRightContent.qml
@@ -97,6 +97,7 @@ Item {
             }
 
             CenterWidgetGroup {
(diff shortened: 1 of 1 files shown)


vananh2801:
After fixing:
image

create-juicey-app:
Bandaid fix, how am i supposed to use the calendar & it's tools properly with this fix??
image

vananh2801:
so, did u have any fix for it? hmm, i have no idea

juiceydev:
maybe turning them into toggleable categories, ~~however i'm not sure if end4 gave us the components to do one.~~* actually yes they do on the calendar itself, maybe wrap the notifications too with it?
image

juiceydev:
sad

PR #3148, opened 2026-03-27
feat(ai): auto-resize input with scroll

LilFishyChan:
Add the sidebar ai's input box to the scrolling view to avoid overflowing underneath and making it unreadable when there's too much text.

image


end-4:
thanks
just one small issue of the send button being shaky

PR #3149, opened 2026-03-27
Fedora: adjust dotfiles to sync with dist-arch

fb24m:
## Describe your changes
@clsty 

What changed:

- updated Quickshell to sync with dist-arch
- added cpptrace (required by new quickshell)
- Fix local Quickshell installation (previously installed from COPR fallback) 

it looks good and works on my PC, unlike the previous version, so it's ready to merge


clsty:
Thank you for the update!

PR #3150, opened 2026-03-27
audio: prefer to show node description

moetayuko:
## Describe your changes

nickname looks like `USB Audio #1`, `USB Audio #2` when a device has multiple ports, which is not distinguishable.

ready


end-4:
For my case the nickname is better:

image
image

Maybe we can have something like this?
**NicknameInBold**: super mega long description device

moetayuko:
FYI this is my case:
图片图片

just wondering, is it possible to make the long text scroll horizontally?

PR #3158, opened 2026-03-28
background: Add audio visualizer widget

ParanoidExtreme:
## Describe your changes

Adds an audio visualizer widget to the background, rendered using cava. The implementation is inspired by the media controls widget but with its own distinct styling. Two display styles are available: bars and wave, both can be customized with opacity, border width, max height etc. Customization is exposed through several variables (target bar/point width, gap, smoothing, border, and more), these are configurable in BackgroundConfig like the other widgets are.

**Resource management:** cava + bar/wave rendering is automatically paused when all monitors are fullscreened, when a window is detected over the visualizer, or when gamemode is active.

**Performance:** includes a configurable performance limiter to cap rendering load, this applies only to wave which uses a canvas to render which happens to be quite GPU and CPU heavy. Bars hardly has any CPU usage and quite low GPU usage hence it doesn't need a performance limiter.

## Changed files

**New files:**
- `widgets/PinnedWidget.qml` - Base component for static (non-moving/anchored) widgets
- `widgets/visualizer/VisualizerWidget.qml` - The audio visualizer widget (bars and wave styles, performance limiter)

**Modified files:**
- `background/Background.qml` - Reworked layout to support static background widgets; wraps content in a scope, integrates the new visualizer widget, and adds gamemode-aware cava pausing + window/fullscreen/gamemode detection
- `settings/BackgroundConfig.qml` - Adds a configuration section with options for the audio visualizer (style, sizing, smoothing, border, performance limiter, etc.)
- `common/Config.qml` - Adds config entries for the visualizer (`enable`, `height`, `barRounding`, `smoothing`, `opacity`, `barSpacing`, `targetBarWidth`, `mono`, `showWhenLocked`, and more)

Mostly ready, happy with the current state but open to feedback on both the visual styling and the code structure.

## Showcase
https://github.com/user-attachments/assets/4c6d9d89-a788-459f-a968-044d313409e9


daisith:
Nice, great job !

ParanoidExtreme:
> Nice, great job ! Is it colored by matugen ?

Thanks!

Yeah it's just the primary color used in the rest of the quickshell widgets.

daisith:
i hope it will be merge, that's a missing feature.
thanks to make it happen ^^

fb24m:
there should be a "wave" mode, I think, like in the existing media widget in the bar
it would be more minimalistic and consistent

something like this maybe

image

ParanoidExtreme:
> there should be a "wave" mode, I think, like in the existing media widget in the bar it would be more minimalistic and consistent
> 
> something like this maybe
> image

I originally had this implemented but decided to remove it as I personally didn't think I made it well enough, I'll try it again though and see if I can manage to do it better this time.

ParanoidExtreme:
> there should be a "wave" mode, I think, like in the existing media widget in the bar it would be more minimalistic and consistent
> 
> something like this maybe
> image

About this... I decided to retry it and it turned out quite well. Sadly though rendering  a wave uses quite a bit more processing/rendering power than the bars so not sure how well it will run for others, but personally I see around 10% cpu (on a single thread) and 15-20% GPU rendering wave and under 10% for both when rendering bars. That's for a single 60hz monitor. Running both my secondary 60hz monitor and my primary 240hz monitor I see around 35-40% gpu usage and 40-ish% cpu (single thread). I don't always render the widget though so if one screen is fullscreened, nothing is rendered on it, and if both screens are covered there is basically 0 resource usage since I stop cava and the renderer. 

I could perhaps require the background to actually be in focus (no windows that aren't floating) for it to render as an optimization but I'm open to suggestion.


https://github.com/user-attachments/assets/4d7907c0-edfc-432e-b36b-2e35412dbd78

fb24m:
that's a good idea to add toggle hides visualizer when something is in focus (or even make it default option), since it will distract attention and, as you said, consume resources, but i'd also add a "Show only on primary monitor" toggle, to save resources and keep minimalistic look
you could also try reducing frame rate, i believe we don't need 240 fps for this widget

i'm not sure if this feature will be merged, but you could keep it as optional because i really like this

ParanoidExtreme:
> that's a good idea to add toggle hides visualizer when something is in focus (or even make it default option), since it will distract attention and, as you said, consume resources, but i'd also add a "Show only on primary monitor" toggle, to save resources and keep minimalistic look you could also try reducing frame rate, i believe we don't need 240 fps for this widget
> 
> i'm not sure if this feature will be merged, but you could keep it as optional because i really like this

Yeah that wouldn't be too hard to add, although you don't actually set the primary monitor in hyprland so it would probably just have to be a dropdown selector or so. I did attempt to reduce framerate yesterday, it didn't work very well given how quickshell works, it ended up looking terrible and flickering was a major issue. 

If you have any ideas on how to properly add framerate limiting, I'd love to hear them.

unixr77:
"wave" mode look better

ParanoidExtreme:
> that's a good idea to add toggle hides visualizer when something is in focus (or even make it default option), since it will distract attention and, as you said, consume resources, but i'd also add a "Show only on primary monitor" toggle, to save resources and keep minimalistic look you could also try reducing frame rate, i believe we don't need 240 fps for this widget

I added a helper function that checks for each monitor if it's covered by a window or not while managing to ignore floating windows. Works really really well it seems, cpu/gpu usage is basically not effected when it's not actively running. I also added toggles accordingly in settings. 

TODO: MAYBE add an option to lock it to a single monitor
TODO: Add fps cap to decrease GPU usage.
TODO: Look into a way of decreasing CPU usage for Wave canvas, bars are very efficent already.

https://raw.githubusercontent.com/ParanoidExtreme/Images/main/recording_2026-03-29_14.59.25.mp4

dots/.config/quickshell/ii/modules/ii/background/widgets/visualizer/VisualizerWidget.qml
```diff
@@ -0,0 +1,204 @@
...
+
+    Canvas {
+        id: waveCanvas
+        anchors.fill: parent
+        opacity: 0
+        visible: opacity > 0
+        
+        onWidthChanged: requestPaint()
+        onHeightChanged: requestPaint()
+
```
fb24m, line 123:
```suggestion

property int frames: 0

```

dots/.config/quickshell/ii/modules/ii/background/widgets/visualizer/VisualizerWidget.qml
```diff
@@ -0,0 +1,204 @@
...
+        onPaint: {
+            var ctx = getContext("2d");
```
fb24m, line 135:
```suggestion
            if (waveCanvas.frames >= 2) {
                waveCanvas.frames = 0;
            }
            else {
                waveCanvas.frames += 1
                return
            }
            
            var ctx = getContext("2d");
```

fb24m (commented):
i'd suggest updating chart every 3rd frame (just as an example)

this reduces cpu usage from 6-7 to 2-3 for me and looks smooth

ParanoidExtreme:
> i'd suggest updating chart every 3rd frame (just as an example)
> 
> this reduces cpu usage from 6-7 to 2-3 for me and looks smooth

Yes thank you, I'll make this a configurable performance limiter in the settings widget.

dots/.config/quickshell/ii/modules/ii/background/widgets/visualizer/VisualizerWidget.qml
```diff
@@ -0,0 +1,214 @@
+import QtQuick
+import QtQuick.Shapes
+import qs
+import qs.services
+import qs.modules.common
+import qs.modules.common.functions as CF
+import qs.modules.common.widgets
+import qs.modules.common.widgets.widgetCanvas
+import qs.modules.ii.background.widgets
+
```
fb24m, line 12:
```suggestion
import qs.modules.common.models.hyprland

```

dots/.config/quickshell/ii/modules/ii/background/widgets/visualizer/VisualizerWidget.qml
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    Timer { id: silenceTimer; interval: 1000; onTriggered: root.activityOpacity = 0 }
+
+    onPointsChanged: {
+        if (points.some(p => p > 0)) {
+            root.activityOpacity = 1.0;
+            silenceTimer.restart();
+        }
+    }
```
fb24m, line 240:
```suggestion
    }
    
    HyprlandConfigOption {
        id: enableAnimations
        key: "animations:enabled"
    }
```

dots/.config/quickshell/ii/modules/ii/background/widgets/visualizer/VisualizerWidget.qml
```diff
@@ -0,0 +1,214 @@
...
+        mode: configEntry?.mode ?? "wave",
+        waveFillOpacity: configEntry?.waveFillOpacity ?? 0.5,
+        waveBorderWidth: configEntry?.waveBorderWidth ?? 3
+    })
+
+    readonly property color waveFillColor: Qt.rgba(primaryColor.r, primaryColor.g, primaryColor.b, config.waveFillOpacity)
+    property var pixelHeights: []
+
+    height: config.height
+    opacity: (shown && baseVisibility) ? 1 : 0
```
fb24m:
```suggestion
    opacity: (shown && baseVisibility && enableAnimations.value) ? 1 : 0
```

fb24m (commented):
to respect "game mode"

ParanoidExtreme:
> i'd suggest updating chart every 3rd frame (just as an example)
> 
> this reduces cpu usage from 6-7 to 2-3 for me and looks smooth

Thanks for the logic, I added a variable to it which is controllable 

> to respect "game mode"

Thank you I will try this, I added a configuration item for the performance limiter, max is 3x currently which I think anything below looks like a slideshow haha. But yeah it helps CPU and GPU usage a lot. Probably useful for laptops and such with iGPU's. On my 240hz monitor, running half speed lowers cpu usage by around half (one thread) and gpu by around half as well, while looking almost identical to full speed.

fb24m:
> > i'd suggest updating chart every 3rd frame (just as an example)
> > this reduces cpu usage from 6-7 to 2-3 for me and looks smooth
> 
> Thanks for the logic, I added a variable to it which is controllable
> 
> > to respect "game mode"
> 
> Thank you I will try this, I added a configuration item for the performance limiter, max is 3x currently which I think anything below looks like a slideshow haha. But yeah it helps CPU and GPU usage a lot. Probably useful for laptops and such with iGPU's. On my 240hz monitor, running half speed lowers cpu usage by around half (one thread) and gpu by around half as well, while looking almost identical to full speed.

yes, updating every 3rd frame is perfect for me too

as for game mode, some users may prefer to always use the smoothest wave, even when there're tiled windows, so this option could be helpful

by the way, you can mention a co-author by adding `Co-authored-by: NAME <NAME@EXAMPLE.COM>`, for example `Co-authored-by: fb24m <hello@fb24m.ru>`, in the commit message (not the title)

dots/.config/quickshell/ii/modules/settings/BackgroundConfig.qml
```diff
@@ -614,4 +614,275 @@ ContentPage {
...
+                        text: Translation.tr("Hide when fullscreen/maximized")
+                        checked: Config.options.background.widgets.visualizer.hideWhenFullscreen
+                        onCheckedChanged: {
+                            Config.options.background.widgets.visualizer.hideWhenFullscreen = checked
+                            if (!checked) {
+                                Config.options.background.widgets.visualizer.hideWhenCovered = false
+                            }
+                        }
+                        StyledToolTip {
+                            text: Translation.tr("Hide the visualizer when a window is in fullscreen or maximized mode. (Reccomended)")
```
fb24m:
```suggestion
                            text: Translation.tr("Hide the visualizer when a window is in fullscreen or maximized mode. (Recommended)")
```

fb24m (commented):
typo

ParanoidExtreme:
> by the way, you can mention a co-author by adding `Co-authored-by: NAME <NAME@EXAMPLE.COM>`, for example `Co-authored-by: fb24m <hello@fb24m.ru>`, in the commit message (not the title)

Thanks, I mostly just use `git commit -m "message"` but I should probably start writing in the commit body as well.

 I improved the check to stop rendering when gamemode is active by also adding the check to background.qml so the cava process is stopped.

fb24m:
> > by the way, you can mention a co-author by adding `Co-authored-by: NAME <NAME@EXAMPLE.COM>`, for example `Co-authored-by: fb24m <hello@fb24m.ru>`, in the commit message (not the title)
> 
> Thanks, I mostly just use `git commit -m "message"` but I should probably start writing in the commit body as well.
> 
> I improved the check to stop rendering when gamemode is active by also adding the check to background.qml so the cava process is stopped.

```bash
git commit -m "message

other strings"
```

via shift+enter is also ok

dots/.config/quickshell/ii/modules/common/Config.qml
```diff
@@ -209,6 +209,24 @@ Singleton {
...
+                        property real smoothing: 0.05
+                        property real opacity: 1
+                        property real barSpacing: 10 // in pixels
+                        property int targetBarWidth: 50 // in pixles, rough target
+                        property bool mono: true
+                        property string mode: "bars" // "bars" or "wave"
+                        property real waveFillOpacity: 0.5
+                        property real dataSmoothing: 0.5
+                        property int waveBorderWidth: 3    // 0 = no border
+                        property int renderEveryXFrames: 1  // 1 = every frame, 2 = every other frame, etc. Only for "wave" mode
```
fb24m:
```suggestion
                        property int renderEveryXFrames: 2  // 1 = every frame, 2 = every other frame, etc. Only for "wave" mode
```

dots/.config/quickshell/ii/modules/settings/BackgroundConfig.qml
```diff
@@ -614,4 +614,275 @@ ContentPage {
...
+                        },
+                        {
+                            displayName: Translation.tr("Half speed"),
+                            icon: "directions_run",
+                            value: 2
+                        },
+                        {
+                            displayName: Translation.tr("Low speed"),
+                            icon: "directions_walk",
+                            value: 3
```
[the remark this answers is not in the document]
fb24m: maybe 4 for low speed mode?

it's not so slideshow yet, but should be more efficient

fb24m:
i'm a bit confused about github reviews feature, so i'll just write here

i'd change low speed option value from 3 to 4, because it's not so slideshow, but should be more efficient

with a value of 4, my cpu usage is <1%

ParanoidExtreme:
> i'm a bit confused about github reviews feature, so i'll just write here
> 
> i'd change low speed option value from 3 to 4, because it's not so slideshow, but should be more efficient
> 
> with a value of 4, my cpu usage is <1%

Hmm, I think the gap would be a bit big going from 2 --> 4, perhaps I should add another setting (Efficency mode) or something OR I could just use a fully customizable value like I do elsewhere.

fb24m:
i think 1 -> 2 -> 4 would be perfect
the gap between 2 and 3 is so small that there're almost no difference
i also don't think we need a fully customizable value, since even 5 performs so bad

i think, low speed mode is only for those who wants to use the wave on an old cpu or whatever, so basically it's an efficiency mode

fb24m:
you could also consider renaming the performance modes I think, because calling them "speed" is not so accurate in our case

so they could be called like “smooth”, “balanced”, and “efficient” modes

ParanoidExtreme:
> you could also consider renaming the performance modes I think, because calling them "speed" is not so accurate in our case
> 
> so they could be called like “smooth”, “balanced”, and “efficient” modes

Alright, I'll give this a try.

ParanoidExtreme:
> you could also consider renaming the performance modes I think, because calling them "speed" is not so accurate in our case
> 
> so they could be called like “smooth”, “balanced”, and “efficient” modes

I think I fixed this up pretty nicely, I also added an "Auto" option which syncs with the current system power profile. Performance --> smooth, Balanced --> balenced and Power saver --> Efficency

fb24m:
yes, that's just perfect now!!

dots/.config/quickshell/ii/modules/settings/BackgroundConfig.qml
```diff
@@ -614,4 +614,280 @@ ContentPage {
...
+                        text: Translation.tr("Hide when fullscreen/maximized")
+                        checked: Config.options.background.widgets.visualizer.hideWhenFullscreen
+                        onCheckedChanged: {
+                            Config.options.background.widgets.visualizer.hideWhenFullscreen = checked
+                            if (!checked) {
+                                Config.options.background.widgets.visualizer.hideWhenCovered = false
+                            }
+                        }
+                        StyledToolTip {
+                            text: Translation.tr("Hide the visualizer when a window is in fullscreen or maximized mode. (Reccomended)")
```
fb24m:
```suggestion
                            text: Translation.tr("Hide the visualizer when a window is in fullscreen or maximized mode. (Recommended)")
```

fb24m: typo again

ParanoidExtreme: Oops, thanks for pointing it out!

fb24m:
image

there's a bottom border that looks inconsistent here

ParanoidExtreme:
> image
> 
> 
> 
> there's a bottom border that looks inconsistent here 

Ah yeah didn't notice this, probably because the border is rendered around the canvas, does it get better/worse depending on border width? I'll look into it once I get home.

fb24m:
> > image
> > there's a bottom border that looks inconsistent here
> 
> Ah yeah didn't notice this, probably because the border is rendered around the canvas, does it get better/worse depending on border width? I'll look into it once I get home.

i actually think there shouldn’t be a bottom border

ParanoidExtreme:
> i actually think there shouldn’t be a bottom border

Agreed, managed to solve it by drawing just the top border (sides had the same issue)

unixr77:
The performance on the lock screen is very bad. Is this just me or is it the same for everyone?

https://github.com/user-attachments/assets/b690e424-ac34-4032-a9dd-b84779b24f7e

fb24m:
> The performance on the lock screen is very bad. Is this just me or is it the same for everyone?
> 
>  recording_2026-03-30_18.08.44.mp4

try switching to "Smooth mode"

unixr77:
GPU usage crazy %95+
screenshot-2026-03-30_18 17 01

fb24m:
> GPU usage crazy %95+ screenshot-2026-03-30_18 17 01

then try "Performance mode"
if it doesn't help, the GPU is likely the bottleneck

fb24m:
you could also try increasing the "Point width"

ParanoidExtreme:
> GPU usage crazy %95+ screenshot-2026-03-30_18 17 01

Interesting, is it just in "wave" mode or does "bars" also underperform? Asking as bars is much less GPU intensive. My school laptop ran extremely well with about 30% iGPU usage or so. What CPU/GPU do you have? Personally my laptop uses an Intel Core i5-1335U + Intel Iris Xe Graphics

unixr77:
@ParanoidExtreme  
using intel iGPU ( HD 630 ) no have discreate GPU
bar mode gpu usage %84 Still usage is too high

ParanoidExtreme:
> @ParanoidExtreme using intel iGPU ( HD 630 ) no have discreate GPU bar mode gpu usage %84 Still usage is too high

Mhmm I see, I could try putting my config on an older system I have laying around to see how horrible it is.

ParanoidExtreme:
> @ParanoidExtreme using intel iGPU ( HD 630 ) no have discreate GPU bar mode gpu usage %84 Still usage is too high

Mhmm, so a quick follow up on this. I decided to try it on my homelab server, which has an i5-4590t (2014), meaning it's like 12 years old. And with it I see around 70-80% gpu usage which is high of course but pretty acceptable CPU usage. Despite the high GPU usage it actually ran extremely smooth and looked really nice (basically no lag). It ran fine on the lockscreen as well, so aside from the high GPU usage (which is not optimal of course) it wasn't as bad as I expected. 

If you have any ideas on how to decrease GPU usage (aside from the framerate cap) feel free to share them.

fb24m:
> > @ParanoidExtreme using intel iGPU ( HD 630 ) no have discreate GPU bar mode gpu usage %84 Still usage is too high
> 
> Mhmm, so a quick follow up on this. I decided to try it on my homelab server, which has an i5-4590t (2014), meaning it's like 12 years old. And with it I see around 70-80% gpu usage which is high of course but pretty acceptable CPU usage. Despite the high GPU usage it actually ran extremely smooth and looked really nice (basically no lag). It ran fine on the lockscreen as well, so aside from the high GPU usage (which is not optimal of course) it wasn't as bad as I expected.
> 
> If you have any ideas on how to decrease GPU usage (aside from the framerate cap) feel free to share them.

I think you should move away from canvas, but have no idea how to replace it

ParanoidExtreme:
> I think you should move away from canvas, but have no idea how to replace it

Yeah, I think it's mostly fine for now but I'll def have to look into that once I've got more time, maybe later this week/next week.

unixr77:
> i5-4590t

ihave i7-9700K 
A more up to date processor compared to yours
but my screen resolution is 1440x2560
That's why GPU usage is high
I think it will perform better on a 1080 monitor

ParanoidExtreme:
> > i5-4590t
> 
> ihave i7-9700K A more up to date processor compared to yours but my screen resolution is 1440x2560 That's why GPU usage is high I think it will perform better on a 1080 monitor

Fair enough I'm using a 1080p monitor for my tests, although I can drive 2 screens with the visualizer which is about the same as a 1440p monitor. I see maybe around 40-50% then using a rtx 2060 dedicated GPU.

BharathBala21 (commented):
Works Perfectly well

kepalakubik:
How about enable visualizer only on lockscreen?

PR #3160, opened 2026-03-29
Russian localization improvements

fb24m:
## Describe your changes

Removed some unnecessary reductions (see screenshots)
The font settings for numbers that extend beyond the area have been reduced (see screenshots again)

image

image

image

image


tokKurumi (approved):
Looks good to me.
I don't like the practice of truncating words in the UI and adding a trailing period instead of using full wording. I appreciate that you removed these unnecessary *ugly* dots.

fb24m:
> Looks good to me.
> I don't like the practice of truncating words in the UI and adding a trailing period instead of using full wording. I appreciate that you removed these unnecessary *ugly* dots.

I agree

clsty:
It seems that you have different opinions about the translation with the previous contributor.

I do not understand your language so normally I'll just accept the new one. If the disagreement deepens in the future, we will reconsider.

Thanks for the update.

fb24m:
> It seems that you have different opinions about the translation with the previous contributor.
> 
> I do not understand your language so normally I'll just accept the new one. If the disagreement deepens in the future, we will reconsider.
> 
> Thanks for the update.

what do you mean?

clsty:
> > It seems that you have different opinions about the translation with the previous contributor.
> > 
> > I do not understand your language so normally I'll just accept the new one. If the disagreement deepens in the future, we will reconsider.
> > 
> > Thanks for the update.
> 
> what do you mean?

e.g. https://github.com/fb24m/dots-hyprland/commit/fc17e29ebf8ae1481900e773836d654c44047695

I do not understand Russian language so I can not decide which is better, the old one or new one. Normally I just accept the newer one.

There's possibility that the previous contributor may revert your patch again. If that happens then we shall reconsider, instead of blindly accepting newer ones.

fb24m:
> > > It seems that you have different opinions about the translation with the previous contributor.
> > > I do not understand your language so normally I'll just accept the new one. If the disagreement deepens in the future, we will reconsider.
> > > Thanks for the update.
> > 
> > 
> > what do you mean?
> 
> e.g. [fb24m@fc17e29](https://github.com/fb24m/dots-hyprland/commit/fc17e29ebf8ae1481900e773836d654c44047695)
> 
> I do not understand Russian language so I can not decide which is better, the old one or new one. Normally I just accept the newer one.
> 
> There's possibility that the previous contributor may revert your patch again. If that happens then we shall reconsider, instead of blindly accepting newer ones.

Some of my changes are truly necessary (for example, declension agreements), others don't change the essence of words, but remove abbreviations, which the previous participant also did. However, there are also controversial issues regarding the fact that long Russian words don't fit into some settings lines, so this is still a subject for discussion. Still, as I already said, some changes are truly necessary, and in any case, we will find a compromise

fb24m:
@xleb-normal what do you think?

xleb-normal:
This looks good! Looks like I forgot to double-check the classic clock style settings for my translation. 😅

Issue #2803, opened 2026-01-02
No doas support


UcraNerd:
### What would you like to be added?

Support for doas instead of sudo -v

<img width="588" height="145" alt="Image" src="https://github.com/user-attachments/assets/d2bee194-e436-474a-a8bb-0dc1f3bc7520" />

relativemodder:
who tf uses doas bro 😭

K0-RR:
I use, so that's at least two people interested in minimizing the reliance on `sudo` in this single issue.
_Besides, why do you even comment if it doesn't concern you?_

K0-RR:
While this issue could be resolved on the end user side by using https://aur.archlinux.org/packages/doas-sudo-shim-v or editing the lines
https://github.com/end-4/dots-hyprland/blob/d5b599da3de8e879745945e5e74de74c32a6e23d/sdata/lib/functions.sh#L101
https://github.com/end-4/dots-hyprland/blob/d5b599da3de8e879745945e5e74de74c32a6e23d/sdata/lib/functions.sh#L111

Maybe just **replace** `sudo -v` **with** `sudo true`? That would get around other privilege escalation tools missing that sudo option.

Probably every doas / [pleaser](https://gitlab.com/edneville/please) / [dosr](https://github.com/LeChatP/RootAsRole) / run0 user has a symlink at the sudo's original location, so that would be enough.

PR #3167, opened 2026-03-30
 support for sudo alternatives

marwuint:
## Describe your changes
replace incompatible `sudo -v` flag with true, based on https://github.com/end-4/dots-hyprland/issues/2803

ready :)


clsty:
Well, this does no harm anyway so I'm merging it. Thanks for contribution.

PR #3176, opened 2026-04-01
gentoo: removed oneui

jwihardi:
removed it 😄


clsty:
Thanks for the update!

PR #3177, opened 2026-04-01
Edit screen snip keybind to only take a screenshot of the focused monitor

lunstia:
I edited the command exec to parse the monitors and find the focused monitor to pass into grim. Before, any user with more than one monitor would have all their monitors included in the screen snip which isn't ideal.


end-4:
Why did you include this selection? Is the focused monitor null after something's disconnected or something?

```sh
"\(.x),\(.y) \(.width)x\(.height)"')"
```

I find it much elegant to do the whole monitor selection like this, am I missing something? 
```sh
-o "$(hyprctl activeworkspace -j | jq -r '.monitor')"
```

lunstia:
> Why did you include this selection? Is the focused monitor null after something's disconnected or something?
> 
> ```shell
> "\(.x),\(.y) \(.width)x\(.height)"')"
> ```
> 
> I find it much elegant to do the whole monitor selection like this, am I missing something?
> 
> ```shell
> -o "$(hyprctl activeworkspace -j | jq -r '.monitor')"
> ```

Nope, that's way better. Somehow missed the -o when looking at the man page. I tunnel visioned after seeing -g, my bad. I also like your solution way better

PR #3183, opened 2026-04-03
(FIX TO "INVALID MODEL") Convert model names to lowercase in Ai.qml 

XChaosSaMa:
Fix Ollama model validation by applying safeModelName normalization and lowercasing consistently when both storing and looking up model keys, so that models with colons (e.g. phi4:latest) or mixed-case names (e.g. Qwen3-Coder-30B) are correctly recognized instead of throwing 'Invalid model'.

### DIFF
--- a/dots/.config/quickshell/ii/services/Ai.qml
+++ b/dots/.config/quickshell/ii/services/Ai.qml
@@ -485,7 +485,7 @@ Singleton {
 
     function setModel(modelId, feedback = true, setPersistentState = true) {
         if (!modelId) modelId = ""
-        modelId = modelId.toLowerCase()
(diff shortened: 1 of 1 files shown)


end-4:
Thanks. Why not add the lowercasing to `safeModelName` though?

XChaosSaMa:
XD, you are right.  I should have put it there from the start, I'll update the PR

end-4:
wait why undo the setModel part

end-4:
i think it's fine to leave the normalization there

XChaosSaMa:
My bad, i used the last file and forgot to change it back, it's fixed now

PR #3184, opened 2026-04-03
FIX: Screenshot regions alpha value + improvements

Manriel:
This commit fixes an issue with the alpha channel (transparency) not displaying correctly in the region selector when taking screenshots.

### Data Type:
- Changed the contentRegionOpacity property type from Boolean to numeric (real), as described in the configuration
- This allows for the correct application of transparency values ​​in the range from 0 to 1

### Coordinate Handling:
- Added scaling of detected object coordinates based on the monitor scale
- Coordinates are divided by the monitor scale factor for correct positioning

### Region Display:
- Replaced the opacity property with regionAlpha for better control over target region transparency
- This provides more precise and consistent control over border and fill transparency regardless of region content

### Region Information:
- Added display of the window title in the region text label along with the application class name

### Transparency Management:
- Added a new regionAlpha property to replace the opacity property for managing region transparency
- Reworked Transparency application logic — the alpha channel is now applied to the fill and borders, but not to the region content.

### Visual improvements:
- Transparency is now applied only to the region background, not to the title and icon, nor to the region coordinates.
- Icon display has been fixed — the application class icon is now displayed correctly, ignoring the background transparency, which made it unreadable at high transparency values ​​(which is typical).
- The structure of the region information panel has been changed.
- Region coordinates and dimensions (x,y width x height) are now displayed.

### DIFF
--- a/dots/.config/quickshell/ii/modules/ii/regionSelector/RegionSelection.qml
+++ b/dots/.config/quickshell/ii/modules/ii/regionSelector/RegionSelection.qml
@@ -51,7 +51,7 @@ PanelWindow {
     property color imageFillColor: ColorUtils.transparentize(imageBorderColor, 0.85)
     property color onBorderColor: "#ff000000"
     property real targetRegionOpacity: Config.options.regionSelector.targetRegions.opacity
-    property bool contentRegionOpacity: Config.options.regionSelector.targetRegions.contentRegionOpacity
+    property real contentRegionOpacity: Config.options.regionSelector.targetRegions.contentRegionOpacity
 
     // Vars for indicators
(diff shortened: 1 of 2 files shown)


end-4:
Extremely long description, vague and does not make sense:

> Replaced the opacity property with regionAlpha for better control over target region transparency

Makes windows look extremely weird and break content region

image

Come back when you've understood and fixed what the transformer gave you

Manriel:
Nothing is broken. It depends on your config.json.
It shuold look like this.
2026-04-07-182129_hyprshot

I made changes to the default config, which make it look like the screenshot.
Coordinates display became configurable also.

PR #3187, opened 2026-04-04
Improvement: Scrolling statusText on quickToggles

Leo-Aqua:
Adds a little scrolling animation to the status text of quick toggles. This is better than truncating, as it makes everything visible to the user without sacrificing more space.


tokKurumi:
> Adds a little scrolling animation to the status text of quick toggles. This is better than truncating, as it makes everything visible to the user without sacrificing more space.

Agree! It would also resolve debates about translations: either use short truncated form or full form

PR #3193, opened 2026-04-06
feat: allow users to authenticate with the Gelbooru API

GazzD:
## Describe your changes

Gelbooru's API supports optional authentication via `api_key` and `user_id` query parameters, which unlocks full API access and removes anonymous rate limits.

This PR adds two new config options `sidebar.booru.gelbooru.userId` and `sidebar.booru.gelbooru.apiKey`  that, when set, are automatically included in the Gelbooru API request. Credentials are optional; unauthenticated usage still works as before.

A contextual note is also shown when switching to the Gelbooru provider, pointing users to [Gelbooru API documentation page](https://gelbooru.com/index.php?page=wiki&s=view&id=18780)  to obtain the credentials.

The functionality is ready, but it is also open to feedback if something needs to be changed. 

## Screenshots

Params added to `config.json`
image

Sidebar with contextual note
image


GazzD:
This PR should solve this reported issue: https://github.com/end-4/dots-hyprland/issues/1534

Satoxyan:
succesfully fetched, but thumbnail wasn't appeared

nonetrix:
I think this is similar to my PR https://github.com/end-4/dots-hyprland/pull/2520 but I used Keyring like Gemini API key etc.

PR #3217, opened 2026-04-10
Add bar shadow option for all bar types

altrup:
## Describe your changes

- use `layer.effect` for shadows on all bar types
- remove old `StyledRectangularShadow` which was for floating only
- add option in settings to toggle (default true)

image
<br><br>
image image image

Seems good to me, the only thing I'll say is that the shadow basically does nothing on dark mode


end-4:
seems to work, but it's barely different so it doesn't really "work"

end-4:
also, it currently blocks off the transparency. we should apply opacity on the shadow and the bar background as a whole to prevent that

altrup:
> seems to work, but it's barely different so it doesn't really "work"

Yeah, main difference is that now the other bar types can also have shadows

I will try and fix the transparency issue though

EDIT: transparency is now fixed, the bar itself should be identical whether or not you have bar shadow on

PR #3221, opened 2026-04-11
Add Bluetooth connect feedback

### LINKED ISSUE #3220: Bluetooth connect button has no pending feedback

A small fixed-size animated indicator beside the Connect/Disconnect button would improve feedback without causing layout shift.

### How will it help?

When clicking Connect in the Bluetooth sidebar, there is no visible pending state while the connection is being attempted. This makes it unclear whether the click registered or whether the connection is still in progress.

### Extra info

_No response_

buzzdead:
## Describe your changes

Adds a pending feedback state to the Bluetooth device connect/disconnect button.

  - Keeps the button label stable to avoid layout shift
  - Shows a fixed-width pulsing Bluetooth icon while connect/disconnect is pending
  - Disables the button while the request is pending
  - Clears the pending state when the device state changes or after a timeout

Fixes #3220 

Ready for review (Solution provided by AI)

### DIFF
--- a/dots/.config/quickshell/ii/modules/ii/sidebarRight/bluetoothDevices/BluetoothDeviceItem.qml
+++ b/dots/.config/quickshell/ii/modules/ii/sidebarRight/bluetoothDevices/BluetoothDeviceItem.qml
@@ -1,5 +1,4 @@
 import qs.modules.common
-import qs.modules.common.functions
 import qs.modules.common.widgets
 import qs.services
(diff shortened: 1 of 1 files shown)


buzzdead:
One thing I did not fix was layout shift that happens when finally connected and "Forget" appears next to "Disconnect". But it doesn't always happen because the item dropdown usually folds before it happens.

PR #3224, opened 2026-04-12
Update install script to use prebuilt RPMs on Fedora

fb24m:
## Describe your changes

This PR reworks how install script installs packages that are missing in the fedora or copr repos

### Previous behavior
every package builds using .spec 

**Pros**: 
- dotfiles aren't dependent on fedora & copr repos 
- version of any package can be updated just by editing a few strings 

**Cons**: 
- installation takes too long... 
- ...and might fail if user doesn't have enough RAM 

_or/and_

- installation takes tens of minutes, if user has an older CPU 

### New behavior
all packages already prebuilt as a RPM 

**Pros**: 

- dotfiles aren't dependent on fedora & copr repos 
- installation takes a few seconds (just like normal installation from common repos) 

**Cons**:
- to update a package (it's matugen and quickshell for now), contributor needs to build a new RPM on their machine

But even if an RPM file is broken or some contributor is unable to prebuild it, install script checks version of RPM & version of spec file and if they are not equal, it rebuilds the RPM

Why not use the copr repo? We'll lose the main benefit of open source, when anyone can update anything, If i create a COPR repo, the whole install script would be dependent on it.

It works far better and doesn’t break even on low-end machines, so yes, i think


clsty:
Well, good point but no, we're not having package binaries in this Git repo.

fb24m:
> Well, good point but no, we're not having package binaries in this Git repo.

maybe then we shouldn't rebuild and reinstall packages, if they're up to date?
because installation takes too long and script is doing unnecessary work

clsty:
> maybe then we shouldn't rebuild and reinstall packages, if they're up to date?

If you'd like to add logic to prevent unnecessary rebuild and reinstallation, this is acceptable.

Another acceptable thing is to use GitHub Actions to automatically build packages and provide them in GitHub Releases. See https://github.com/clsty/arch-build for a quick example, though this repo is far from automated and is unmaintained.

Issue #3075, opened 2026-03-06
Dock preview not showing up. after last quickshell update


unixr77:
### Step 1. Before you submit

- [x] I have read the [Troubleshooting](https://ii.clsty.link/en/ii-qs/04troubleshooting/) and [Usage](https://ii.clsty.link/en/ii-qs/02usage/) pages.
- [x] I've successfully updated to the latest version following the [guidance](https://ii.clsty.link/en/ii-qs/01setup/#updating).
- [x] I've successfully updated the system packages to the latest.
- [ ] I've ticked the checkboxes without reading their contents

### Step 2. Quick diagnose info

<details><summary>Quick diagnose</summary>

```

/========================\
| Checking git repo info |
\========================/

[===diagnose===] git remote get-url origin
https://github.com/end-4/dots-hyprland
[---SUCCESS---]

[===diagnose===] git rev-parse HEAD
1dcf90ac574caffa1626da0835f0ded3525a1bd9
[---SUCCESS---]

[===diagnose===] git status
On branch main
Your branch is up to date with 'origin/main'.

nothing to commit, working tree clean
[---SUCCESS---]

[===diagnose===] git submodule status --recursive
 e31ec4cb4ebf6a46b267f5c42eabf6874916fa16 dots/.config/quickshell/ii/modules/common/widgets/shapes (remotes/origin/HEAD)
[---SUCCESS---]

/=================\
| Checking distro |
\=================/

[===diagnose===] _check_distro_id
distro ID: arch
distro ID_LIKE: 
[---SUCCESS---]

[===diagnose===] cat os-release
cat: os-release: No such file or directory
[---EXIT 1---]

/====================\
| Checking variables |
\====================/

[===diagnose===] declare -p XDG_CACHE_HOME
./diagnose: line 36: declare: XDG_CACHE_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p XDG_CONFIG_HOME
./diagnose: line 36: declare: XDG_CONFIG_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p XDG_DATA_HOME
./diagnose: line 36: declare: XDG_DATA_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p XDG_STATE_HOME
./diagnose: line 36: declare: XDG_STATE_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p ILLOGICAL_IMPULSE_VIRTUAL_ENV
[...]

PR #3228, opened 2026-04-13
Fix dock preview not showing up after first hover

wbbwbwbwbwbb:
## Describe your changes

This fixes **https://github.com/end-4/dots-hyprland/issues/3075**
The preview popup would show on the first hover but stop showing on any subsequent hovers without restarting quickshell. This should fix it by keeping the wayland surface mapped during fade out and adding an explicit handler for rehovering the same button.

This only changes how the hovering is handled in DockApps.qml and fixes the issue completely for me.

### DIFF
--- a/dots/.config/quickshell/ii/modules/ii/dock/DockApps.qml
+++ b/dots/.config/quickshell/ii/modules/ii/dock/DockApps.qml
@@ -17,14 +17,20 @@ Item {
     property real windowControlsHeight: 30
     property real buttonPadding: 5
 
-    property Item lastHoveredButton
+    property Item lastHoveredButton: null
     property bool buttonHovered: false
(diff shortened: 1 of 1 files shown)


marepallisanthosh999333:
also can you fix scratchpad or special workspace apps do not show up in dock apps. Right now even if i send an app to scratchpad or special workspace they are showing up in the dock while they are completely hidden in bar workspaces

image

unixr77:
working fine

daisith:
Thanks for your work, it fixes the issue like a charm.

end-4:
Wrong alignment: the title should stay at the top, only the window preview should be vertically centered

image

end-4:
nevermind, that's my fault not this PR

end-4 (approved):

PR #3239, opened 2026-04-15
Fix broken waifu.im image and tag search

Kashi754:
## Describe your changes

Updated the Booru.qml to fix broken waifu.im tag an image search. Search requests now align with current waifu.im api.

Tested and everything works as expected now.


Kashi754:
Completed more testing and realized tags were not being correctly applied to the search request params, so made corrections.

end-4:
thank you. i fixed that one last bit that was preventing this from working on my end

PR #3244, opened 2026-04-18
ID failing to recognize

jwihardi:
Not sure why but the ID and ID_LIKE seems to be stripping with _"_ instead of _'_ now. This should work for both.


clsty:
LGTM, thanks for the fix.

PR #3254, opened 2026-04-24
fix Hyprsunset default color temperature

Reluckycf:
## Problem
The "Night Light" button toggle in the sidebar was consistently appearing in the "On" state when restarting a Hyprland session. This occurred because hyprsunset now defaults to 6000K instead of 6500K ([See the docs](https://wiki.hypr.land/Hypr-Ecosystem/hyprsunset/#:~:text=00%3A00-,temperature,6000,-gamma)). In the `fetchState` logic, the system temperature was compared against `6500` and since `6000 != 6500`, the toggle button thought it was "Active".
## Fix
- Added a `defaultColorTemperature` variable to the Hyprsunset service with the new default color temp. 
- Updated the fetchState logic in `Hyprsunset.qml` to validate against this new default (6000K)

Hope it works. 
Any feedback would be appreciated.


end-4 (approved):

end-4:
so that's why... thanks

PR #3259, opened 2026-04-28
Update version tag to 4.1.0 in matugen.spec

MVN123123123123123:
do i even need to say anything? matugen spec file is outdated, that's it.

### DIFF
--- a/sdata/dist-fedora/SPECS/matugen.spec
+++ b/sdata/dist-fedora/SPECS/matugen.spec
@@ -1,11 +1,10 @@
-%global commit      e65259d68edc034905da477b6c1a349e89e2aa8d
+%global commit      4112d352914742ba69f6380fd07984adba02d376
 %global shortcommit %(c=%{commit}; echo ${c:0:7})
-%global commits     719
-%global snapdate    20260213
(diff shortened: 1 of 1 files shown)


clsty:
Yeah that's valid according to https://github.com/InioX/matugen/tags . Please test it to make sure it actually works and then the PR will be ready to merge.

MVN123123123123123:
idk how to prove that it works. this is the video that i tries some basic color testing, include testing the colour generation. good enough ?

[https://drive.google.com/file/d/1QUEMmAK2Pxi3IfDtHYya766tUdBYI-Xi/view?usp=sharing](url)

here's the installation video, there might be some ramdom things i do in the video, just don't mind abt it
[https://drive.google.com/file/d/1u90X68x31EVbtqlrH458fqPZr-J0RUKR/view?usp=sharing](url)

clsty:
Well, we here normally do not require you to prove whether it works. You apply the change and test running it, you observe and think it's ok, and you tell the truth that you actually believe, then that's ok.

Since end-4 and I do not use Fedora, we rely on Fedora users to maintain the Fedora support.

Thanks for the update, I'll merge it.

PR #3262, opened 2026-04-29
fedora: remove hyprland-qt-support from SPECS

fb24m:
## Describe your changes

Removed hyprland-qt-support from SPECS for Fedora. It is now installed from sdegler/hyprland copr repo ensuring compatibility with the hyprland version from the same repo. It also fixes the bug when dnf updates this package to newer version, but script rolls it back again.

it works so yes


clsty:
Thanks for the update.

PR #3269, opened 2026-05-02
conf to lua change

Linuxius-Thomas:
I have copied the old confs and translated them the best I could to the new syntax.

I tested it, catchall keysim changed to much might need to use alternative, also submaps changed might need to put all keybinds and custom keybinds in a submap.
But I'd appreciate feedback if you spot a mistake I made.
some parts I have completely commented or add a explanations/comments where I couldn't find the correct fix.


Linuxius-Thomas:
I just tried launching with the configs and I got a few errors as expected, but then I fixed them.
so now it is at the point hyprland gives no error but the quickshell elements wont show neither will the background.
It might be because I had to install hyprland-git or in the quickshell files it crashes because it can't deal with the lua files yet?
I haven't looked into it yet, but this is the current standing.

Linuxius-Thomas:
It now completely opens with everything working at first glance. the problem was the way the variables are shared. Now it stays in the lua files so none of the variable's are visible with "echo $variable" so I changed them to be enviroment variable using hl.env("variable", "value") which made everything work again, but you can also decide to use lua variables you just have to change all the $ variables in the command and keybinds, but I didn't do that because I don't know if the dots standard variable are also being used in the qml files.

So to summarise the lua confs fully work at first glance, even the cheatsheet layout works

Linuxius-Thomas:
Catchall/submap has changed alot so that is one of the things that does not work.

end-4:
I'll take a look when Hyprland 0.55 comes. We're not in a rush and I can't be bothered to switch to -git now

Linuxius-Thomas:
I understand I wouldn't try it either on my main os, but I have a test environment and was curious.

Linuxius-Thomas:
I have changed some things. 
Im still working to remove scripts call from the conf, because some are not working/deprecated. 

zoom works but their is currently a bug in the new hyprland that doesnt update conf variable after the conf has been loaded. 

I also cleaned up the code, by introducing variables and for loop, functions and if statements as that is now possible. 

just pressing super does nothing I dont know if that is because I cant get catchall to work or because their is deprecated hyprctl commands in the qml files.

Linuxius-Thomas:
I think I have change all that I can and most things work as they should, in the hyprland config files that is at least.
You might want to look at the following parts as I think you know more about it:

In /hyprland/colors.lua:
I commented the plugin hyprbars plugin as I don't know how it changed its syntax.

In /hyprland/general.lua:
I commented a part of the gestures as I can't test it because I don't have a track pad or device with one.
And I can't seem to find clarification on it's use in the old or new syntax.
Also under the shadow config part ignore_window seems to be deprecated
and under the misc config part vfr seems to be deprecated.

In /hyprland/keybinds.lua:
the exec command setting submap to global seems to have become obsolete, so I commented it.
I moved the catchall part to the top and commented it as I can't find what isn't working and also I can't find what it should do.
For the scripts I have replaced the scripts that stopped working for in config functions.
"The zoom is currently not working because of a bug in hyprland. But what I made does change the variable which I confirmed via hyprctl in a terminal." --edit it is working now 🥳
all other keybinds work as they should.

In /hyprland/rules.lua:
I commented the line that was already partially commented in the .conf variant. the line I'm talking about is the one with --wlogout.

In /hyprland/variables.lua:
I didn't change the 'dontLoad' variables into env's, because I don't really know how you wan't to use them.

end-4:
Updated Hyprland just now, taking a look. I didn't expect with the lua change it would break IPC too. Goddammit

dots/.config/hypr/hyprland/general.lua
```diff
@@ -1,39 +1,63 @@
...
+hl.gesture({
+    fingers = 4,
+    direction = "horizontal",
+    action = "workspace"
+})
+hl.gesture({
+    fingers = 4,
+    direction = "up",
+    action = function()
+        -- For whatever reason hl.dsp.global doesn't work here???
```
Linuxius-Thomas: You have too do hl.dispatch(hl.dsp.global())

end-4:
That took one evening, but without your PR it would've been two. So thanks a lot. 

Some notes:
- I removed the "dontLoad..." variables. I'm unfamiliar with Lua so for now I'll skip the advanced stuff
- For the zoom being reset on config reload, I'll let that slide, since that extremely rarely happens, if ever
- Idk if this was caused by some automated batch replacement, but FYI shell variables don't escape in single quotes
- The catchall was previously used to prevent the super-only keybind release event to go through to quickshell even when another key has been pressed before that release. It's now fixed so this is indeed no longer necessary
- The shell needs fixes too, but that's beyond the scope of this PR

dots/.config/hypr/hyprland/general.lua
end-4: thanks, fixed in 9f4afde0c5a7ffa56823a944a7196e4d4ccfec54

end-4:
okay i realized why i needed catchall now, without it we wouldn't be able to have the app icons dodging the numbers on the bar... now they always dodge once super key has been touched

moetayuko:
leftovers
https://github.com/end-4/dots-hyprland/blob/6c041b953ab05cc6162f54cdb2dd5357e7190c22/sdata/subcmd-install/3.files-exp.yaml#L45-L49

VietNguyenVN:
@clsty sorry for ping but you might want to change the install/update script a bit since monitors.conf and workspaces.conf do not exist anymore

> leftovers
> 
> https://github.com/end-4/dots-hyprland/blob/6c041b953ab05cc6162f54cdb2dd5357e7190c22/sdata/subcmd-install/3.files-exp.yaml#L45-L49

clsty:
I'm currently busy but willing to review related PR.

PR #3273, opened 2026-05-04
fedora: Install packages right after building

fb24m:
## Describe your changes

I just noticed an error message while building [here](https://github.com/end-4/dots-hyprland/pull/3259#issuecomment-4336618541). The build fails because the script tries to build all packages and then installs them. So installation worked if executed it a few times, but failed for the first one—that's why I didn't notice any problems before. Now, the script installs every package right after it's built so we are able to install cpptrace before quickshell is built

it works fine on a clean Fedora Everything installation


clsty:
Thanks for the update!

PR #3281, opened 2026-05-06
Avoid spawning qs ipc clients from Quickshell UI actions

samscarrow:
## Summary

Replace Quickshell UI click handlers that shell out to `qs ... ipc call ...` with in-process or global-shortcut actions.

This avoids starting an extra Quickshell instance/client when clicking region screenshot/search/record controls from inside the running shell.

## Problem

Several QML controls invoke actions like this from inside Quickshell:

```qml
Quickshell.execDetached(["qs", "-p", Quickshell.shellPath(""), "ipc", "call", "region", "screenshot"])
```

Because `-p` selects a QML path/config folder, this can create another Quickshell instance/client instead of only dispatching to the already-running shell. On my setup, clicking the top-bar screen snip button caused a duplicate bar to appear and left processes like:

```text
/usr/bin/qs -p /home/.../.config/quickshell/ii/ ipc call region screenshot
```

The region selector already exposes the relevant actions as Quickshell global shortcuts, so these buttons do not need to spawn `qs`.

## Changes

- Use existing global shortcuts for region actions:
  - `global quickshell:regionScreenshot`
  - `global quickshell:regionSearch`
  - `global quickshell:regionRecordWithSound`
- Replace local UI toggle IPC calls with direct `GlobalStates` updates where applicable.
- Remove in-shell `qs -p ... ipc call ...` usage from affected UI click handlers.

## Why This Approach

Hyprland keybinds already route through Quickshell global shortcuts for these actions. Using the same path from QML keeps behavior aligned with the existing architecture while avoiding a second Quickshell process.

For panel toggles that are already in the same QML runtime, updating `GlobalStates` directly preserves the existing target behavior without launching another `qs` process.

## Validation

- Confirmed the affected buttons no longer spawn `qs ... ipc call region ...` processes.
- Confirmed only one `ii` Quickshell instance remains after using the screen snip control.
- Confirmed the config loads without new QML errors in my local setup.

### DIFF
--- a/dots/.config/quickshell/ii/modules/common/models/quickToggles/ScreenSnipToggle.qml
+++ b/dots/.config/quickshell/ii/modules/common/models/quickToggles/ScreenSnipToggle.qml
@@ -1,5 +1,6 @@
 import QtQuick
 import Quickshell
+import Quickshell.Hyprland
 import qs
 import qs.services
 import qs.modules.common
@@ -21,7 +22,7 @@ QuickToggleModel {
         interval: 300
         repeat: false
         onTriggered: {
-            Quickshell.execDetached(["qs", "-p", Quickshell.shellPath(""), "ipc", "call", "region", "screenshot"]);
+            Hyprland.dispatch("global quickshell:regionScreenshot")
         }
     }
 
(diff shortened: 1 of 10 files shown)


end-4 (commented):
Using GlobalStates variable is a good idea, but I want to keep the other ipc calls as is to make lives of people adapting the shell to other non-Hyprland environments easier

samscarrow:
Closing after further local diagnosis. The original premise of this PR was that in-shell qs ipc calls were spawning duplicate shell UI. Testing shows qs -p/-c ... ipc behaves as a normal short-lived IPC client, and the reproduced screenshot path reaches the existing region selector/global shortcut path instead. The stronger evidence points at local Quickshell runtime/package instability plus regionSelector UI warnings, so this PR is not the right fix.

PR #3284, opened 2026-05-07
Fix kitty config

fb24m:
## Describe your changes

Just moved the comment below as kitty can't parse this correct (see screenshot)

image


yosuga7:
same problem here

end-4 (approved):

PR #3288, opened 2026-05-08
Use packages from end-4/ii-package-builds for fedora

fb24m:
## Describe your changes

#3282

Changes:
- script doesn't build packages on user's machine anymore
- instead of that, script creates a local repo and downloads rpms there

Pros:
1. installation much faster (tens of minutes on low-end machines)
1. it looks cool in package list
1. there will be fewer "work on my machine" problems

I also replaced note in "Tested" section in readme since i don't know if it will work on Fedora 43, but it should. And Fedora 44 is stable now so it's not a problem

image

It's ready, BUT end-4/ii-package-builds#1 should be merged first


clsty:
Thanks for contribution!

fb24m:
we probably should wait when action complete before you merged this, though i don't think it's a big problem

sdata/dist-fedora/install-deps.sh
```diff
@@ -23,61 +23,20 @@ function r() {
...
-
-    if [[ -f "$rpmbuildroot/RPMS/x86_64/$spec_stamp.rpm" ]]; then
-	echo -e "${STY_BLUE}Next command:${STY_RST} sudo dnf install $rpmbuildroot/RPMS/x86_64/$spec_stamp.rpm -y"
-    	r x sudo dnf install "$rpmbuildroot/RPMS/x86_64/$spec_stamp.rpm" -y
-    fi
-  done
-  x cd ${REPO_ROOT}
+# Init local RPM repo and download rpms from releases there.
+function init_local_repo() {
+    url="https://api.github.com/repos/end-4/ii-package-builds/releases/tags/packages-fedora"
```
fb24m, line 28:
```suggestion
    url="https://api.github.com/repos/end-4/ii-package-builds/releases/tags/packages"
```

@clsty could you do this? it seems like one of my commits to builder somewhy wasn't applied and i don't want to create a new pr

clsty: I've moved `packages` to `packages-fedora` on the builder side.

fb24m: Well, that's even better

PR #3292, opened 2026-05-09
feat: notification on specific monitor

electricsteve:
## Describe your changes

Adds an option to show notifications on a specific monitor instead of the currently focused one. This may come in handy in multiple situations: such as when streaming/screen-sharing, and you still want to see your notifications, but don't want others to. Or just when you like it better this way.

IMO it is ready, attached a demo, but if you have feedback please provide it.

https://github.com/user-attachments/assets/06c6232d-948e-41a7-b452-77626b23fdf4

### DIFF
--- a/dots/.config/quickshell/ii/modules/common/Config.qml
+++ b/dots/.config/quickshell/ii/modules/common/Config.qml
@@ -399,6 +399,10 @@ Singleton {
 
             property JsonObject notifications: JsonObject {
                 property int timeout: 7000
+                property JsonObject monitor: JsonObject {
+                    property bool enable: false
+                    property string name: "" // Name of the monitor to show notifications on, like "eDP-1". Find out with 'hyprctl monitors' command
+                }
             }
 
             property JsonObject osd: JsonObject {
--- a/dots/.config/quickshell/ii/modules/settings/InterfaceConfig.qml
+++ b/dots/.config/quickshell/ii/modules/settings/InterfaceConfig.qml
@@ -268,10 +268,10 @@ ContentPage {
 
         ConfigSwitch {
             buttonIcon: "monitor"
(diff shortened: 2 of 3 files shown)


dots/.config/quickshell/ii/modules/common/Config.qml
```diff
@@ -399,6 +399,10 @@ Singleton {
 
             property JsonObject notifications: JsonObject {
                 property int timeout: 7000
+                property JsonObject monitor: JsonObject {
+                    property bool enable: false
```
end-4, line 403: `notifications.monitor.enable` doesn't sound right... let's change `monitor` to `forceMonitor`

dots/.config/quickshell/ii/modules/settings/InterfaceConfig.qml
```diff
@@ -265,6 +265,31 @@ ContentPage {
...
+            MaterialTextArea {
+                Layout.fillWidth: true
+                placeholderText: Translation.tr("Monitor name to show notifications on (e.g., eDP-1)")
+                text: Config.options.notifications.monitor.name
+                wrapMode: TextEdit.Wrap
+                onTextChanged: {
+                    Config.options.notifications.monitor.name = text;
+                }
+            }
+        }
```
end-4, line 292: Useless ConfigRow

end-4: nvm i see, it's for the disabling

dots/.config/quickshell/ii/modules/common/Config.qml
electricsteve: Damn I was thinking for some time how to name it, I'm stupid bruh

PR #3303, opened 2026-05-12
fix: show escaped text for selected entry in clipboard

ilnb:
## Describe your changes

This would show escaped text for selected entry in the clipboard. Previously the text wasn't escaped which lead to bugs like "#include <stdio.h>" being shown as "#include " when selected. 

It should be ready, as far as I have tested.


end-4 (approved):
thank you

PR #3312, opened 2026-05-12
Fix monitor scale type

PetLucy:
Changes the default monitor scale value from `"1"` to `1`.

This fixes the typo in the default monitor configuration.

See the Hyprland wiki [here](https://wiki.hypr.land/Configuring/Basics/Monitors/) for confirmation of the formatting.


end-4:
somehow "1.5" works... hyprland is really smart
but this fix is correct, thanks

PR #3306, opened 2026-05-12
fix(keybinds): correct fullscreen and maximize bind syntax

Rasmion:
## Describe your changes
The keybinds for fullscreen (e.g. `SUPER + D/F/ALT + F`) used a wrong lua syntax.
The `description` was passed to `hs.dsp.window.fullscreen` instead of `hsl.bind` and the fullscreen `mode` was passed as an array element instead of a named table field. 

## Is it ready? Questions/feedback needed?
Yes it ready, no feedback needed (i hope)

PR #3313, opened 2026-05-13
Fix fullscreen keybindings

An-dz:
## Describe your changes

The new Lua configs have a bug with the `Super` + `D` shortcut, it just goes fullscreen instead of maximising. The problem was it just lacked the parameter name and the default behaviour is fullscreen.

Perfect, just tested it.

Docs for reference:
https://wiki.hypr.land/Configuring/Basics/Dispatchers/#window-1


ar-Raqmi:
A PR addressing this problem is already open https://github.com/end-4/dots-hyprland/pull/3306

PR #3316, opened 2026-05-13
Fix HOME env in lua

imitoy:
## Describe your changes
Replace hardcoded $HOME with home_dir concatenation in env.lua for environment variables to ensure proper expansion.

It is ready and doesn't need any feedback.

## Reason
In env.lua, Lua strings are set as environment variables without shell expansion, so $HOME remains a literal string. Using os.getenv("HOME") guarantees the correct home path.

## Note
While exec.lua and keybinds.lua are unaffected because their strings are executed directly (e.g., via os.execute) where $HOME is expanded by the shell, I didn't modify them.

### DIFF
--- a/dots/.config/hypr/hyprland/env.lua
+++ b/dots/.config/hypr/hyprland/env.lua
@@ -1,8 +1,10 @@
+local home_dir = os.getenv("HOME")
+
 -- Wayland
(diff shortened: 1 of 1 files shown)


end-4 (approved):

end-4:
thanks
i also realized hyprland can adjust this on the fly now, nice

PR #3334, opened 2026-05-14
qs: wallpaper selector: fix missing thumbnails on hidpi monitors

daganbronstein:
## Describe your changes

On a multi-monitor setup with one 1.0-scale and one 2.0-scale display, the wallpaper selector shows blank cells on the hidpi monitor.

The cause is a DPR mismatch: `WallpaperSelectorContent.qml` calls `Images.thumbnailSizeNameForDimensions(grid.cellWidth - margins, ...)` in logical pixels, but `ThumbnailImage` (a `StyledImage`) computes its own `thumbnailSizeName` from `sourceSize`, which `StyledImage.qml` already multiplies by `devicePixelRatio`. So on a 2x screen with ~280 logical-px cells the generator writes to `large/` (256) while the display reads from `x-large/` (512), and the cache miss leaves the cells blank.

This patch multiplies `cellWidth`/`cellHeight` by the same DPR `StyledImage` uses, so generator and display agree. On 1x screens the multiplication is a no-op; on 2x screens the missing bucket gets generated on first open.

Ready.

### DIFF
--- a/dots/.config/quickshell/ii/modules/ii/wallpaperSelector/WallpaperSelectorContent.qml
+++ b/dots/.config/quickshell/ii/modules/ii/wallpaperSelector/WallpaperSelectorContent.qml
@@ -18,7 +18,8 @@ MouseArea {
 
     function updateThumbnails() {
         const totalImageMargin = (Appearance.sizes.wallpaperSelectorItemMargins + Appearance.sizes.wallpaperSelectorItemPadding) * 2;
-        const thumbnailSizeName = Images.thumbnailSizeNameForDimensions(grid.cellWidth - totalImageMargin, grid.cellHeight - totalImageMargin);
+        const dpr = (QsWindow.window as QsWindow)?.devicePixelRatio ?? 1;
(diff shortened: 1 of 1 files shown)


daganbronstein:
This is the fix I applied locally that resolved the issue for me. 

I had found this open issue: https://github.com/end-4/dots-hyprland/issues/1902

but I suspect that this is mostly a collection of related, but independent issue, a some of the proposed "workarounds" there are naturally irrelevant to the nature of this problem. (e.g. length of filename, capitalization...)

daganbronstein:
Just saw this is unmerged. Bump, any concerns? This has been working fine for me. A more fundamental approach / rewrite would be needed in order to prevent this from happening again in new places that try to generate thumbnails, but this does fix the selector.

anfsity:
Opened #3653 for the same issue. It passes the size name down to the delegates rather than only fixing the generator, so both sides always agree. I can close it if you would rather land this one.

anfsity:
Ran into the same thing and opened #3653. Mine passes the thumbnail size name down to the delegates instead of only changing the generator side.

PR #3345, opened 2026-05-16
update hypridle's dispatchers

ypurpl:
i dont think the dpms works properly yet, no real docs either

## Describe your changes

- fixed `hyprctl dispatch` to use lua

- above


end-4:
uhh i merged #3393 which is exactly the same as this
thanks anyway

PR #3358, opened 2026-05-17
Fix: XDG_DATA_DIRS now expands correctly (fixes 3354)

GregorVal:
Previously the environmental variable would become literally  *:$XDG_DATA_DIRS
Now the variable is expanded correctly


Smartyguy001:
My nvim just randomly broke, only to find this issue as the root cause!

pwease someone merge this goated PR

end-4:
thank you

PR #3370, opened 2026-05-18
fix environments by adding a parser

Jott55:
## Describe your changes

adds a parser so it substitutes any  existing environment variable 

Yes. This is a solution that suits my needs, and works fine at least tested with all of my own  variables


end-4:
I'd rather stick with the home_dir obtained from `os.getenv`, since it's simpler and less error prone than a custom parser

PR #3365, opened 2026-05-18
change lock command in hypridle.conf to work with latest hyprland

Pommersche92:
`loginctl lock-session` didnt work as of latest hyprland version and latest dots.

- Cause: when `loginctl lock-session` triggers hypridle, it executes a command containing `global quickshell:lock`. The new Lua parser hits the colon, gets confused, thinks it’s an invalid Lua function, and silently crashes.
- Fix: bypass the broken Hyprland global dispatcher and target Quickshell directly

PR #3372, opened 2026-05-19
Fix old syntax hyprctl dispatchers in hypridle.conf

VietNguyenVN:
## Describe your changes

Fix some dispatchers in `hypridle.conf` that were not modified for Hyprland 0.55 (I think end4 simply just missed those ones).
This should fix `quickshell:lockFocus` and dpms on lockscreen after sleep.
https://wiki.hypr.land/Configuring/Basics/Dispatchers/#cursor

Yes, I should've noticed this long ago but here we are

### DIFF
--- a/dots/.config/hypr/hypridle.conf
+++ b/dots/.config/hypr/hypridle.conf
@@ -16,7 +16,7 @@ listener {
 
 listener {
     timeout = 600 # 10mins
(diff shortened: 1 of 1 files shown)


VietNguyenVN:
Related to #3365

end-4:
uhhhhhhhhhhhhhhh whoops i merged #3393 which is the same as this one

PR #3375, opened 2026-05-20
qs: preload sidebar dialogs and bar popups (closes #3350)

### LINKED ISSUE #3350: Extend keepRightSidebarLoaded to inner toggle dialogs to eliminate first-open warmup

The `sidebar.keepRightSidebarLoaded` flag currently keeps only the root Sidebar shell instantiated, while the inner `ToggleDialog` components (NightLight, Volume, Bluetooth, Wifi) still use `Loader { active: shown }` — so the **first** open of each dialog instantiates and compiles its QML on the hot UI thread, producing a visible warmup hitch on slower systems (notably NVIDIA + Wayland).

I propose extending the flag's semantics: once the sidebar is loaded, an idle timer (~2.5s) flips a `dialogsPreloaded` property that pre-instantiates each ToggleDialog. Preloaded items are kept invisible via `opacity: 0` + `enabled: false` on the Loader, so they don't render and don't swallow clicks — but their QML is already compiled, so opening them is instant.

### How will it help?

First-open latency for each sidebar dialog drops from "noticeable hitch" to instant on systems where QML compilation is expensive.

The case is especially noticeable on **NVIDIA laptops driving an external monitor** — in my setup (RTX 4060 Laptop, driver 595, 1440p@165Hz external + 1080p@144Hz internal, Hyprland on Wayland) the first open of NightLight/Bluetooth/Wifi dialogs produced freezes lasting noticeably more than a frame. I have not rigorously bisected the cause (could be QML compile, EGL context setup, or display-list reconstruction across two outputs) but the symptom disappears with the proposed preload.

Cost is small and bounded: an extra ~20–30 MB RAM with the flag enabled, on top of the ~15 MB already documented for the existing flag — the user has already opted in to that tradeoff. Behavior is identical when the flag is off.

### Extra info

The patch is contained to a single file: `modules/ii/sidebarRight/SidebarRightContent.qml`. `WindowDialog.qml` does **not** need to change — the opacity-hide approach avoids touching it.

**1. Add preload state + idle timer to the root `Item`:**

```qml
property bool dialogsPreloaded: false
Timer {
    id: dialogsPreloadTimer
    interval: 2500
    running: Config?.options?.sidebar?.keepRightSidebarLoaded ?? false
    repeat: false
    onTriggered: root.dialogsPreloaded = true
}
```

**2. Patch the `ToggleDialog` component (4 lines added, 1 line changed):**

```diff
 component ToggleDialog: Loader {
     id: toggleDialogLoader
     required property string shownPropertyString
     property alias dialog: toggleDialogLoader.sourceComponent
     readonly property bool shown: root[shownPropertyString]
     anchors.fill: parent

-    onShownChanged: if (shown) toggleDialogLoader.active = true;
-    active: shown
+    onShownChanged: {
+        if (shown) {
+            toggleDialogLoader.active = true;
+            if (toggleDialogLoader.item) {
+                toggleDialogLoader.item.show = true;
+                toggleDialogLoader.item.forceActiveFocus();
+            }
+        }
+    }
+    active: shown || root.dialogsPreloaded
+    asynchronous: !shown
+    opacity: shown ? 1 : 0
+    enabled: shown
     onActiveChanged: {
-        if (active) {
+        if (active && shown && item) {
             item.show = true;
             item.forceActiveFocus();
         }
     }
     Connections {
         target: toggleDialogLoader.item
         function onDismiss() {
             toggleDialogLoader.item.show = false
             root[toggleDialogLoader.shownPropertyString] = false;
         }
         function onVisibleChanged() {
-            if (!toggleDialogLoader.item.visible && !root[toggleDialogLoader.shownPropertyString]) toggleDialogLoader.active = false;
+            if (!toggleDialogLoader.item.visible
+                && !root[toggleDialogLoader.shownPropertyString]
+                && !root.dialogsPreloaded) {
+                toggleDialogLoader.active = false;
+            }
         }
     }
 }
```

I've been running this patch locally for a couple of hours; happy to send a PR if the approach is acceptable. (Open to gating this behind a separate `preloadDialogs` sub-flag instead of overloading the existing one.)

zzallirog:
Preloads sidebar inner dialogs (NightLight/Bluetooth/Wifi/Volume in+out) and bar popups (Resources/Battery/Clock/Weather/SysTray) so the first interaction with each is instant. Most visible improvement on NVIDIA + Wayland.

**Sidebar dialogs** — reuses existing `sidebar.keepRightSidebarLoaded`. A 2.5s idle Timer flips `dialogsPreloaded`; ToggleDialog becomes `active: shown || dialogsPreloaded` with `asynchronous: !shown`.

**Bar popups** — new `bar.preloadPopups` flag (default true). Same Timer pattern in StyledPopup; PanelWindow `visible` is bound to hover so preloaded instances compile without mapping a Wayland surface.

**WindowDialog hotfix (1st commit)** — needed for preload to be sound. The existing `visible: dialogBackground.implicitHeight > 0` paired with imperative `implicitHeight = show ? backgroundHeight : 0` from `onShowChanged` worked for direct-toggle but ghost-rendered under hidden-state instantiation (default `show: false` never emits `showChanged`, so implicitHeight stayed at content-derived value). Replaced with declarative `implicitHeight: root.show ? backgroundHeight : 0`. `backgroundHeight` default rebased to content directly to avoid self-referential binding.

Closes #3350 (the opacity-only patch I proposed there was unsound for exactly this reason).

Tested: NVIDIA RTX 4060, Hyprland on Wayland, 165Hz ext + 144Hz int. First open of each sidebar dialog and first hover of each bar popup are instant; no artifacts during the 2.5s preload window.

Opt-out: `sidebar.keepRightSidebarLoaded: false` / `bar.preloadPopups: false`.

### DIFF
--- a/dots/.config/quickshell/ii/modules/common/Config.qml
+++ b/dots/.config/quickshell/ii/modules/common/Config.qml
@@ -241,6 +241,7 @@ Singleton {
                 property bool showBackground: true
                 property bool verbose: true
                 property bool vertical: false
+                property bool preloadPopups: true // Eagerly compile bar popups after a short idle so first hover is instant
                 property JsonObject resources: JsonObject {
                     property bool alwaysShowSwap: true
                     property bool alwaysShowCpu: true
--- a/dots/.config/quickshell/ii/modules/common/widgets/WindowDialog.qml
+++ b/dots/.config/quickshell/ii/modules/common/widgets/WindowDialog.qml
@@ -9,8 +9,8 @@ Rectangle {
     id: root
 
     property bool show: false
-    default property alias contentData: contentColumn.data
-    property real backgroundHeight: dialogBackground.implicitHeight
+    default property alias data: contentColumn.data
+    property real backgroundHeight: contentColumn.implicitHeight + dialogBackground.radius * 2
     property real backgroundWidth: 350
     property real backgroundAnimationMovementDistance: 60
     
@@ -30,7 +30,6 @@ Rectangle {
 
     onShowChanged: {
         dialogBackgroundHeightAnimation.easing.bezierCurve = (show ? Appearance.animationCurves.emphasizedDecel : Appearance.animationCurves.emphasizedAccel)
-        dialogBackground.implicitHeight = show ? backgroundHeight : 0
     }
 
     radius: Appearance.rounding.screenRounding - Appearance.sizes.hyprlandGapsOut + 1
@@ -51,7 +50,7 @@ Rectangle {
         property real targetY: root.height / 2 - root.backgroundHeight / 2
         y: root.show ? targetY : (targetY - root.backgroundAnimationMovementDistance)
         implicitWidth: root.backgroundWidth
-        implicitHeight: contentColumn.implicitHeight + dialogBackground.radius * 2
+        implicitHeight: root.show ? root.backgroundHeight : 0
         Behavior on implicitHeight {
             NumberAnimation {
                 id: dialogBackgroundHeightAnimation
--- a/dots/.config/quickshell/ii/modules/ii/bar/StyledPopup.qml
+++ b/dots/.config/quickshell/ii/modules/ii/bar/StyledPopup.qml
@@ -12,12 +12,21 @@ LazyLoader {
     property Item hoverTarget
     default property Item contentItem
     property real popupBackgroundMargin: 0
+    property bool preloaded: false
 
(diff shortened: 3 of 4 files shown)


zzallirog:
Superseded by #3376 (WindowDialog hotfix) + #3377 (preload feature) — split into two PRs as the hotfix is a real latent bug independent of preload, and reviewing them separately is cleaner. Closing this one.

PR #3381, opened 2026-05-20
Restore nwg-displays entry

mQzLjP:
## Describe your changes

The 0.4.3 update of nwg-displays now creates config files in both lua and hyprlang, so now we can bring back the entries in the main config.


end-4:
Let's also add the file existence check

PR #3395, opened 2026-05-24
fix: music/media box doesn't disappear when turn on auto hide.

MVN123123123123123:
## Describe your changes
i noticed that when u turn on auto hide bar, it hides when you move ur mouse out of the bar. but the box still there. it doesn't hide.
this will make it so that only when ur mouse moves out of both the box and the bar, it will hide the box first and then the bar.

yeah.

https://github.com/user-attachments/assets/73099782-74d0-4d4a-bab4-b679fb109ccf


MVN123123123123123:
i will add in a option in the setting, so you can toggle if you want it or not.

MVN123123123123123:
https://github.com/user-attachments/assets/d3ba02c6-de6e-47ff-baa1-6a85913eb15c

here's the result. don't mind the background and the music i'm hearing.

MVN123123123123123 (commented):
hmm, what should i write here...
i haven't seen this button before.
ask for review ?
alright. here's the result of my modification.
a button to toggle the new setting, idk how to tell that here so take a look at the video.
the text and button don't look like they perfectly blended in with the other buttons. i should have do the yes/no like the automatically hide button up there.

Image

https://github.com/user-attachments/assets/554d8c11-1dd7-4816-aff7-b99156f2f9a3

An-dz:
I don't think an option is necessary, your fix really feels like a fix. But let's see what the maintainers have to say.

MVN123123123123123:
> I don't think an option is necessary, your fix really feels like a fix. But let's see what the maintainers have to say.

Some people really have some odd preference. So I just add it in as an option.

An-dz:
> > I don't think an option is necessary, your fix really feels like a fix. But let's see what the maintainers have to say.
> 
> Some people really have some odd preference. So I just add it in as an option.

Obligatory xkcd comic:
https://xkcd.com/1172/

MVN123123123123123:
lol

PR #3396, opened 2026-05-24
Implement nwg-displays bridge in general.lua

GraveEaterMadison:
Add nwg-displays bridge to manage monitor configurations


end-4:
image

end-4:
fixed in #3381

PR #3397, opened 2026-05-25
fix: remove duplicate SUPER+ALT+number keybinds

jakedev796:
**Description:**

In commit c070625, character-based binds for `SUPER+ALT+<number>` were added back alongside the existing keycode-based binds for cheatsheet display purposes.

On standard keyboard layouts, pressing a number row key fires both the character bind (`"SUPER + ALT + 1"`) and the keycode bind (`"SUPER + ALT + code:10"`) simultaneously. This causes `hl.dsp.window.move` to dispatch twice per keypress, moving two windows to the target workspace instead of one.

The numpad binds are unaffected because they only have keycode binds, which is why the behavior differs between number row and numpad.

**Triage note:** This was initially suspected to be a Hyprland 0.55 regression. After ruling out groups, window rules, ydotool, and the Lua dispatcher itself (verified via `hyprctl dispatch 'hl.dsp.window.move({workspace=N})'` which correctly moves only the focused window), the duplicate bind was identified as the root cause.

**Fix:** Remove the character-based loop. The keycode binds (`code:10-19`) already cover all keyboard layouts including non-US ones. The `--#/#` cheatsheet comment line is preserved.

### DIFF
--- a/dots/.config/hypr/hyprland/keybinds.lua
+++ b/dots/.config/hypr/hyprland/keybinds.lua
@@ -196,11 +196,11 @@ hl.bind("SUPER + ALT + F", hl.dsp.window.fullscreen_state({ internal = 0, client
 hl.bind("SUPER + P", hl.dsp.window.pin(), { description = "Window: Pin" })
 
 --#/# bind = SUPER+ALT, Hash,, -- Send to workspace -- (1, 2, 3,...)
-for i = 1, 10 do
-    hl.bind("SUPER + ALT + " .. (i % 10), function()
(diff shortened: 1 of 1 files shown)


end-4:
I decided to comment out the keycode binds instead d619ddcd8264b6daecf98b50e2ddd19f61ad1ecd
The good thing: Normal number keybinds show proper keys in the cheatsheet (not `code:number`)
The bad thing: This hurts keyboards whose number row keys aren't numbers, **but** even if we leave the keycode binds users with such a setup would still have to edit the other keybinds that maps weirdly to their layout anyway

PR #3398, opened 2026-05-25
[bug+feature] Generate wallpaper thumbs efficiently with magisk (No more killing the OS)

An-dz:
## Describe your changes

This does 2 things:

1) First it stops the fallback thumbnail generation (the one that uses magisk through a bash script) from killing the OS. Because it just creates as many processes as there are images on the folder, so you end up with hundreds of processes loading hundreds of images. This can cause the kernel from dumping quickshell out of the memory or even worse. Now the script only runs on half the CPUs so the PC barely notices anything.
2) It adds the same behaviour from the normal thumb generation that shows the thumbs on the UI as they are generated.

Yes. I've tested on my PC and it generates the thumbs updating the UI as the thumbs are generated and the OS doesn't kill itself because there are 1 billion magisk processes running at 100%.

But my bash skills are not pro, so there's always a chance of bugs. The `signal_dir` was a suggestion from an AI as I was stuck on how to get the progress with async background processes, but it does makes sense.

### DIFF
--- a/dots/.config/quickshell/ii/scripts/thumbnails/generate-thumbnails-magick.sh
+++ b/dots/.config/quickshell/ii/scripts/thumbnails/generate-thumbnails-magick.sh
@@ -7,6 +7,20 @@
 
 set -e
 
+matches_pattern() {
+    local filename="$1"
+    local patterns_str="$2"
+    local IFS='|'
+    read -ra pattern_array <<< "$patterns_str"
+
+    for p in "${pattern_array[@]}"; do
+        if [[ "$filename" == $p ]]; then
+            return 0
+        fi
+    done
+    return 1
+}
+
 # Thumbnail sizes mapping
 get_thumbnail_size() {
     case "$1" in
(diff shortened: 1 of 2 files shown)


Leo-Aqua:
Works for me too. No problems on my end

An-dz:
Thanks for the test.

I tested on my machine on a folder with 342 very high quality pictures having 4.9 GB total size and it went very smoothly and fast.

Also, this fallback script always runs on Arch because the first script requires the `gnome-desktop-4` package installed, and none of the ii dependencies depend on it. Apart from this PR, it could be made an optional dependency of `illogical-impulse-basic` or `illogical-impulse-widgets`.

anfsity:
Also rewrote this locally. Differences in mine:
- `realpath -s` instead of `realpath` — Quickshell hashes the unresolved path
- byte-wise urlencode (multi-byte names; `!` and `'` are left unescaped by encodeURIComponent)
- `XDG_CACHE_HOME` instead of hardcoded `~/.cache`
- `[0]`, `>` and `Thumb::URI`/`Thumb::MTime` per the spec

anfsity:
Also rewrote this locally. Mine uses realpath -s, byte-wise urlencode, XDG_CACHE_HOME, and writes [0], > and the Thumb:: fields.

Issue #1032, opened 2025-01-01
[Issue] Some songs on spotify don't change their album covers  


oztopenessafa:
- I have read the Usage and Troubleshooting pages of the wiki: https://end-4.github.io/dots-hyprland-wiki/en/i-i/04troubleshooting/
- I have made sure that both my config and system packages are up to date
- Linux distro: 
CachyOS
---
## The issue
 There's this weird issue with some songs that don't change their covers automatically but it changes when spotify's paused. And playerctl metadata shows the correct image url at all times  

<details>
<summary>Logs</summary>

```plain
Paste logs here
```

</details>

PR #3405, opened 2026-05-26
fix(ii): stabilize album cover across player and panel quirks

Beetix:
## Summary
The media controls panel's album cover behavior has several failure modes that compound, especially with Firefox and Chromium. This PR addresses them in `MprisController.qml` and `PlayerControl.qml`:

1. **Cover blanks on panel close/reopen.** PlayerControl is destroyed and recreated each time the panel opens; the previous design reset `downloaded` to false, so the cover was hidden until the next download completed even when the cached file was still on disk.
2. **Cover degrades to a thumbnail mid-track.** firefox-mpris emits multiple artUrls per song at different resolutions (e.g., 544×544 then 60×60). The previous design swapped to whichever arrived last.
3. **Cover stays low-res after a track auto-advances while the panel is closed.** The previous design ran the downloader inside PlayerControl, which only exists while the panel is open. Emissions during a closed-panel period are missed; on reopen, only the current (typically low-res) URL is downloaded.
4. **Multi-player cross-contamination.** Two unrelated players that both happen to be on their first track share `player.uniqueId === 1` (Quickshell's `MprisPlayer.uniqueId` is per-track, not per-player), and their cache entries clobber each other.
5. **trackKey changes mid-track because album fills in late.** firefox-mpris emits metadata progressively: first artUrl + title + artist with `album=""`, then later updates album. With album in the trackKey, the high-res emitted with partial metadata and the low-res emitted with completed metadata look like different tracks, and the low-res ends up indexed under the (full-metadata) key the consumer reads with.

## Implementation
**MprisController** (`services/MprisController.qml`):
- New per-player `PlayerWorker` (QtObject) created via `Instantiator` over `Mpris.players`. Each worker:
  - listens to its player's `trackArtUrlChanged`,
  - downloads the new file via a `Process` that runs `[ -f file ] || curl ... ; stat -c %s file`,
  - on exit, calls `rememberBestArt(player, trackKey, file, sizeBytes)`.
- Two persistent per-player caches:
  - `stableArtUrlByPlayer`: last non-empty `trackArtUrl` per player, used by PlayerControl as a fallback during empty-URL windows.
  - `bestArtByPlayer`: best (largest) downloaded file per player + track, with `rememberBestArt()` / `getBestArt(player, trackKey)` helpers. Strict `>` size guard.
- Both caches are keyed on `player.dbusName`, not `uniqueId`.
- trackKey is `title|artist` (album deliberately omitted; see bug #5 above).

**PlayerControl** (`modules/ii/mediaControls/PlayerControl.qml`):
- No longer owns a downloader. It's a pure viewer of `MprisController.bestArtByPlayer`:
  ```qml
  readonly property string displayedArtFilePath: {
      const entry = MprisController.getBestArt(player, _trackKey);
      return entry ? Qt.resolvedUrl(entry.artFilePath) : "";
  }
  ```
- `displayedArtFilePath` automatically updates when the worker's `rememberBestArt` reassigns `bestArtByPlayer`.

## Test plan
- [x] Firefox + YouTube Music: switch several tracks; cover stays sharp on each (no degradation to a thumbnail a second later)
- [x] Firefox + YouTube Music: open panel, close, reopen — cover restores at full quality
- [x] Firefox + YouTube Music: leave panel closed, let track auto-advance, then reopen — new track's cover is sharp (worker captured the high-res emission while panel was closed)
- [x] Firefox + Chromium both playing different tracks: each row shows its own cover at full resolution; no cross-contamination
- [x] Chromium + YouTube Music alone: no flicker on track changes (per-track tmp-file flap rejected by strict `>` size guard)
- [ ] Spotify desktop / other native MPRIS player: no regression
- [ ] No-art player (e.g., browser without `mpris:artUrl`): placeholder behavior unchanged

## Related
- Overlaps with #1032 ("Some songs on spotify don't change their album covers").
- Orthogonal to #3265 / #3378 (YouTube cover fallback for players that *never* set `mpris:artUrl`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/dots/.config/quickshell/ii/modules/common/functions/StringUtils.qml
+++ b/dots/.config/quickshell/ii/modules/common/functions/StringUtils.qml
@@ -295,4 +295,27 @@ Singleton {
             }
         );
     }
+
+    /**
+     * Returns a URL the cover-art downloader can fetch as artwork for a
+     * YouTube link, or "" if the URL is not from YouTube. Used as a fallback
+     * when the player exposes no mpris:artUrl (Firefox MPRIS bridge,
+     * plasma-browser-integration on YouTube Music).
+     *
+     * - Watch URL (?v=, youtu.be/, /embed/, /shorts/) → deterministic
+     *   i.ytimg.com/vi/<id>/hqdefault.jpg.
+     * - Playlist / channel / other YouTube URL → public oEmbed endpoint;
+     *   the downloader resolves its `thumbnail_url` field at fetch time.
+     *
+     * @param { string } url
+     * @returns { string }
+     */
+    function getYoutubeArtUrl(url) {
+        if (!url || !/(?:^|\/\/|\.)(?:youtube\.com|youtu\.be)/.test(url))
+            return "";
+        const m = url.match(/(?:[?&]v=|youtu\.be\/|\/(?:embed|shorts)\/)([A-Za-z0-9_-]{11})/);
+        if (m)
+            return `https://i.ytimg.com/vi/${m[1]}/hqdefault.jpg`;
+        return `https://www.youtube.com/oembed?url=${encodeURIComponent(url)}&format=json`;
+    }
 }
--- a/dots/.config/quickshell/ii/modules/ii/mediaControls/PlayerControl.qml
+++ b/dots/.config/quickshell/ii/modules/ii/mediaControls/PlayerControl.qml
@@ -15,39 +15,27 @@ import Quickshell.Services.Mpris
 Item { // Player instance
     id: root
     required property MprisPlayer player
-    // Some MPRIS players (Firefox via firefox-mpris, Spotify) emit empty
-    // trackArtUrl between/during tracks. MprisController persists the last
-    // non-empty value per player so we can fall back to it when the panel is
-    // (re)opened during an empty window.
-    property string artUrl: {
-        const cur = player?.trackArtUrl ?? "";
-        if (cur.length > 0) return cur;
-        const id = player?.uniqueId;
-        if (id === undefined) return "";
-        return MprisController.stableArtUrlByPlayer[id] ?? "";
-    }
-    property string artDownloadLocation: Directories.coverArt
-    property string artFileName: Qt.md5(artUrl)
-    property string artFilePath: artUrl.length > 0 ? `${artDownloadLocation}/${artFileName}` : ""
     property color artDominantColor: ColorUtils.mix((colorQuantizer?.colors[0] ?? Appearance.colors.colPrimary), Appearance.colors.colPrimaryContainer, 0.8) || Appearance.m3colors.m3secondaryContainer
     property list<real> visualizerPoints: []
     property real maxVisualizerValue: 1000 // Max value in the data points
     property int visualizerSmoothing: 2 // Number of points to average for smoothing
     property real radius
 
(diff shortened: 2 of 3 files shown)


Beetix:
Heads-up: this PR conflicts with #3378 on `dots/.config/quickshell/ii/modules/ii/mediaControls/PlayerControl.qml` — same `artUrl` property and `coverArtDownloader` Process. The two changes are complementary in intent (#3378 handles *no `mpris:artUrl` at all*, this PR handles *transient empty + multi-resolution + lifecycle*), so whichever lands second will need to rebase. If this PR lands second, the rebase work is moving #3378's `StringUtils.getYoutubeArtUrl()` lookup and the oEmbed-aware curl command into the `PlayerWorker` in `MprisController.qml` — ~30-line port, no architectural change.

zzallirog:
built on top of this — nice architecture. added a YT thumbnail fallback in _fetchArt: when trackArtUrl is empty, falls back to StringUtils.getYoutubeArtUrl(xesam:url). covers the no-art browser case from your unchecked test item.

Beetix:
Added the YouTube cover synthesis for players that never set `mpris:artUrl` (Firefox MPRIS bridge, plasma-browser-integration on YT/YT Music), which closes the previously-unchecked test item on this PR.

Full credit to **@zzallirog** — this is their change ([zzallirog/dots-hyprland@4aa64ce](https://github.com/zzallirog/dots-hyprland/commit/4aa64ceb745c02fb7994076bd7aac3fc6b35b039)), cherry-picked with authorship preserved. It reuses the centralized `PlayerWorker` path (same `trackKey` / `bestArtByPlayer` keying), and also hardens the downloader by passing the URL as a positional arg to `bash -c` instead of interpolating it (removes a shell-injection vector for web-supplied MPRIS URLs) and adds `curl -f`.

I added one follow-up commit: the original needed `import qs.modules.common.functions` in `MprisController.qml` — `StringUtils` was otherwise unresolved and `_fetchArt` would throw a `ReferenceError` exactly when the fallback fires.

Note for reviewers: the fallback reads `player.metadata["xesam:url"]`; it degrades gracefully to a no-op if a player does not expose it, so it is safe, but the `xesam:url` availability is worth a sanity check per browser.

Beetix:
Tested the YouTube fallback live (Quickshell 0.2.1, Firefox on Wayland) — it works end to end:

- **`metadata["xesam:url"]` is exposed by Quickshell's `MprisPlayer`** and reads correctly; the follow-up import (`qs.modules.common.functions`) means `_fetchArt` executes without a `ReferenceError`.
- The file the shell wrote (name = `Qt.md5` of the synthesized URL) is **byte-identical** to a direct fetch of `i.ytimg.com/vi/<id>/hqdefault.jpg` (480×360 JPEG), so `getYoutubeArtUrl` + the download command are correct.

One real-world nuance worth noting: on current Firefox the `firefox-mpris` bridge *does* extract mediaSession artwork to a local PNG, so `trackArtUrl` is only empty for a **brief window at the start of playback**. The fallback fires in that gap (fills the otherwise-empty cover), then Firefox's higher-res PNG arrives and the existing best-art size guard supersedes it. So on setups where Firefox provides artwork it's a startup-gap filler; on setups where it never sets `trackArtUrl` (the case @zzallirog hit) it's the whole cover. Either way: no regression, and it degrades to a no-op if a player exposes no `xesam:url`.

PR #3415, opened 2026-05-29
OpenRGB integration

Leo-Aqua:
This change adds support for OpenRGB! 

When the accent changes it automatically updates the colors on all OpenRGB devices with a nice transition too!

Preview:

https://github.com/user-attachments/assets/4aeda4bf-d980-4a44-9c6f-b593237aeffc

A settings entry has also been inplemented:

image

The user can toggle this effect (disabled by default because not everyone has RGB devices)

and set the duration for the fade animation


hexahigh:
I suggest adding device selection since some might not want specific devices to be changed, I for instance configure my keyboard manually.
I recently implemented OpenRGB support into my own fork with device selection, you can take a look if interested https://github.com/hexahigh/dots-hyprland
I didn't know transitions was a thing so might steal that feature into my fork :)

Leo-Aqua:
Good idea! I might look into that.

Leo-Aqua:
I've combined your device selector with my fade transition effect and also did some minor tweaks to it.
However i was wondering, maybe also having access to zones would be even better.
My mobo has multiple ARGB interfaces so having control over that would be pretty cool.

hexahigh:
Cool! Adding support for zones would be great, but might be a little hard without cluttering up the UI.

Keep in mind that I split the terminal theming options into two in my fork and you partially merged that. So the option to enable terminal theming for Kitty is visible in the options but does nothing as the rest of the code is missing.
image
I should probably try to get some of my changes merged into upstream at some point.

Leo-Aqua:
I removed it, fixed a bug where the devices where stuck on disabled and relocated the devices under the main toggle switch. I have no idea why they weren't there in the first place, maybe the fact that it's 3:00AM and I'm tired played a role here 😅

PR #3419, opened 2026-06-03
fix move window

daniel0alves:
## Describe your changes

When I was moving a window from one workspace to another it was moving the entire workspace, now this is solved

It is ready


end-4:
It doesn't fix it? You're changing the numpad binds which are fine because they don't have dupe binds, while the number row keys are unchanged

daniel0alves:
So, I'm sorry, I did it on my computer and it worked, tried on a friend's computer and it worked on some apps and others didn't, no idea the cause.

PR #3449, opened 2026-06-11
Feat/cheatsheet: add search for keybinds and periodic table elements

cheikhhachem:
## Describe your changes

Adds search to the Quickshell cheatsheet (Super+/):

!!(when I say "x" I mean the icon in the search/filter field and not the letter x letter)

-Search UI : Collapsible field under the tab bar with a “Type or click here to search…” hint, type-to-search, Esc/"x" to clear, and "x" inside the field
-Keybind filtering : Filters categories in place by description, key, and modifiers; shows “No matching keybinds” when nothing matches
-Key/mod matching : Shared cheatsheet_search.js haystack matches display names and symbols (Shift, Super, Enter, LMB/RMB, etc.), including case variants and mouse button text when symbols are shown
-Periodic table : Non-matching elements stay visible but dimmed (greyed out in place)
-Performance : Deferred keyboard focus (150ms after open) to avoid the slow spawn caused by WlrKeyboardFocus.OnDemand on creation.
-Translations : Two new en_US strings

### I made the field very compact to not mess with the visuals for those who don't use it, do its collapse in to the hint text when not used .

https://github.com/user-attachments/assets/dfb9f6f0-df76-4a0b-99c5-34cd420734ba

Ready for review, tested and working locally on my both setups, 

Open cheatsheet (Super+/) opens quickly with deferred focus
Type-to-search and click hint to search
Filter keybinds by description, key (e, enter), modifier (shift, super), and mouse buttons (lmb, rmb)
Periodic table grey-out filter ....
Esc clears search first, then closes, "x"and backspace clear search
Tab switching (Ctrl+PageUp/Down)

Feedback welcome on UX (e.g. hint wording, empty-state behavior) or whether deferred focus timing should be configurable.

### DIFF
--- a/dots/.config/quickshell/ii/modules/ii/cheatsheet/Cheatsheet.qml
+++ b/dots/.config/quickshell/ii/modules/ii/cheatsheet/Cheatsheet.qml
@@ -42,16 +42,45 @@ Scope { // Scope
             function hide() {
                 cheatsheetLoader.active = false;
             }
+            property bool keyboardReady: false
             exclusiveZone: 0
             implicitWidth: cheatsheetBackground.width + Appearance.sizes.elevationMargin * 2
             implicitHeight: cheatsheetBackground.height + Appearance.sizes.elevationMargin * 2
             WlrLayershell.namespace: "quickshell:cheatsheet"
-            // Setting this value makes it take its sweet time to open
-            // WlrLayershell.keyboardFocus: WlrKeyboardFocus.OnDemand
+            WlrLayershell.layer: WlrLayer.Overlay
+            //--> It dont take that sweet time to open Anymore!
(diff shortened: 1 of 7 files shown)


Satoxyan:
cool, i find this useful to spare some time from reading allat

cheikhhachem:
> cool, i find this useful to spare some time from reading allat


@Satoxyan ,Filtering is not about speed alone. For many users, dense blocks of text are overwhelming and hard to read. Adding search functionality is a huge accessibility win, as it allows users to isolate the specific information they need without the visual noise.
For users with conditions like ADHD, Dyslexia, or Visual Processing Disorders, this feature can be a good way to help in my opinion.

Issue #3429, opened 2026-06-06
Distrobox Containers fail to launch from the Quickshell Menu


Nuxview:
### Step 1. Before you submit

- [x] I have read the [Troubleshooting](https://ii.clsty.link/en/ii-qs/04troubleshooting/) and [Usage](https://ii.clsty.link/en/ii-qs/02usage/) pages.
- [x] I've successfully updated to the latest version following the [guidance](https://ii.clsty.link/en/ii-qs/01setup/#updating).
- [x] I've successfully updated the system packages to the latest.
- [ ] I've ticked the checkboxes without reading their contents

### Step 2. Quick diagnose info

<details><summary>Quick diagnose</summary>

```
/========================\
| Checking git repo info |
\========================/

[===diagnose===] git remote get-url origin
https://github.com/end-4/dots-hyprland
[---SUCCESS---]

[===diagnose===] git rev-parse HEAD
3cb611c04e6b166c0c2b07755302f239f0c84cac
[---SUCCESS---]

[===diagnose===] git status
On branch main
Your branch is up to date with 'origin/main'.

nothing to commit, working tree clean
[---SUCCESS---]

[===diagnose===] git submodule status --recursive
 e31ec4cb4ebf6a46b267f5c42eabf6874916fa16 dots/.config/quickshell/ii/modules/common/widgets/shapes (heads/main)
[---SUCCESS---]

/=================\
| Checking distro |
\=================/

[===diagnose===] _check_distro_id
distro ID: cachyos
distro ID_LIKE: arch
[---SUCCESS---]

[===diagnose===] cat os-release
cat: os-release: No such file or directory
[---EXIT 1---]

/====================\
| Checking variables |
\====================/

[===diagnose===] declare -p XDG_CACHE_HOME
declare -x XDG_CACHE_HOME="/home/pierce/.cache"
[---SUCCESS---]

[===diagnose===] declare -p XDG_CONFIG_HOME
declare -x XDG_CONFIG_HOME="/home/pierce/.config"
[---SUCCESS---]

[===diagnose===] declare -p XDG_DATA_HOME
declare -x XDG_DATA_HOME="/home/pierce/.local/share"
[---SUCCESS---]

[===diagnose===] declare -p XDG_STATE_HOME
declare -x XDG_STATE_HOME="/home/pierce/.local/state"
[---SUCCESS---]

[===diagnose===] declare -p ILLOGICAL_IMPULSE_VIRTUAL_ENV
[...]

PR #3451, opened 2026-06-12
Fix terminal command escaping in launcher

Nuxview:
## Describe your changes
I modified the `launchersearch.qml` to correctly handle terminal execs by properly escaping each argument before finally passing over to the default terminal.

This should ensure terminal apps with searchable desktop entries open instead of returning "Failed Child process errors"


elrodrigues:
I've tried this patch with terminal applications with menu entries on my system, and it works well. This should close #3429. Nice

@end-4

PR #3454, opened 2026-06-13
Updated BarConfig.qml with the new Transparent background toggle.

im-vihar:
## Describe your changes
First time contributing on GitHub, sorry in advance. 
#### (I am not aware as to why the line deletion and creation counts are jagged, only 5 or so lines have been added.

This addition adds a new toggle to the bar section in settings, "Transparent Background" which evidently removes the solid colour from the bar and lets the wallpaper shine through in that section. This works with even if the bar is Top, Bottom, Left, or Right.

Possibly. This change does not effect shadows, pill colours, etc, and may feel more complete & look better if done more comprehensively, but this looks pretty nice by itself. This change does not make changes to any other UI elements.

I must advise, it looks best with Hug or Rect corner-style and Line-seperated group style, and may not look appealing with specific wallpapers that have content that shines through the bar and makes bar content unreadable.
### Screenshots
image
image
image
image

### DIFF
--- a/dots/.config/quickshell/ii/modules/settings/BarConfig.qml
+++ b/dots/.config/quickshell/ii/modules/settings/BarConfig.qml
@@ -1,350 +1,359 @@
-import QtQuick
-import QtQuick.Layouts
-import qs.services
-import qs.modules.common
-import qs.modules.common.widgets
-
-ContentPage {
-    forceWidth: true
-
-    ContentSection {
-        icon: "notifications"
-        title: Translation.tr("Notifications")
-        ConfigSwitch {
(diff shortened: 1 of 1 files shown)


satori-5423:
The "jagged" line count is happening because there are trailing whitespaces (two spaces) at the end of every line in your modified file. Git flags this as a modification for almost the entire file.
image

im-vihar:
> The "jagged" line count is happening because there are trailing whitespaces (two spaces) at the end of every line in your modified file. Git flags this as a modification for almost the entire file. image

I see, I thank you for this.

PR #3456, opened 2026-06-14
gentoo microtex update

jwihardi:
changed microtex ebuild uri to match the arch pkgbuild's.


clsty:
Thanks for the update.

PR #3474, opened 2026-06-17
Fix source install-deps.sh

ghost:
Fix


clsty:
I'm closing this cuz it does not make sense.

PR #3482, opened 2026-06-20
Dev

Mukul-raii:


DRKCTRLDEV:
Was this an accidental submission? I see many changes and regressions such as `.conf` files rather than `.lua`...

Nuxview:
@DRKCTRLDEV I think it was cause it looks more like his custom configs

DRKCTRLDEV:
> @DRKCTRLDEV I think it was cause it looks more like his custom configs

sorry for the late reply, Ur probs right, looks like an accident...

PR #3486, opened 2026-06-23
Update three-finger pinch gesture action

Nitesh4546:
Changed the action for a three-finger pinch gesture from fullscreen to floating.

## Describe your changes
Previously, whenever I pinched using 3 fingers, it would make the focused window float. However, after the recent update, the action mapped to that gesture changed to fullscreen. To fix this, I changed the value from ```'fullscreen'``` to ```'float'``` on line 17 in the file ```.config/hypr/hyprland/general.lua```.

Yes

### DIFF
--- a/dots/.config/hypr/hyprland/general.lua
+++ b/dots/.config/hypr/hyprland/general.lua
@@ -14,7 +14,7 @@ hl.gesture({
 hl.gesture({
     fingers = 3,
     direction = "pinch",
-    action = "fullscreen"
+    action = "float" -- to make the window floating instead of full screen
 })
 hl.gesture({
     fingers = 4,


Xlini22:
Hi! I think this change would be better placed in your custom config instead of modifying hyprland/general.lua

This seems more like a personal preference, and if you change it directly in the default config, the modification may be overwritten or lost when you update the dots in the future. If you put it in your custom config instead, you won’t have to reapply the change after every update.

You can add this to:
~/.config/hypr/custom/general.lua

Since the default config already defines the 3-finger pinch gesture as 'fullscreen', it’s better to unset it first and then set your preferred action:

hl.gesture({ fingers = 3, direction = "pinch", action = "unset" })
hl.gesture({ fingers = 3, direction = "pinch", action = "float" })

This keeps the defaults untouched while still letting you use 3-finger pinch to toggle floating locally.
Hope this helps!

Nitesh4546:
Thanks, but I thought there was some mistake while luafication, as the gesture was supposed to make the window float. But if after luafication the gesture is set as intended by developer then it okay.

end-4:
it was intended and there was no mistake

PR #3494, opened 2026-06-28
fix(wallpapers): derive defaultFolder from saved wallpaperPath

Daverkex:
The defaultFolder was hardcoded to ~/Pictures/Wallpapers, causing randomFromCurrentFolder() to silently bail out (folderModel.count === 0) when that directory doesn't exist — e.g. when wallpapers are stored elsewhere such as ~/Wallpapers.

Now defaultFolder is derived from the parent directory of the persisted wallpaperPath in config.json, falling back to ~/Pictures/Wallpapers only when no wallpaper has been set yet (fresh install).

Ready and tested in my local


Daverkex:
September is here and this is currently up to day. Waiting for maintainers to keep an eye (1 line change :D)
Happy to help if doubs or need modifications!

PR #3536, opened 2026-07-22
feat(sidebar): add Manjaro distro icon support

Sanaruca:
## Describe your changes

Add `manjaro-symbolic.svg` (from font-logos, Unlicense) and map the `manjaro` distro ID to it in `SystemInfo.qml` so the **sidebar displays the Manjaro logo** instead of the generic Linux icon.

| Before | After |
|-----------|---------|
| image | image |

Ready, no questions/feedback needed.


end-4:
would be nice if the icon was also 20x20 like the others, but practically it doesn't matter. thanks.

PR #3539, opened 2026-07-23
perf(AppSearch): avoid rebuilding desktop entry list repeatedly

Spalishe:
## Describe your changes

AppSearch was rebuilding the desktop entry list too often, causing ui freezes. This makes the list persistent and rebuilds it only when needed.

Yea its ready


kurokawaerne:
Missing:
import QtQuick
import QtQml

image

Spalishe:
Sorry, forgot to copy from my own config to repo
Should be fine now

Issue #2666, opened 2025-12-09
Improve Clipboard


ahmed-mekky:
### What would you like to be added?

i think clipboard should have more functions, like pins/favorites, delete on restart, delete_all keymap,  private mode and its keymap, etc..

PR #3546, opened 2026-07-24
feat(clipboard): Clear clipboard buttons

Sanaruca:
## Describe your changes

I added **buttons to wipe the clipboard and search results** in the **Quickshell overview**.

Pasted image (2)

The Clipboard widget allows users to manage and preview their clipping history. It includes a search filter to refine the information displayed.

Currently, when you perform a specific search and want to clear only the filtered results (or the entire list), there's no direct way to do so from the interface. This forces you to delete items one by one or manually clear the search box, impacting the user experience and workflow efficiency.

To address this, **two new buttons were added to the Clipboard interface**:

- **Clear Results**: Removes from the list only the items that match the currently active search filter.
- **Clear All**: Clears all items from the clipboard, regardless of whether a filter is active or not.

By allowing you to clear only visible results with `Clear results`, you can quickly segment and discard filtered output or commands without losing the rest of your history or current work context from the clipboard. Additionally, `Clear all` offers a quick shortcut to completely clear the clipboard directly within the same widget.

https://github.com/user-attachments/assets/f9971c56-c1c6-4991-944a-cf9c798fabf5

Ready, the changes have already been tested locally and the state's behavior is as expected. I would like to receive feedback, especially on the visual layout and accessibility of the new buttons in the UI.

### Related

Contributes to [Improve Clipboard](https://github.com/end-4/dots-hyprland/issues/2666)

### DIFF
--- a/dots/.config/quickshell/ii/modules/ii/overview/SearchItem.qml
+++ b/dots/.config/quickshell/ii/modules/ii/overview/SearchItem.qml
@@ -40,6 +40,7 @@ RippleButton {
     property int buttonHorizontalPadding: 10
     property int buttonVerticalPadding: 6
     property bool keyboardDown: false
+    property bool clearBtnHasFocus: false
     readonly property bool selected: (root.hovered || root.focus)
 
     implicitHeight: rowLayout.implicitHeight + root.buttonVerticalPadding * 2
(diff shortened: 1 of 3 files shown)


ar-Raqmi:
Don't we already have `/wipeclipboard` command for this?

image

Sanaruca:
> Don't we already have `/wipeclipboard` command for this?

Yes, that command exists, but it's not very intuitive for most users. Adding these physical buttons in the clipboard witget makes the feature much more discoverable and accessible. Plus, the `Clear Results` button provides a quick way to clear only the filtered items, which offers better control than just wiping the whole history via command

PR #3552, opened 2026-07-27
[Fix] mouse workspace inverse keybind

nx-smul:
## Describe your changes

It was so annoying to inverse scroll the workspace

yes


Sanaruca:
I have a similar problem, but I suspected it was due to some global setting that wasn't related to Hyprland or this configuration

xBiei:
It's unrelated to the dots, you can keep the fix as a temp fix until [this PR](https://github.com/hyprwm/Hyprland/pull/15507) is released.

nx-smul:
hmm interesting ok

Sk-aista:
Also SUPER + SHIFT/ALT + mouse_scroll sends the windows to the workspace with inverse direction

PR #3584, opened 2026-08-12
feat(install): add dinit init system support

Unrectified:
## Describe your changes

Added dinit init system support
- enable dinit services 
- new install dinit-only packages for arch-like distro IF dinit is recognize (Artix linux in mind)

Should be ready, I tried it and it seems to work fine. But I was using working artix system, I couldn't try on fresh new one.

### DIFF
--- a/sdata/dist-arch/install-deps.sh
+++ b/sdata/dist-arch/install-deps.sh
@@ -110,7 +110,7 @@ install-dinit-service-packages(){
     return 0
   fi
 
-  for pkg in userspawn-dinit bluez-dinit networkmanager-dinit pipewire-dinit pipewire-pulse-dinit wireplumber-dinit; do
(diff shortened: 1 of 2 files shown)


clsty:
Thanks, this looks reasonable at the first glance. I'll delay the review though since you're still adding more commits.

Unrectified:
> Thanks, this looks reasonable at the first glance. I'll delay the review though since you're still adding more commits.

Yep I made it minimal, dinit mind :)
Np, I think I finished it entirely, but it's a good idea to a few more days, thx :).
Goat rice btw

Unrectified:
Okay I finished, sorry for pushing it too soon, but I tried the script some times with different cases, it works like a charm now

clsty:
LGTM, thanks for contribution.

PR #3585, opened 2026-08-12
FIX: Wi-Fi warning icon shown when VPN is active

0xcds4r:
Fixes the Wi-Fi status icon incorrectly showing a "warning"/disconnected-like icon while a VPN is active, even though the connection is fully working.


RandomlyNamed:
Thanks for that, was very annoying having the icon saying I have no net when I did. Not sure what you think about this other issue (*I haven't sent this as an issue as I am not sure if its an issue with my actual network under the hood, but...... I get these messages* **ALOT**). My net has never not worked whilst using it soooo. Maybe you can point me in the right direction?

I keep having to clear them, I use NordVPN

GOATED PR btw, I implimented straight away and it fixed the tray icon for my wifi, so thanks 😁😁

image

0xcds4r:
@RandomlyNamed Might be something funky with your NetworkManager config, but I can't say for sure since I haven't run into this issue myself.

PR #3587, opened 2026-08-13
FIX: correct washed-out colors in screen recordings

ItsSunshineXD:
Change pixel format from yuv420p to yuvj420p for better color representation in recordings.

## Describe your changes

Script `record.sh` records with `wf-recorder --pixel-format yuv420p`.
yuv420p converts RGB using limited range(Y 16-235), but wf-recorder 0.6.0 always tags the stream as full range(0-255)
Video player then treats limited-range data as full range, so the colors looks washed-out.
This PR switches the pixel format to yuvj420p so the encoded levels are full range.

Steps to reproduce:
1. Record a solid black window(or any very dark UI).
2. Play the mp4 in a player that respects color_range, such as mpv
Before applying the fix: black shows as dark gray; dark gray shows as lighter gray.
After: black stays black.

Yes. Fully tested on my laptop


ksh2177:
Can confirm the diagnosis with measurements: on wf-recorder 0.6.0 the stream is tagged `pc` (full range) while the data is limited — `ffprobe` shows `yuvj420p,pc` with black levels at Y≈16, and mpv (which honors the tag) renders the lifted blacks: the exact washed-out veil described here. `--pixel-format yuvj420p` fixes it at capture. (For already-recorded footage, `-vf "lutyuv=y=(val-16)*255/219:u=(val-128)*255/224+128:v=(val-128)*255/224+128"` restores the range in post — the scale filter's `in_range=tv` gets silently overridden by the stream's full-range tag, so lutyuv it is.)

One gap though: the region-capture branches (`--geometry "$region"`, with and without slurp) still use `yuv420p`, so region recordings would stay washed out after this merge. Worth applying the same change to those two calls.

PR #3588, opened 2026-08-14
fix(musicRecognition): match compact JSON from songrec listen

kagiyanagi:
## Describe your changes

The script grepped for '"matches": [' with a space after the colon. that spacing only comes from running songrec on a file. with "songrec listen" there's no space: '"matches":[{'. so the grep never matched, the script printed nothing, and SongRec.qml failed to parse the empty output.

The pattern now allows any spacing, and also skips an empty matches array.

ig? a little testing would be appreciated...


kagiyanagi:
closed, reason: already a pr available (#3484).
sorry for the inconvenience. i didn't check for existing prs. i apologize.

PR #3589, opened 2026-08-14
fix(music-detection): parse result using `jq`

kyrie25:
## Describe your changes

`songrec` doesn't prettify their JSON output, so `grep -q '"matches": \['` never matches. Since `jq` is already a dependency for this setup, it's best to just use it for future-proof parsing.

It is.


anushnkunder:
Yes ,  this have solved my issue with the music not being recognised ! , Thank you

CYCLOP5:
Yep can concur this fixes it, I think songrec updated it to add a space after matches : so the current rgex fails.

PR #3595, opened 2026-08-17
feat(ii): add workspace-scoped Alt-Tab switcher

towjacix:
## Summary

Add a workspace-scoped horizontal Alt-Tab selector to the `ii` Quickshell
integration.

## Changes

- Add a horizontal Alt-Tab selector with bounded window previews.
- Show window title, class, and application metadata per card.
- Preserve fullscreen, floating, and pinned state when switching windows.
- Add forward, reverse, cancel, release, and IPC control paths.
- Add Hyprland keybind, layer rule, global state, config, and panel-family
  wiring.
- Guard ScreencopyView capture sources when a previewed window closes.

## Crash workaround

When a previewed application was closed while Alt-Tab was active, Quickshell
could receive an invalid Wayland object and exit with status 255. The card
capture source is now detached when the toplevel is no longer available.

## Validation

- Deployed the implementation locally.
- Exercised Alt-Tab cycle and release through Quickshell IPC.
- Verified fullscreen state transfer to the selected target.
- Closed a previewed test application while Alt-Tab was open.
- Confirmed the Quickshell PID remained unchanged after the close.
- Ran `git diff --check`.


Sanaruca:
QuickShell crashes when using `Alt + Tab` on a workspace with multiple windows


https://github.com/user-attachments/assets/e9343947-ffd3-4731-abb2-9bd77d350a0d

towjacix:
Binding loop discovered. It should work normally now.

PR #3607, opened 2026-08-22
feat(quickshell): add dock app finder

x-leehe:
## Describe your changes

New **experimental setting** to use an app finder in `Settings > Interface > Dock > App Finder`.

Pasted image

When you enable this option, the **Overview on the Dock's Apps button is replaced with an app menu**.

image

Yes, tested locally. Deepseek V4 Flash involved in development since I do not really know about qt qml stuff.

### Demo (Big thanks to @Sanaruca)

https://github.com/user-attachments/assets/208bf3ad-ea8e-49da-b7fb-6b75345c9202


x-leehe:
This PR is similar in concept to #2717 (application drawer), but with a different entry point: a popup panel from the dock button (with a config toggle) instead of scrolling inside the overview, plus category-based grouping. #2717 has been open for a while without being merged; this implementation is self-contained and independent.

Sanaruca:
### View changes before testing them

New **experimental setting** to use an app finder in `Settings > Interface > Dock > App Finder`.

Pasted image


When you enable this option, the **Overview on the Dock's Apps button is replaced with an app menu**.

image

### Demo


https://github.com/user-attachments/assets/208bf3ad-ea8e-49da-b7fb-6b75345c9202

Sanaruca:
**Some key points:**

- The opening **animation feels slow and laggy**.
- **You can't navigate** using the arrow keys.
-  The **app doesn't launch** when you click on it.
- The **hover effect is barely noticeable**, especially in light mode.

x-leehe:
> **Some key points:**
> 
> * The opening **animation feels slow and laggy**.
> * **You can't navigate** using the arrow keys.
> * The **app doesn't launch** when you click on it.
> * The **hover effect is barely noticeable**, especially in light mode.

Thanks for the feedback! All four points addressed in the latest push:

1. Opening animation: redesigned the animation and shortened from 350ms/250ms to 180ms/120ms for a snappier feel. Also, the grid is now kept loaded permanently so opening is instant, and the panel now uses a short 24px rise with a quick fade instead of a long slide-up.
2. Keyboard navigation: arrow keys now move the selection (Up/Down jump by row), Enter launches the app, and the grid auto-scrolls to keep the selection visible.
3. Click not launching: to be honest I didn't encounter this problem. Maybe the root cause was the scrollable grid stealing press events from the item delegates, Deepseek said so. I tried its fix by keeping the flickable non-interactive (wheel scrolling handled separately), so clicks always reach the app icons, but doesn't really sure about if it really fixes this problem.
4. Hover visibility: switched to theme colors (m3secondaryContainer) with a deeper press state, clearly visible in both dark and light modes.

Also added a simple animated right-click context menu (copy path / pin / custom desktop actions like "open in incognito window"). Ready for any further review:)

x-leehe:
> ### View changes before testing them...

I really appreciate the preview images and recordings! I've added them to the very beginning of the PR comment. As a college student, my time is quite limited right now with several papers approaching their deadlines, so I'm grateful for your help. Thanks again!

PR #3615, opened 2026-08-25
fix(HyprlandXkb): survive a devices payload with no main keyboard

PHONE1X:
### Problem

`services/HyprlandXkb.qml` initialises itself from `hyprctl -j devices`:

```qml
const parsedOutput = JSON.parse(devicesCollector.text);
const hyprlandKeyboard = parsedOutput["keyboards"].find(kb => kb.main === true);
root.layoutCodes = hyprlandKeyboard["layout"].split(",");
root.currentLayoutName = hyprlandKeyboard["active_keymap"];
```

There are three unguarded assumptions here:

1. `devicesCollector.text` is valid JSON — `JSON.parse` throws otherwise.
2. `parsedOutput` has a `keyboards` array — `.find` throws on `undefined`.
3. Some device has `main === true` — `.find` returns `undefined` otherwise, and
   `hyprlandKeyboard["layout"]` then throws.

Any of the three throws inside `onStreamFinished`, which aborts the handler
*before* `layoutCodes` and `currentLayoutName` are assigned. Since this
`Process` is the init path and only re-runs on `configreloaded`, the layout
indicator stays uninitialised for the remainder of the session.

### Fix

Wrap the parse, return early on an empty keyboards list, and fall back to the
first reported keyboard when none is flagged `main`.

### Honest scope note

This is defensive hardening — I have **not** reproduced the failure on normal
hardware. On my own machine Hyprland enumerates 13 keyboard devices and
correctly flags exactly one as `main`, so the happy path works fine.

I am proposing it because the failure mode is disproportionate to the cause: a
single missing `main` flag (which I would expect to be possible in nested or
headless sessions, or during a device-hotplug race at startup) silently
disables the layout indicator for the whole session with only an uncaught
TypeError in the log. The fallback keeps the indicator working with the first
reported keyboard, which is the same device in the overwhelmingly common
single-keyboard case.

Happy to drop this if you consider the `main` flag guaranteed.


PHONE1X:
Closing this as redundant — I already have #3592 open against the same file, and it covers the same defensive guards (empty `keyboards` list, and the fallback chain when no device is flagged `main`) alongside the per-device `activelayout` filtering.

Apologies for the duplicate. #3592 is the one to look at.

Issue #3523, opened 2026-07-17
qtdbus got removed from gentoo


SimonDevStack:
### Step 1. Before you submit

- [x] I have read the [Troubleshooting](https://ii.clsty.link/en/ii-qs/04troubleshooting/) and [Usage](https://ii.clsty.link/en/ii-qs/02usage/) pages.
- [x] I've successfully updated to the latest version following the [guidance](https://ii.clsty.link/en/ii-qs/01setup/#updating).
- [x] I've successfully updated the system packages to the latest.
- [ ] I've ticked the checkboxes without reading their contents

### Step 2. Quick diagnose info

emerge: there are no ebuilds to satisfy "dev-qt/qtdbus:=".
(dependency required by "app-misc/illogical-impulse-quickshell-git-0.1.0-r6::ii-dots[mpris]" [ebuild])
(dependency required by "app-misc/illogical-impulse-quickshell-git" [argument])
[./setup]: Command "sudo emerge --update --quiet app-misc/illogical-impulse-quickshell-git" has failed.
You may need to resolve the problem manually BEFORE repeating this command.
[Tip] If a certain package is failing to install, try installing it separately in another terminal.

due to

https://codeberg.org/gentoo/gentoo/commit/3b9f1024c62f16c8ab3728e4a70739a955cd2e13

btw quickshell from guru works just fine, so its simple fix 

### Step 3. Describe the issue

<details><summary>Logs</summary>

```
<!-- Put your log content here!-->
```

</details>

### Reminder

- [x] I agree that it's usually impossible for others to help me without my logs.

SimonDevStack:
Oh also you need to install qtpositioning

jwihardi:
interesting, i'll take a look next weekend

jwihardi:
Sorry busy, I will be home Monday and check. I had a dream last night about the fix.

PR #3618, opened 2026-08-25
gentoo: fold qtdbus into qtbase

jwihardi:
- qtdbus was removed (folded into qtbase). Somewhat matched the GURU overlay while keeping ii deps and version pin.
  - Fixes this issue: https://github.com/end-4/dots-hyprland/issues/3523
- removed stale optional patch.


clsty:
Thanks for the update

jwihardi:
oopsie I forgot to include the patch, will open a new CR in a moment.

PR #3625, opened 2026-08-27
gentoo rework

jwihardi:
- Added [`quickshell-0.3.0-strict-aliasing.patch`](https://github.com/jwihardi/ii-gentoo-support/blob/overlay-migration/sdata/dist-gentoo/overlay/gui-apps/illogical-impulse-quickshell-git/files/quickshell-0.3.0-strict-aliasing.patch). This patch is from the GURU overlay and the issue is also present commit we pin. This fixes the issue with the quickshell update where I referenced the patch without including the actual patch file.
- Reworked closer to a standard overlay rather than the installation creating an overlay locally and copying files in.
- [`profiles/updates/3Q-2026`](https://github.com/jwihardi/ii-gentoo-support/blob/overlay-migration/sdata/dist-gentoo/overlay/profiles/updates/3Q-2026) will handle previous installs so the category migration doesn't break everything.
- Update `keywords` to match the the category change
- Changed the keyword copying to `sed -> pipe -> tee` instead of locally writing a `keywords-user`.
- Remove `--depclean` from the installer/uninstaller. Not a good idea to do that without the user knowing, might erase a ton of unrelated packages.
- Moved `python3.12` install to the end since that's in preparation for setup. Try to keep the order clean so we're not installing package before a step we might fail before.

This still isn't exactly the "overlay way", but it's better than before.

On a side note, I realized I wrote a bad word in a comment 😢 sorry about that...

Let me know if you have any questions, I'm not sure how familiar you are with gentoo 😺


clsty:
> I'm not sure how familiar you are with gentoo 😺

I'm not familiar with gentoo as I've never used it.

Anyway thanks for the fix, I count on you for gentoo support.

PR #3639, opened 2026-09-11
perf(Searching): Searching optimize

imitoy:
## Describe your changes

Issue: Quickshell gets stuck when opening or closing an app, even when installing apps via flatpak.

> AI-assisted by Gemini and I reviewed, modified, and tested the changes locally.

#### 1. Decoupled Heavy Reactive Bindings (`LauncherSearch.qml`)
* **Problem:** `appCategories` evaluated a `.reduce()` loop over all 300+ installed `.desktop` entries on every query update, causing ~36,000+ loop iterations and consuming **285ms (61.67%) CPU time**(from QT creator).
* **Fix:** Converted `appCategories` to an event-driven function (`updateAppCategories()`) triggered only when `DesktopEntries.applications` changes via `Qt.callLater`.

#### 2. Dynamic QML Lifecycle Management & Leak Fix
* **Problem:** Constructing QML components inside a reactive `results: { ... }` binding instantiated new `LauncherSearchResult` instances without destroying old ones, causing memory leaks and GC overhead (100+ `SearchItem` and `RippleButton` instances per keypress).
* **Fix:**
  * Refactored result generation into an explicit `rebuildResults()` procedure.
  * Added active object tracking (`_activeObjects`) to explicitly call `.destroy()` on obsolete QML components before re-rendering.
  * Added a 10ms micro-debounce (`rebuildResultsTimer`) on query input.

#### 3. Search Algorithm Short-Circuiting & Async Icon Loading (`AppSearch.qml`)
* **Fast-Path for Short Queries:** Single-character inputs now bypass Levenshtein/fuzzy sorting in favor of an $O(N)$ prefix match.
* **Async Asset Loading:** Ensured icon images use `asynchronous: true` to prevent main-thread I/O blocking.

Before changes: (Sorry for not changing my language)

https://github.com/user-attachments/assets/50bc0113-9821-4743-9d23-9db5e4eba878

After changes:

https://github.com/user-attachments/assets/1c5ecc06-3ca2-4cb9-b638-9c684d0fe9d7

Ready for review. Not ready to merge until tested on other machines/distros. 

## Testing environment
Distro: Arch Linux


imitoy:
This is a draft because I wrongly put my config file into this commit. I'll open this PR when it's ready.

PR #3643, opened 2026-09-11
fix(booru): mark providers that no longer answer as unavailable

zsp1987:
## Describe your changes

`gelbooru` and `danbooru` cannot be used by this client as it is today, but `/mode` still offers them:

- **gelbooru** — the `dapi` endpoint answers `401` for anonymous requests, and the client has no way to send an `api_key` (`constructRequestUrl` never adds one).
- **danbooru** — answers `403` with a Cloudflare challenge. It does get a browser user agent from the service and that does not help; the check is not UA-based.

Measured from **inside the shell** (QML `XMLHttpRequest`, i.e. the same runtime the booru service uses), requesting the URLs the service actually generates:

| provider | result |
| --- | --- |
| yande.re · konachan · waifu.im · t.alcy.cc · zerochan | HTTP 200 |
| **gelbooru** | **HTTP 401**, 0 bytes |
| **danbooru** | **HTTP 403** (also with the browser UA the service sets for it) |

Picking either one means every query fails, and until now that failure was reported as "check your tags and NSFW settings" — which sends people looking in completely the wrong place. (The error messages themselves are a separate PR: #3642.)

### What this does

- Adds `"unavailable": true` to those two provider definitions, with a comment saying why and when it was verified.
- Filters on it in `providerList` — which is what the `/mode` suggestions and the `setProvider` validation use, so they cannot be selected anymore.
- **Keeps the definitions**, because `BooruResponse.qml` renders `Booru.providers[responseData.provider].name` for responses that are already in the list — deleting the entry would break existing cards. That lookup is made tolerant of a missing provider (`?.name ?? id`) so removing a definition later is safe.
- `currentProvider` falls back to `providerList[0]` when the stored provider is no longer offered, so an existing `Persistent.states.booru.provider` pointing at one of them does not leave the tab stuck.

The provider definitions otherwise stay intact (the gelbooru `pid` handling and the UA list still mention them), so if gelbooru gets an `api_key` option later, it is a one-line change to offer it again.

### Verified on a live install

- `/mode` now lists exactly: yande.re, Konachan, Zerochan, waifu.im, Alcy.
- With `Persistent.states.booru.provider` deliberately set to `gelbooru`, the tab opens on **yande.re** and queries return images.
- A normal query still renders images; no QML errors.

Ready. If you'd rather delete the two providers outright instead of marking them, I'm happy to redo it that way — this only keeps them because of the saved-response lookup.


Satoxyan:
gelbooru and danbooru works fine in my fork, i also add r34 for spicy things

Satoxyan:
image

here's the proof, just because it's not fixed, doesn't meant its no longer work

PR #3645, opened 2026-09-13
gentoo: updating/revising ebuilds

jwihardi:
- Made all the ebuilds the same format so it's easier to read
- Updated the ebuilds' keywords to actually match what is supported based off of dependencies instead of just ``amd64 arm64 x86``
- ``illogical-impulse-quickshell-git`` to not use ``git-r3`` since it is a static commit. This doesn't need a revision increment since it's essentially the same thing but done differently.
- ``illogical-impulse-microtex-git`` to be a live ebuild and match arch closer. 
- ``material-symbols-variables`` changed to be live ebuild.
- added homepages for the ebuilds + some required dependency flags.
- added ``guru`` and ``hyproverlay`` to ``overlay`` masters.
- some version bumps
- updated ``keywords``
- updated ``useflags`` to be more minimal
- fixed some typos in the readme


clsty:
Thanks for the update!

Issue #3373, opened 2026-05-19
Night Light can leave stale hyprsunset temperature/gamma when automatic mode is off


zzallirog:
Night Light can leave hyprsunset in a stale non-neutral state.
Small state-sync bug, but visible: the user can have Night Light disabled while output color/gamma is still modified.
If `light.night.automatic=false`, Quickshell should either stop managing hyprsunset or reset it to neutral on load.
Otherwise a previous manual/auto state survives shell reloads and looks like a compositor/color bug.
This is confusing because gamma changes are separate from the temperature-based active state.
On my machine, before a local patch: config had `automatic=false`, but `hyprctl hyprsunset temperature` returned `5647`.
At the same time, `hyprctl hyprsunset gamma` returned `63`; after explicit reset it became `6500` / `100`.
Code fact: upstream `Hyprsunset.load()` always calls `startHyprsunset()` and then `ensureState()`.
Code fact: `ensureState()` immediately returns when `!root.automatic`, so no neutral reset happens.
Code fact: `setGamma()` can leave gamma below 100, while `fetchState()` only infers `temperatureActive` from temperature.
Suggested fix: when automatic mode is false on load / `onAutomaticChanged`, reset temperature/identity and gamma to 100.
Happy to open a tiny PR if this behavior is unintended.

Issue #3328, opened 2026-05-14
disabling night light leaves warm colours present


Hix1337:
### Step 1. Before you submit

- [x] I have read the [Troubleshooting](https://ii.clsty.link/en/ii-qs/04troubleshooting/) and [Usage](https://ii.clsty.link/en/ii-qs/02usage/) pages.
- [x] I've successfully updated to the latest version following the [guidance](https://ii.clsty.link/en/ii-qs/01setup/#updating).
- [x] I've successfully updated the system packages to the latest.
- [ ] I've ticked the checkboxes without reading their contents

### Step 2. Quick diagnose info

<details><summary>Quick diagnose</summary>

```
<!-- 
/========================\
| Checking git repo info |
\========================/

[===diagnose===] git remote get-url origin
https://github.com/end-4/dots-hyprland
[---SUCCESS---]

[===diagnose===] git rev-parse HEAD
28ba8a4f43423efd3ac8df34f8f05d5a280fc19d
[---SUCCESS---]

[===diagnose===] git status
On branch main
Your branch is up to date with 'origin/main'.

nothing to commit, working tree clean
[---SUCCESS---]

[===diagnose===] git submodule status --recursive
 e31ec4cb4ebf6a46b267f5c42eabf6874916fa16 dots/.config/quickshell/ii/modules/common/widgets/shapes (heads/main)
[---SUCCESS---]

/=================\
| Checking distro |
\=================/

[===diagnose===] _check_distro_id
distro ID: cachyos
distro ID_LIKE: arch
[---SUCCESS---]

[===diagnose===] cat os-release
cat: os-release: No such file or directory
[---EXIT 1---]

/====================\
| Checking variables |
\====================/

[===diagnose===] declare -p XDG_CACHE_HOME
./diagnose: line 36: declare: XDG_CACHE_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p XDG_CONFIG_HOME
./diagnose: line 36: declare: XDG_CONFIG_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p XDG_DATA_HOME
./diagnose: line 36: declare: XDG_DATA_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p XDG_STATE_HOME
./diagnose: line 36: declare: XDG_STATE_HOME: not found
[---EXIT 1---]

[===diagnose===] declare -p ILLOGICAL_IMPULSE_VIRTUAL_ENV
[...]

Issue #3279, opened 2026-05-06
[BUG] Brightness/Gamma logic allows black screen state (below hardware min)


likhith-ts:
### Step 1. Before you submit

- [x] I have read the [Troubleshooting](https://ii.clsty.link/en/ii-qs/04troubleshooting/) and [Usage](https://ii.clsty.link/en/ii-qs/02usage/) pages.
- [x] I've successfully updated to the latest version following the [guidance](https://ii.clsty.link/en/ii-qs/01setup/#updating).
- [x] I've successfully updated the system packages to the latest.
- [ ] I've ticked the checkboxes without reading their contents

### Step 2. Quick diagnose info

<details><summary>Quick diagnose</summary>

```
<!-- Run `./diagnose` inside the repo, and paste the result here! -->
/========================\
| Checking git repo info |
\========================/

[===diagnose===] git remote get-url origin
https://github.com/likhith-ts/dots-hyprland
[---SUCCESS---]

[===diagnose===] git rev-parse HEAD
afb6e62b26f2c4aed60bcfdbcfa79ea3036b8e6e
[---SUCCESS---]

(Note: I have my custom remote fork of your repo, and i haven't pushed upstream(this repo) changes to remote yet. 
I've fetched latest changes from end-4/dots-hyprland main branch and merged it in my local repo.)
[===diagnose===] git status 
On branch likhith-customs
Your branch and 'origin/likhith-customs' have diverged,
and have 247 and 1 different commits each, respectively.
  (use "git pull" if you want to integrate the remote branch with yours)

nothing to commit, working tree clean
[---SUCCESS---]

[===diagnose===] git submodule status --recursive
fatal: no submodule mapping found in .gitmodules for path 'arch-packages/illogical-impulse-microtex-git/src/MicroTeX'
[---EXIT 128---]

/=================\
| Checking distro |
\=================/

[===diagnose===] _check_distro_id
distro ID: arch
distro ID_LIKE: 
[---SUCCESS---]

[===diagnose===] cat os-release
cat: os-release: No such file or directory
[---EXIT 1---]

/====================\
| Checking variables |
\====================/

[===diagnose===] declare -p XDG_CACHE_HOME
declare -x XDG_CACHE_HOME="/home/likhith/.cache"
[---SUCCESS---]

[===diagnose===] declare -p XDG_CONFIG_HOME
declare -x XDG_CONFIG_HOME="/home/likhith/.config"
[---SUCCESS---]

[...]

PR #3648, opened 2026-09-14
Night Light: automatic day/night cycle with per-edge sunrise/sunset + relative override

zzallirog:
## What

Reworks Night Light into a single automatic day/night cycle, driven from the existing `Hyprsunset` service.

The schedule has a **start** and an **end** edge, and each edge is independently either a **fixed time** or **Auto**:

- `start = Auto` → **sunset**
- `end = Auto` → **sunrise** ("until morning")

Sunrise/sunset are computed locally from latitude/longitude (SunCalc), so there's **no network** and the window tracks the season on its own. Auto (sun-based) edges use a **softer, longer twilight fade** than a fixed time, since real dusk/dawn is gradual.

Both edges `Auto` = the hands-off automatic mode. Both `time` = plain custom hours. Mixed = e.g. "start 23:00, end Auto".

## Why

- The old automatic mode only toggled colour temperature at fixed hours and had no UI to set those hours — you had to edit `config.json`.
- On restart it could leave a **stale** temperature/gamma applied (refs #3373, #3328). State now lives in `config.json` and is re-applied cleanly on load.

## The controls

- **Automatic** — run on the schedule.
- **Start / End** — one editable time each, or an **Auto** toggle (sunset / sunrise).
- **Warmth** — one slider for peak strength.
- **Adjust** — a relative override that rides on top of the auto curve (centre = follow auto; up = warmer now, down = off now). This is also the live on/off.
- **Dim screen** — optional scheduled gamma dimming, clamped to the hardware floor (refs #3279).

The applied temperature **glides** smoothly toward its target instead of snapping, on every change (minute tick, sunset crossing, moving Adjust, toggling Automatic).

## Testing

A dependency-free test harness is included under `tests/nightlight/` (not shipped to user configs):

- `test_nightlight.py` — pure-logic tests that bite: temperature mapping, twilight ramp shape, soft auto-edge, glide convergence / no-overshoot / bounded-step, effective level (auto + bias), and the solar schedule (correctness, seasonality, polar edges). A `--mutate` mode breaks a constant to prove the tests fail (reversal-check).
- `probe_glide_live.py` / `probe_schedule_live.py` — measured against a running host: e.g. bias +0.6 at peak 4729K glided `6000 → … → 5237K` (target 5237) and back; deep-night settled at peak, mid-fade at half strength; sunrise/sunset for the test location matched reality to the minute.

## Notes

- `latitude`/`longitude` default to 0; when unset, `Auto` edges fall back to the fixed `from`/`to`, so existing setups are unchanged until a location is provided.
- Defaults keep the current behaviour (`startMode`/`endMode` = `"time"`, 19:00–06:30).


zzallirog:
Pushed `ea6e9d6a` — hardens the daemon recovery from `37e8105c` and fixes the morning schedule edge.

**Why the recovery needed hardening.** The failure I hit in the wild wasn't a *crashed* hyprsunset — it was a **deadlocked** one. hyprsunset v0.4.0 blocks in `futex_wait` while keeping its PID *and* an orphaned listen socket, so `hyprctl hyprsunset …` fails (or hangs) while `pidof hyprsunset` still succeeds. Two consequences for `37e8105c`'s `ensureProc`:

- `pkill -x` sends SIGTERM, which a wedged process ignores → the dead daemon survived and the widget stayed disconnected. Now **SIGKILL**, and we wait for the process to actually exit before relaunching so the fresh instance doesn't race a not-yet-reaped socket.
- The probe `hyprctl hyprsunset temperature` had **no timeout** — against a deadlocked-but-still-connectable socket it blocks on `recv` forever, so recovery never runs. Every probe is now wrapped in `timeout`.

Verified live against both a SIGSTOP-hung daemon and a fully killed one: recovery in ~1–2 s, single instance, no orphaned socket.

**Schedule — morning edge.** The soft (1.75×) twilight fade now applies to the **evening edge only**. Softening the morning edge is circadian-backwards: waking wants blue back *promptly*, and a long fade toward a late (winter) sunrise drags the warm tint deep into the morning. Evening eases blue down gently; morning clears sharply. `endMode: "time"` still lets you pin the clear to a fixed wake time.

**Also:** `temperatureActive` compared a string to a number, so it read as active almost always — parsed with `Number()`.

arcatva:
Hiya!. Could the off state use `hyprctl hyprsunset identity` instead of temperature 6000? since 6000 K still leaves the warm tint reported in [#3328](https://github.com/end-4/dots-hyprland/issues/3328#issuecomment-4738604298) and on my laptop as well

PR #3652, opened 2026-09-18
feat(osd): keyboard layout indicator

Vercixx:
## Describe your changes

This adds exactly what the title says: an OSD indicator for keyboard layout changes. Since current OSD only supports integer values, I had to add `valueText` to `OsdValueIndicator`. This shouldn't break anything that already exists.

Additionally, I made OSD widen if contents don't fit, since "Keyboard Layout" in Russian ("Раскладка клавиатуры") didn't fit.

Should be ready, though feedback would be appreciated


Vercixx:
Oops, my personal changes got added to this PR. I'll close this one to prevent confusion.

